PR #160, opened 2026-04-17
Feature/sh 103 shop in the venue

J-Melon:
## Summary

SH-103 asks for the shop to live at the edge of the court and to drag from there without a menu. The existing shop was a hidden Control panel; meeting the ticket meant rebuilding as a diegetic world-space object. `Shop` is now a `Node2D` at the court's right edge; each item is a `RigidBody2D` under gravity. Drag an item and release outside `ShopArea` to fire `ItemManager.take(key)`. The Control drag pipeline (`ClearanceBox`, `ShopTooltip`, `ItemDragging`, `DragManager` autoload) is gone; physics picking handles everything. Matches `designs/01-prototype/08-shop.md`.

### Stacked scale hacks

Court, ball, paddles, and shop had been authored small and scaled up by parent transforms to fit `Camera2D(zoom=0.52)`. `RigidBody2D` strips parent scale from its physics shape, so shop-item visuals and collision disagreed — the "only some items drag" bug. Every `scale = (n, n)` removed; court, ball, paddles, walls, miss-zone, and shop re-authored at their intended world sizes; `camera.zoom = 1.0`. `base_stats.tres` retuned for the new arena.

### Input pipeline leaks

With picking fixed, leftmost items still didn't respond: decorative `ColorRect`s default to `mouse_filter = STOP` and were eating events before physics picking ran. Every decorative `ColorRect` set to `IGNORE`, and `Venue` itself demoted to `IGNORE` so clicks reach the scene. `DevShopPanel` carries a debug canary that warns if a click lands on a ShopItem AABB but picking never reaches the body.

### Save resilience (SH-34 folded in)

Intermittent "cleared save came back" reports were a race between `clear_save()` and the deferred scene reload: autosaves fired in the gap and overwrote the cleared file. `clear_save()` now stops the autosave timer and blocks `save()` until `unblock_writes()`. `FileSaveStorage` writes atomically (temp + rename) and keeps three rolling backups; `load_from_disk` walks primary-then-backups until one parses. `save_to_disk` serialises through `JSON.stringify`; the storage layer parks the primary to `.incoming`, commits the temp, and only then rotates the backup chain, so a failed write leaves existing saves and backups untouched.

### Review follow-ups

The shipping fix for atomic writes reordered backup rotation so a failed temp-write can't destroy the primary: write + flush + error-check the temp first, then rotate and rename. Internal terminology realigned with the design doc — `taken` became `owned` across `ShopItem`, `Shop`, dev panels, and tests; `last_input_frame()` became `get_last_input_frame()` to match getter convention. `DraggableBehavior` now splits `try_start` (from `_gui_input`) and `update` (from `_input`) so fast drags that outrun the control's bounds still deliver mouse-up and don't orphan drag state. Dev panels `queue_free()` themselves in release builds instead of `hide()`ing, so they're not just invisible but truly absent from input and process loops.

### Smaller fixes

Recruit button was `FOCUS_ALL` and Space-auto-recruited Martha; now `FOCUS_NONE`. Shop only spawns `level == 0` items. Cased shop items freeze kinematically so owned items can't bash them.

### Dev tooling

`DraggableBehavior` mixin plus a new `DevMenu` (checkbox toggles) and `DevShopPanel` (grid of per-item state). All four dev panels draggable in debug builds.

### Tooling refresh

Pre-commit's manual-run two-pass (gdformat rewrites files → framework flags the hook as failed → re-run) kept biting during lint sweeps, so the hook runner moved to lefthook, which re-stages auto-fixes cleanly. While touching it, the guardrails widened: gut runs in pre-commit behind a `.gd`-file filter with a wrapper that shows coverage + totals on success and full output on failure; codespell joined the pre-commit chain with a British-English config and a per-project ignore list; commit messages are now format-checked (`[SH-<id> ]<type>: <subject>`); `.editorconfig` pins tabs-for-`.gd` and LF endings; a new `lint.yml` CI workflow mirrors the checks so pre-commit bypass doesn't silently slip through. Same tools, just enforced in more places.

## Test plan

- [ ] Clear save, earn FP, Clearance, drag item outside shop — owned, balance drops, body falls to floor
- [ ] Bash owned item into cased — cased stays put
- [ ] Clear + idle + quit — reloads as cleared
- [ ] Corrupt primary save file — loads from newest parseable backup
- [ ] Every shop item responds to clicks
- [ ] Dev menu toggles each overlay; all dev panels draggable
- [ ] Commit with a malformed message — `commit-msg` hook rejects it

Closes SH-103, SH-34


scripts/shop/shop_item.gd
```diff
@@ -1,196 +1,140 @@
...
+	if mouse_button.button_index != MOUSE_BUTTON_LEFT:
+		return
+	_last_input_frame = Engine.get_physics_frames()
+	if mouse_button.pressed and can_be_dragged() and not _dragging:
+		_start_drag()
 
 
-func _get_cost_text() -> String:
-	return "%d FP" % _item_manager.calculate_cost(item_definition.key)
+func last_input_frame() -> int:
```
J-Melon: suggestion: this is a getter, prepend with get

scripts/shop/shop_item.gd
```diff
@@ -1,196 +1,140 @@
...
+		_refresh_case_overlay()
 
 
-func _on_display_case_gui_input(event: InputEvent) -> void:
-	if not (event is InputEventMouseButton):
+func _refresh_case_overlay() -> void:
+	if case_overlay == null:
 		return
-	if not event.pressed or event.button_index != MOUSE_BUTTON_LEFT:
+	# Taken items have left the shop's price-gate; case stays off regardless.
```
J-Melon: issue: taken is the wrong term, the agreed term is owned as you'll see in the docs
suggestion: rename taken to owned throughout

PR #169, opened 2026-04-17
SH-104 refactor: diegetic signage on the court

J-Melon:
## Summary

Moves the volley counter, personal best, friendship-point balance, autoplay indicator, FP bonus, speed bar, and recruit panel off the screen-space HUD and onto the court as world-space children. The old `Hud` CanvasLayer becomes `DevHUD`, stripped down to developer overlays (dev menu, item panel, shop panel, stat panel, clearance button).

Each piece of signage is its own script under `scripts/court/` that subscribes to the relevant domain signal (or `ItemManager` for the FP balance) and updates itself. `court.gd` only emits events; it never reaches into a label. This matches the design-doc framing that these are starting-point stand-ins for later systems (a partner announcing the streak, a character handing over FP), so promoting any one piece to a richer scene later is a drop-in swap with no structural churn. The FP bonus retains its pink/red tint; labels carry explicit `theme_override_font_sizes` so they stay legible at Court's 0.6 scale.

While refactoring the signage wiring, `ball_speed_updated` turned out to be a per-frame poll dressed up as a signal. `Ball` now owns `speed_changed` and emits only at actual change sites; `SpeedBar` subscribes to the ball directly, and reads the permanent-cap base off `ItemManager.item_level_changed`. Court's `_physics_process` no longer re-emits speed every tick. A small drive-by also fixes `shop_item.gd` using direct `freeze =` during physics-query flush (`set_deferred` now).

Tests are reorganised along player flows: `test_speed_bar_flows.gd` asserts that rallying raises the bar and missing resets it, end-to-end through Ball + Paddle + Court + SpeedBar. Coverage target raised from **50/25 → 75/50**; `speed_bar.gd` and `outcome.gd` are explicitly excluded in `pre_run_hook.gd` (drawing paths unreachable in headless tests; abstract base class), both documented inline. `lefthook.yml`'s `exclude: "^addons/"` pattern was silently a no-op in 2.1.6, letting addon files into codespell/gdformat/gdlint/gut — switched to the list-form glob `["addons/**"]` which actually filters.

Closes SH-104.


scripts/core/court.gd
```diff
@@ -22,14 +22,14 @@ const MissZoneScene: PackedScene = preload("res://scenes/miss_zone.tscn")
 var player_paddle: Paddle
 var partner_paddle: PartnerPaddle
 
-var _volley_count := 0
+var _volley_count: int = 0
```
J-Melon, line 25: style: these types are obvious
suggestion: use inference instead

PR #189, opened 2026-04-19
chore: route reviewer judgment calls to PR comments

J-Melon:
The parallel workflow already splits review into two passes: an automatic one after `gh pr create`, and an optional follow-up when Josh asks for another look. The follow-up rule lives at the bottom of step 4 and already routes its findings to PR comments. The initial pass, though, used to funnel everything through commits and then expected the agent to raise judgment items separately in chat, which fragmented the review trail.

This change codifies the split on the first pass too. Mechanical fixes keep their current path as commits on the branch. Judgment calls go into one formatted PR comment via `gh pr comment <N>`. Handoff to Josh waits on both. The PR view becomes the single place where open review items live, and chat stops carrying review state that the PR should own.


J-Melon:
Auto-review (dogfooding the new rule).

**Mechanical fixes:** none. Diff is clean, codespell passes, markdown renders.

**Judgment calls:**
- The follow-up review paragraph (line 19) now partially restates step 4. Its clarifier "Initial review fixes still auto-commit per step 4; only follow-up reviews are comment-only" was load-bearing when step 4 was silent on comments. Now that step 4 is explicit about the split, that sentence could be tightened or dropped. Left as-is for Josh's call: the redundancy is mild and arguably reinforces the contrast between initial (mixed) and follow-up (comment-only).

J-Melon:
LGTM

PR #190, opened 2026-04-19
chore: scope preview deploy to Web preset only

J-Melon:
The preview workflow only publishes the Web build to itch's html5-preview channel, yet firebelley/godot-export was exporting every runnable preset in export_presets.cfg. With Linux, Windows, and macOS now runnable, preview jobs paid for three extra exports whose artefacts were never pushed anywhere, and a macOS preset misconfiguration could fail the whole preview even though nothing shipped from that preset.

Scoping the preview export to the Web preset keeps CI minutes proportional to what actually reaches players and isolates preview health from native-platform regressions. release.yml still exports all four presets on tagged releases, so Linux, Windows, and macOS continue to ship through their existing path.

Implementation uses the action's native presets_to_export input rather than a bespoke godot --export-release shell step: less custom glue, preserves the Godot version pin and cache, and leaves the existing butler push pointing at the same build_directory/Web folder.


J-Melon:
LGTM

PR #191, opened 2026-04-19
SH-116 chore: set macOS bundle identifier

J-Melon:
Follow-up after #188 merged. macOS export fails without `application/bundle_identifier` (seen in the Preview Deploy run against `main` before the preview preset-scope fix). Sets it to `games.shuck.volley` so `release.yml` can build macOS on the first release.


J-Melon (commented):
LGTM. Minimal, correct fix: `games.shuck.volley` is a valid reverse-DNS identifier and matches the Shuck brand. Branch is current with main, DCO signed, lint + tests green. Unblocks the macOS export step in release.yml.

PR #203, opened 2026-04-19
SH-127: require ticketed branches for chore/infra work

J-Melon:
Chore and infra branches have been slipping onto main as `chore/<slug>` without a Linear ticket behind them, which breaks traceability for anyone asking "why did this land?" six months later. This change closes that gap with a tiny helper and a one-line rule update.

`scripts/dev/new-ticket.sh "<title>"` posts a GraphQL mutation to Linear's API and creates an issue under the Shuck team with the defaults CLAUDE.md already prescribes: Backlog status, Feature label, assigned to Josh, estimate 0 (chore/infra sits with bugs on the pointing rule; stories stay unpointed for Josh). It prints `SH-N` on stdout so it composes cleanly with a branch creation one-liner, and writes the issue URL to stderr so you still see confirmation. The helper lives next to `new-branch.sh` in the gitignored `scripts/dev/` folder because, like its sibling, it's a personal-machine tool that needs `LINEAR_API_KEY` to be useful.

`ai/PARALLEL.md` step 1 now spells out the rule: if you are starting chore/infra work and there's no SH-N, file one first, then branch as `sh-N-...` so the existing `prepend_ticket_prefix.sh` hook picks it up in commit messages. The `chore/<slug>` pattern is explicitly retired for anything that lands on main. Old merged PRs are left alone.

This PR dogfoods the helper: SH-127 was filed via `new-ticket.sh` and the branch name follows the new convention.

See https://linear.app/shuck-games/issue/SH-127


ai/PARALLEL.md
```diff
@@ -6,7 +6,21 @@ Live scratchpad for parallel agent work on individual Linear tickets. One agent
...
+
+   **Cycle placement on pick-up.** Active work lives in the cycle. If you're claiming an existing Backlog ticket that has no `cycle` set, move it into the currently active cycle as part of the claim. New/unclaimed tickets stay cycle-less per the usual Josh convention; pick-up is what triggers the cycle move. Query the active cycle and patch the issue:
+
+   ```
```
J-Melon: can this be via the mcp instead?

J-Melon:
Good call. Swapped the curl/jq block for `mcp__linear__list_cycles` + `mcp__linear__save_issue` in 8ad... — pushed. Agents in a Claude Code session already have those tools, no auth juggling needed.

PR #207, opened 2026-04-19
SH-129: document rc cleanup rule in release playbook

J-Melon:
Captures a small release-hygiene rule the session surfaced: when a `-rc.N` prerelease ships clean, delete it once the real release is live. Leaving rc tags around clutters the Releases page and confuses future semver tooling. Adds the `gh release delete --cleanup-tag` one-liner and a note that failed rcs iterate forward (`rc.2`, `rc.3`).


J-Melon:
LGTM

PR #209, opened 2026-04-19
SH-130: document weekly release cadence

J-Melon:
Captures the weekly release rhythm Josh shared this session: real releases cut on Tuesdays around 09:00 UK time, with hotfix slots reserved for genuine breakage. Anything that lands after Tuesday rides to the following week.


J-Melon:
LGTM

PR #212, opened 2026-04-19
SH-131: CI speed pass (concurrency, permissions, import cache)

J-Melon:
The PR checks currently run every time, in full, on every push, even when a previous attempt is still mid-flight and even when the diff is a single markdown file. Release and preview deployments reimport the entire project from cold on every run because `.godot/` is rebuilt from scratch. None of this buys safety; it buys waiting.

Four changes, one ticket.

**Concurrency groups cancel superseded runs.** A rebased push used to leave two Godot imports racing to completion on the same PR. Now the second cancels the first. The group keys on `github.ref`, so main and each PR get their own lane. Release/push-to-main flows keep `cancel-in-progress: true`; `sync-wiki` stays at `false` so we never partially mirror the designs directory.

**Least-privilege default at the workflow level.** Deploy jobs push to itch.io with `BUTLER_API_KEY`; nothing in this repo needs the default `GITHUB_TOKEN` to be writable. `permissions: contents: read` at the top shrinks the blast radius of every runner. Overlaps with SH-120 but does not duplicate its scope: SHA-pinning for third-party actions stays with that ticket.

**Doc-only PRs skip the test workflow.** A PR that touches only `designs/`, `docs/`, or `**/*.md` gains nothing from a 2-3 minute Godot import plus ggut. Lint still runs on every PR because codespell earns its keep on markdown. One caveat called out in the commit: if Tests is a required status check on the protected branch, GitHub's `paths-ignore + required` interaction may need the required-workflow stub pattern; Josh owns that decision.

**`.godot/` import cache on the preview-deploy flow.** The test workflow already caches import artifacts keyed on tracked assets. The preview deploy did not. Both jobs inside `publish.yml` now hydrate `.godot/` from cache before firing Godot, so unchanged content PRs skip the reimport stage entirely. `firebelley/godot-export` already caches the Godot binary + templates via `cache: true`; this adds the project-side half.

**Release workflow deliberately untouched.** SH-116 is actively editing `release.yml` (Linux preset switch); dropping a parallel diff on the same file would collide. The same concurrency + permissions + import-cache pattern applies there and can land as a follow-up once SH-116 merges.

Expected wall-clock impact on a typical feature PR: concurrency saves a full superseded run (3-5 min recovered whenever a push lands during an in-flight run). Doc-only PRs drop from ~3 min to zero. Preview deploy on unchanged-asset pushes saves the ~30-90s reimport phase. No behaviour change to what is verified or shipped.

Follow-ups worth their own tickets: SHA-pin every third-party action (SH-120's scope); re-apply this same pass to `release.yml` after SH-116 lands; investigate parallel ggut sharding if ggut wall-clock grows past a minute; reconcile the `git describe` in `publish.yml` against the default shallow clone (needs tags, currently implicit).


J-Melon:
LGTM

PR #214, opened 2026-04-19
SH-132: shrink export sizes across platform presets

J-Melon:
Only the Web preset previously had an `exclude_filter`. The Linux, Windows, and macOS presets shipped with empty filters, so every release build was bundling `addons/gut` (3.2M of test framework), `addons/godotiq` (260K), `addons/signal_lens` (248K), `addons/gdfxr` (180K), `addons/item_preview` (44K), `addons/coverage` (48K), and the whole `tests/` tree (516K) into the .pck even though none of that is referenced at runtime. `publish.yml` patches `project.godot` to unhook some of the plugins before export, but the source files remain on disk and Godot's `all_resources` filter happily packages them anyway.

This change unifies the `exclude_filter` across all four presets and drops the console-wrapper sidecar on desktop release builds. It also belt-and-braces excludes the doc directories (`designs/`, `ai/`), shell-only script paths (`scripts/dev/`, `scripts/ci/`), and repository meta files (CLAUDE.md, PARALLEL.md, README.md, CONTRIBUTING.md, lefthook, gdlintrc, etc.). Godot's `all_resources` filter would not pick most of those up regardless, but keeping the filter exhaustive is cheap and prevents a stray `.import` sibling from smuggling anything in later.

Projected per-platform deltas (pre-compression, lower bound; actual exe shrinkage is larger because compiled bytecode + zstd bundling multiplies the savings):

| Platform | Pre-change bundled | Post-change bundled | Projected saving |
|---|---|---|---|
| Linux `volley.x86_64` | full source tree + addons (~5M unpacked) | runtime-only | ~4.5M pre-compression |
| Windows `volley.exe` | same | runtime-only | ~4.5M pre-compression |
| macOS `volley.app` (universal) | same, doubled into both arch slices | runtime-only | ~9M pre-compression (universal doubles everything) |
| Web `index.wasm`/`.pck` | already filtered; picks up tighter closure (+coverage, gdfxr, signal_lens, item_preview, designs/, scripts/dev/, scripts/ci/, repo meta) | tighter runtime-only | ~500K+ off the .pck |

Unchanged knobs (already optimal):

- `script_export_mode=2` (compiled bytecode tokens) on every preset. Text and binary modes both produce larger exports.
- `binary_format/embed_pck=true` on desktop. Keeps distribution to a single binary; no size cost vs shipping a sibling .pck.
- Desktop texture formats: `s3tc_bptc=true, etc2_astc=false`. ETC2/ASTC stays off for desktop-only presets where mobile formats aren't needed.
- Butler `patch_delta_compression_level_zstd=19`. Already maxed; only matters for itch patch updates, not first-push archive size.

Web precompression was investigated and deliberately skipped. itch.io's CDN applies gzip transparently to .wasm/.pck on delivery and strips any `.gz` extension, so pre-gzipping the wasm into the export either duplicates work or gets suffix-mangled. Brotli pre-compression would need a custom HTML shell loader and is out of scope for this ticket.

macOS is left on `binary_format/architecture="universal"`. Cutting to `x86_64` (Rosetta fallback on Apple Silicon) would roughly halve the .app payload, but the perf hit on Apple Silicon hardware is real and that is a product call rather than a size-optimisation call. Flagging for Josh: if .app size becomes a distribution blocker, single-arch x86_64 + Rosetta is the next lever.

Ran `ggut` locally: 345/345 pass. No gameplay-facing behaviour changed; the presets just stop shipping files that never needed to ride along.


J-Melon:
LGTM

PR #216, opened 2026-04-19
SH-133: mandate git worktrees for sub-agents, require post-merge cleanup

J-Melon:
Three concurrent infra agents today (SH-131, SH-132, SH-117) stepped on each other's working-tree changes, with one accidental prefixed commit and a stash-shuffle dance to recover. The fix is to stop sharing the main tree: every sub-agent that writes to the repo now uses Claude Code's `isolation: "worktree"` so each lands in its own `git worktree` checkout with its own `.godot/` cache.

Adds two new ground rules to `ai/PARALLEL.md`: worktree-per-agent is mandatory for parallel dispatch, and the worktree plus branch are removed after merge so stale state does not accumulate. The main working tree stays for interactive single-agent work.

### DIFF
--- a/ai/PARALLEL.md
+++ b/ai/PARALLEL.md
@@ -39,7 +39,8 @@ Live scratchpad for parallel agent work on individual Linear tickets. One agent 
 
 ## Ground rules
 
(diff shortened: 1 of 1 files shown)


J-Melon:
LGTM

PR #219, opened 2026-04-19
SH-135 ci: speed pass on release.yml

J-Melon:
The production release workflow never received the treatment the rest of CI picked up under SH-131: it was skipped then to avoid stomping on SH-132's in-flight edits to the same file. Now that SH-132 has landed, this pass brings `release.yml` back in line with `publish.yml` and `test.yml` so release builds don't sit on stale caches or drift on permissions.

Four small moves, all hygiene:

- A `concurrency` group keyed on the release tag with `cancel-in-progress: false`. Two publishes of the same tag will queue rather than race, but we explicitly don't cancel a build in flight: once the exporter is halfway through, throwing that work away costs more than letting the queued run supersede it.
- Job-level `permissions`: `contents: read` on the test job, widened to `contents: write` on `deploy-production` only. The test job doesn't need write access to fetch and run GUT, and scoping the deploy-only step keeps the blast radius of any third-party action narrow.
- The `.godot/` import-artifact cache on both jobs, keyed the same way as `test.yml` and `publish.yml`. Release re-imports the project twice today (once in test, once in export); caching cuts the cold portion of each by roughly a minute on warm runs.
- `timeout-minutes: 30` on every job, sized at ~2x observed wall-clock. A hung Godot export or butler upload currently sits on a default-hour timeout; halving that surfaces hangs faster without biting real runs.

No behaviour change: same exports, same itch channels, same upload order. Next release is the first one that'll exercise the new config; a dry run isn't really possible on `release:` triggers without cutting a tag.

SH-116 also touches this file but is parked waiting on Josh's Linux export preset commit; whichever PR lands second merges main in before requesting review.


.github/workflows/release.yml
```diff
@@ -42,6 +59,9 @@ jobs:
     name: Deploy to Production
     needs: test
     runs-on: ubuntu-latest
+    timeout-minutes: 30
+    permissions:
```
J-Melon, line 63: Judgment: `contents: write` is broader than this job actually needs. The deploy path here is `firebelley/godot-export` (reads repo, writes to `$RUNNER_TEMP`) and `butler push` (authenticates via `BUTLER_API_KEY`, talks to itch only). Neither writes back to the repo via `GITHUB_TOKEN`. `contents: read` would suffice and keeps the blast radius matching the test job. Worth tightening unless a future step here is expected to create tags/releases/commits.

.github/workflows/release.yml
```diff
@@ -53,6 +73,14 @@ jobs:
           sed -i 's|, "res://addons/gut/plugin.cfg"||g' project.godot
           rm -rf addons/coverage
 
+      - name: Cache Godot import artifacts
+        uses: actions/cache@v4
+        with:
+          path: .godot
+          key: godot-import-${{ env.GODOT_VERSION }}-${{ hashFiles('project.godot', '**/*.tscn', '**/*.tres', '**/*.gd', 'assets/**', 'art/**', 'resources/**') }}
```
J-Melon, line 80: Judgment: cache key on `deploy-production` will not hit the cache populated by `test` in the same run. The `Strip dev addons` step (lines 67-72) mutates `project.godot` before this `hashFiles(...)` evaluates, so the deploy key hashes the stripped file while the test key hashed the un-stripped file. The two jobs compute different keys and each pays a cold import on the first release after a source change. `restore-keys` prefix fallback softens this (stale `.godot/` still partially reused), but the headline "cache cuts a minute off each job" only holds once the prefix restore is warm. Two options if you want test→deploy sharing: move the strip to after the cache restore (accept that the cached `.godot/` was imported with addons present, then re-import on top), or compute the key from a source set that does not include `project.godot`. Neither is obviously correct; flagging for visibility rather than requesting a change.

.github/workflows/release.yml
J-Melon: Yep tighten

.github/workflows/release.yml
J-Melon: Well what is the correct option then?

PR #221, opened 2026-04-19
SH-136: add asset-pipeline specialist reviewer

J-Melon:
The specialist trigger map in `ai/PARALLEL.md` had a gap for Godot project-config diffs. When SH-132 shrunk export sizes via `export_presets.cfg`, no specialist matched, so no review fired. This adds `asset-pipeline` covering `export_presets.cfg`, `project.godot`, and `**/*.import`, and wires it into the trigger map.

The rubric focuses on what other specialists cannot see: preset parity across platforms, exclude filters silently catching `res://` paths the game loads, platform flag tradeoffs (architecture, ETC2 ASTC, codesign), autoload order changes, and import-setting drift. Explicit out-of-scope list keeps it from duplicating `gdlint`, `godot-scene`, or `ci-and-workflows`.


J-Melon:
LGTM

PR #224, opened 2026-04-19
SH-121 chore: add dependabot config for actions and pip

J-Melon:
Adds a `.github/dependabot.yml` so the two dependency surfaces we actually have get tracked without manual drift checks. `github-actions` covers the action pins across `.github/workflows/*.yml` (publish, release, test, lint, sync-wiki). `pip` covers `requirements-dev.txt`, currently `gdtoolkit` and `codespell`.

Cadence is weekly so the PR volume stays predictable and matches the Tuesday release rhythm. Both ecosystems route PRs to `J-Melon` as assignee and reviewer so nothing queues silently in a bot-owned tab.

Nothing else in the repo has a supported Dependabot ecosystem today (the game itself is Godot, assets are handled out-of-band), so the config stays deliberately small rather than speculative.

Closes SH-121.


J-Melon:
LGTM

PR #225, opened 2026-04-19
SH-126: gate pip-audit on requirements-dev changes

J-Melon:
Supply-chain hygiene for the dev toolchain. The pre-commit hook already vets code style, leaks, and tests; there is no equivalent gate on the Python packages the toolchain itself pulls in. This wires `pip-audit` into lefthook so any change that touches `requirements-dev.txt` has to clear a vulnerability scan before the commit lands.

The hook is gated on the `requirements-dev.txt` glob so day-to-day commits pay nothing. When the file is staged, `pip-audit -r requirements-dev.txt --strict` runs against the OSV/PyPI feeds; a known CVE in any listed package (or a transitive dep it resolves to) exits non-zero and blocks the commit. `--strict` also treats dependency-collection failures as errors rather than warnings, so a half-broken audit cannot silently pass.

`pip-audit` is added to `requirements-dev.txt` itself, pinned to `2.*` to match the `gdtoolkit==4.*` pattern, so any contributor who runs `pip install -r requirements-dev.txt` picks it up automatically.

Verified locally by running `lefthook run pre-commit` with `requirements-dev.txt` both staged (pip-audit runs, no vulnerabilities) and unstaged (pip-audit skipped with `no matching staged files`).


J-Melon:
LGTM

PR #226, opened 2026-04-19
SH-118/SH-119: pin Godot and butler downloads to SHA256

J-Melon:
Until now the CI and release pipelines trusted whatever bytes GitHub Releases and broth.itch.zone happened to hand back for Godot 4.6.2, the matching export templates, and `butler`. A compromised mirror, a silent upstream rebuild, or an unannounced `LATEST` bump could have shipped a different binary into our builds without anyone noticing, and a broken butler release could have appeared in a production push without warning.

This PR closes that gap in one go. It pins the three artefacts to the exact SHA256 each workflow currently runs against, and moves `butler` off the floating `LATEST` alias onto `15.26.1`. Every `curl` that fetches one of these artefacts is followed by `sha256sum -c -` on the downloaded file, so any drift fails the step non-zero before unzip, before the action runs, before anything touches itch.io.

The `firebelley/godot-export` action still performs its own download internally; it is fed the same immutable GitHub release URLs we use elsewhere, and we add an explicit pre-flight step in `publish.yml` and `release.yml` that fetches both the editor and the export templates and asserts their hashes match. If GitHub ever returns anything other than the expected bytes, the pre-flight trips first and the action never gets a chance to run against drifted artefacts.

Closes SH-118 and SH-119 together because they cover the same supply-chain attack surface and share the same verification pattern; splitting them would have meant two PRs touching the same three workflow files within a day.

Hashes pinned:
- Godot 4.6.2 Linux editor zip: `30e6b6d141f0cd5bebd629ad1d0ef1324e60091bb20662d026b402ba58c59937`
- Godot 4.6.2 export templates: `942366dc4e27e7686a99da4d3cfb1b8ae8d3eb9444f6d8217eef16245b599ef2`
- butler 15.26.1 linux-amd64 archive: `b724504bccf259eb01668e1d20b97f6e376d965761e306a75530e19225c510e3`


J-Melon:
LGTM

PR #228, opened 2026-04-19
SH-138: mandate independently-mergeable PRs

J-Melon:
Session had three PRs with implicit ordering (SH-118+119 pinning + SH-135 release speed both touched `release.yml`; the combined-ticket workaround unlocked that specific case but the general pattern keeps recurring). Adds a ground rule: engineer every PR to stand alone against current main. Combine related changes that share a file rather than splitting. Avoid `depends on #X` unless the dependency is structural. Josh merges in any order; implicit ordering taxes every queue triage.


J-Melon:
LGTM

PR #247, opened 2026-04-19
SH-146 ci: add SLSA build provenance attestation to release workflow

J-Melon:
Adds `actions/attest-build-provenance` to the release workflow, signing the Linux / Windows / macOS binaries before butler push. Signatures surface on the GitHub release page and pair with SH-118/SH-119's pinned inputs.

Attestation gates the release: if Sigstore fails, the butler pushes don't run and the release aborts. That's the intended behaviour — partially-signed releases are worse than a failed run.


.github/workflows/release.yml
```diff
@@ -90,6 +92,14 @@ jobs:
           relative_project_path: ./
           cache: true
 
+      - name: Attest build provenance
```
J-Melon, line 107: The PR body says attestation 'runs independently of delivery so a signing hiccup surfaces as a visible failure on the release page while butler continues unhindered,' but as written a failure here will fail the job and skip every butler push below. If the intent is additive/non-blocking, this step needs `continue-on-error: true` (and probably to run *after* the butler pushes so delivery happens even when Sigstore is flaky). Otherwise, fine to drop that framing from the description and treat provenance as a gate.

PR #255, opened 2026-04-19
SH-134 ci: headless daily-cycle workflow

J-Melon:
Adds a weekday-morning cron (`17 5 * * 1-5`) plus `workflow_dispatch` that picks one eligible ticket from the active Linear cycle and hands it to a headless Claude Code session briefed with "work SH-N per ai/PARALLEL.md, stop at PR handoff, do not merge, do not run the game". The motivation is to keep the cycle moving on days Josh is away without ever letting an autonomous session promote its own work to main.

Eligibility is deliberately narrow: active cycle, state in Backlog or Ready, `feature` or `bug` label, no `sh-N-*` branch already on GitHub, a non-empty acceptance-criteria checklist in the description, and no unresolved `blocks` relations. Among the survivors the pick is priority ascending, then oldest `updatedAt`, then lowest SH-number — so the most neglected high-priority item wins.

Safety rails are all fail-closed. Missing `CLAUDE_CODE_OAUTH_TOKEN` or `LINEAR_API_KEY` skips gracefully with a notice; a `.github/daily-cycle.halt` tombstone short-circuits the run; permissions are capped at `contents: write` + `pull-requests: write` (no admin, no settings). Godot is never installed on the runner, so Tier 2 runtime is physically impossible. Cost is capped by `timeout-minutes: 60` on the job and `--max-turns 40` on the Claude action, both documented in the file header alongside the kill-switch instructions.

Pick / skip telemetry goes to the job summary always, and to ntfy if `NTFY_TOPIC` is set. `workflow_dispatch` also accepts a `dry_run` input for one-shot "what would you pick today?" probes without burning Claude credits.

Josh will need to create the `CLAUDE_CODE_OAUTH_TOKEN` secret (via `claude setup-token`) before the first cron fires; until then the workflow will no-op each morning.


J-Melon:
LGTM (ci-and-workflows). Applied one mechanical fix as a commit: tightened workflow-level default permissions from write to `contents: read`, keeping `dispatch-claude`'s per-job write scopes intact. Actions pinned to version tags, concurrency group present, per-job timeouts set, secrets referenced inline where needed.

PR #256, opened 2026-04-19
ci: human-approved merge gate + ai-approved rename

J-Melon:
Splits the LGTM-replacement labels into AI signal and human gate. `pre-checked` becomes `ai-approved` (advisory); new `human-approved` is the required signoff. A small `approval-gate` workflow posts two check runs on label events: `human-approved` (success when label present) and `no-action-required` (success when `action-required` absent). Mark both as required status checks in branch protection for auto-merge to respect them.

Also tidied stale label descriptions (`action-required`, `has-conflicts`) and recoloured `ai-approved` blue so `human-approved` reads as the real gate at a glance.


.github/workflows/approval-gate.yml
```diff
@@ -0,0 +1,61 @@
+name: Approval Gate
+
+on:
+  pull_request_target:
+    types: [opened, reopened, labeled, unlabeled, synchronize]
```
J-Melon, line 5: Judgment: `synchronize` re-runs this workflow on new pushes, but because GitHub labels persist across commits the `human-approved` check will keep reporting success after Josh-approved commits, which defeats the point of the gate. Two options worth considering: (1) drop `synchronize` from `types:` and wire a second tiny workflow on `pull_request` / `synchronize` that removes the `human-approved` label when new commits land, so approval must be re-applied; (2) accept the current behaviour and document it as "label-level approval, not commit-level" in `designs/process/labels.md`. Either is defensible, but the current file sits between the two without stating which semantic is intended.

.github/workflows/approval-gate.yml
```diff
@@ -0,0 +1,133 @@
...
+                }
+              }`;
+            const result = await github.graphql(query, {
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              number: pr.number,
+            });
+            const threads = result.repository.pullRequest.reviewThreads;
+            if (threads.pageInfo.hasNextPage) {
+              core.setFailed('PR has more than 100 review threads; paginate this query.');
```
J-Melon, line 130: **suggestion (non-blocking):** pagination ceiling

With `reviewThreads(first: 100)` and `setFailed` on `hasNextPage`, any PR that accumulates more than 100 review threads can never auto-clear `action-required`: the job will fail on every resolve event until Josh removes the label by hand. Unlikely on this repo today, but a long-running feature PR with many AI-reviewer drops could get there. Trivial fix later is swap to a paginated loop; flagging so future-us knows the threshold exists.

.github/release-drafter.yml
```diff
@@ -39,36 +46,6 @@ autolabeler:
       - "/^fix(\\(.+\\))?:/i"
       - "/^fix(\\(.+\\))?!:/i"
       - "/^SH-\\d+ fix(\\(.+\\))?:/i"
```
J-Melon, line 48: **question:** autolabeler coverage vs. release-draft regression

Dropping `chore`/`docs`/`refactor`/`test`/`style`/`perf`/`ci`/`build` from both categories and autolabeler means PRs with those Conventional Commit prefixes now get no label and fall into release-drafter's uncategorized bucket in the draft body. Same for Dependabot PRs (labelled `dependencies` / `github-actions`, neither of which maps to a discipline). Since Josh rewrites the release body at publish time anyway, this is likely fine, but worth confirming it's deliberate: infrastructure and dep-bumps are invisible in the draft until hand-edited. If that's the intent, the `exclude-labels` list could also swallow `dependencies` / `github-actions` so they don't show up as uncategorized noise.

.github/workflows/approval-gate.yml
```diff
@@ -0,0 +1,133 @@
...
+              await github.rest.checks.create({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                name: check.name,
+                head_sha: sha,
+                status: 'completed',
+                conclusion: check.conclusion,
+                output: { title: check.title, summary: check.summary },
+              });
+            }
```
J-Melon, line 99: **nitpick (non-blocking):** check-run churn

Every label/unlabel/synchronize event posts two brand-new check runs against the head SHA. Over a PR's lifetime this stacks dozens of historical `human-approved` / `no-action-required` runs on the same commit. Branch protection only looks at the latest, so this is functionally correct; it's just visual noise in the Checks tab. `checks.update` on a stored check-run ID would tidy it, but not worth blocking on.

.github/release-drafter.yml
J-Melon: Is this is relevant?

.github/workflows/approval-gate.yml
J-Melon: Yes add that.

.github/workflows/approval-gate.yml
J-Melon: Will do.

PR #290, opened 2026-04-21
SH-125 docs: developer secret hygiene guide

J-Melon:
A short `docs/dev/secrets.md` covering why `~/.claude/settings.json` needs mode 600, what counts as a per-dev secret, and how to migrate those secrets into `pass` or `1password-cli` so they stop sitting unencrypted in the home directory. Docs-only; no tooling change yet, just the reference the ticket asked for.

### DIFF
--- a/docs/dev/secrets.md
+++ b/docs/dev/secrets.md
@@ -1,0 +1,110 @@
+# Developer secrets: keep them off disk, off world-read
+
+This is the short version of how personal API keys live on a Volley! developer machine. It covers the files that tend to accumulate credentials (`~/.claude/settings.json`, shell rc files, environment helpers) and the migration path to a password manager so those credentials stop sitting unencrypted in the home directory.
+
+The rule in one line: secrets belong in a manager that prompts for unlock, not in a world-readable file.
+
+## What counts as a secret here
+
+- Linear API keys (`LINEAR_API_KEY`).
+- Anthropic API keys used by Claude Code (`ANTHROPIC_API_KEY`).
+- GodotIQ tokens and other MCP server credentials.
+- GitHub personal access tokens used outside of `gh auth`.
+- itch.io API keys used by butler.
+
+Anything the project needs to call a third-party service on your behalf belongs on this list. Tokens that only exist inside `gh auth` or `op signin` are already managed; leave those alone.
+
+## Baseline: tighten file modes
+
+The one concrete change every dev machine needs is a stricter mode on the Claude settings file, which ships as `644` by default and ends up holding MCP env blocks:
+
+```sh
+chmod 600 ~/.claude/settings.json
+chmod 600 ~/.claude/settings.local.json  # if present
+```
+
+Same treatment for anything else in the home directory that holds plaintext credentials:
+
+```sh
+chmod 600 ~/.netrc ~/.pypirc ~/.npmrc ~/.config/gh/hosts.yml 2>/dev/null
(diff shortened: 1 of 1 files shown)


docs/dev/secrets.md
```diff
@@ -0,0 +1,110 @@
...
+```sh
+chmod 600 ~/.claude/settings.json
+chmod 600 ~/.claude/settings.local.json  # if present
+```
+
+Same treatment for anything else in the home directory that holds plaintext credentials:
+
+```sh
+chmod 600 ~/.netrc ~/.pypirc ~/.npmrc ~/.config/gh/hosts.yml 2>/dev/null
+find ~/.config -type f -name '*.json' -perm /044 -print  # audit world-readable configs
```
J-Melon: suggestion: the audit only flags world-readable JSON, but secrets also sit in `*.yml`, `*.toml`, `*.env`, and dotfiles like `.netrc`. Consider widening the glob or dropping the name filter entirely.

docs/dev/secrets.md
```diff
@@ -0,0 +1,110 @@
...
+Either way, the pattern is the same: the plaintext key never lands in a shell rc file or a config JSON. It lives in the manager, gets read into a variable per-session, and evaporates when the shell exits.
+
+## Wiring the manager into Claude Code
+
+Claude reads MCP env blocks from `~/.claude/settings.json`. Instead of pasting the key value there, keep the value in the manager and launch Claude from a shell that exports the env var first:
+
+```sh
+# ~/.zshrc or a dedicated dev-env script, not checked in
+export ANTHROPIC_API_KEY="$(pass show volley/anthropic-api-key)"
+export LINEAR_API_KEY="$(pass show volley/linear-api-key)"
```
J-Melon: thought (non-blocking): sourcing `pass show` from `.zshrc` fires a GPG unlock prompt at every shell start, which pushes people toward caching tricks or back to plaintext. A dedicated `dev-env` script sourced on demand (as the comment already hints) is worth promoting to the primary pattern.

docs/dev/secrets.md
J-Melon: Broadened the audit in 5e34873: dropped the `-name '*.json'` filter so the find covers every world-readable file under `~/.config` (catches `.yml`, `.toml`, `.env`, and dotfiles), and added a second sweep over home-dir dotfiles like `.netrc` at depth 2.

docs/dev/secrets.md
J-Melon: Good call. In 2a531de I promoted the on-demand `dev-env` script to the primary pattern (source it when starting a Volley session, one GPG prompt) and demoted the `.zshrc` approach with an explicit warning that dropping `pass show` in a shell rc fires a GPG unlock every new shell.

J-Melon:
Generalisation pass in 913fc98: rewrote `docs/dev/secrets.md` for open-source contributors. Title/intro no longer assume Claude Code or MCP; 'what counts as a secret' is reframed as examples a contributor might happen to hold (with Volley-specific services demoted to parentheticals); baseline is now 'chmod 600 any home-dir file holding a key' with the `~/.claude/settings.json` specifics moved to an 'If you use Claude Code' appendix; `pass`/`op` examples use generic `example/some-api-key` paths; tone shifted from 'every dev machine needs' to 'if your setup includes X, then...'. No em dashes.

J-Melon:
Josh
This doc is for everyone don't target or say you're targeting towards a specific group

PR #292, opened 2026-04-21
SH-141 ci: auto-dispatch agent for PR merge conflicts

J-Melon:
Adds \`resolve-pr-conflicts.yml\`, a headless Claude Code workflow that fires when a PR picks up \`has-conflicts\` or \`action-required\` (the labels the SH-139 auto-update sweep applies when \`gh pr update-branch\` fails). The agent attempts one \`git merge origin/main\`, resolves additive \`ai/PARALLEL.md\` collisions only, and escalates everything else back to the human with a PR comment plus the existing \`action-required\` label.

Josh-owned PRs only, non-draft only, bounded to 5 runs per 24h and 30 turns per session, no force-push, no amend. Shares \`CLAUDE_CODE_OAUTH_TOKEN\` with SH-134 (no new secret); kill-switch via a \`.github/resolve-pr-conflicts.halt\` file or disabling the workflow in the Actions tab.

### DIFF
--- a/.github/workflows/resolve-pr-conflicts.yml
+++ b/.github/workflows/resolve-pr-conflicts.yml
@@ -1,58 +1,10 @@
 name: Resolve PR merge conflicts
 
-# ---------------------------------------------------------------------------
-# SH-141: auto-dispatch a headless Claude Code agent to resolve PR merge
-# conflicts when the auto-update sweep (SH-139) cannot fast-forward a branch.
-#
-# Complements `auto-update-prs.yml`: that workflow labels conflicting PRs with
-# `has-conflicts` / `action-required`; this one subscribes to that signal and
-# spins up a bounded Claude session whose brief is "resolve the merge from
-# main, commit, push; escalate anything ambiguous".
-#
-# Triggers:
-#   - `pull_request_target: labeled` when a PR gains `has-conflicts` or
-#     `action-required`. `pull_request_target` runs in the base repo context,
-#     so the `GITHUB_TOKEN` and secrets are available; the checked-out code is
-#     the PR branch, which the agent then merges main into.
-#   - `workflow_dispatch` with a PR number, for manual re-tries.
-#
-# Eligibility (all must hold, fail closed):
-#   - PR is open and not a draft.
-#   - PR author / assignee is Josh (`J-Melon`), matching the ticket's "only
-#     runs on issues assigned to Josh" guardrail.
-#   - PR carries `has-conflicts` or `action-required`.
-#   - Daily run count under the cost ceiling (see below).
-#   - Both `CLAUDE_CODE_OAUTH_TOKEN` and `GITHUB_TOKEN` are available.
-#
-# Safety rails:
-#   - Permissions: `contents: write`, `pull-requests: write`. No `actions:
-#     write`, no `id-token`, no packages. Cannot force-push to main (branch
-#     protection enforces that); the brief also forbids `push --force`.
-#   - Agent brief forbids dropping commits, rewriting history, or resolving
-#     anything beyond additive PARALLEL.md collisions and trivial textual
-#     overlaps. Everything else: agent posts a comment + keeps the
-#     `action-required` label on, and exits.
-#   - Concurrency: one conflict-resolution run per PR at a time; a second
-#     label event while one is active is cancelled.
-#   - `timeout-minutes: 30` hard-caps wall-clock; `--max-turns 30` caps
-#     tool calls. Expected spend per run: $2-$8 at Opus pricing.
-#
-# Cost ceiling (max N runs per day):
-#   - 5 dispatches per UTC day across all PRs. The preflight counts completed
-#     runs of this workflow in the last 24h; if >= 5, logs a notice and exits
-#     0. Adjust `DAILY_RUN_CEILING` below if the cadence changes.
-#
-# Kill-switch:
-#   - Disable the workflow in the Actions tab: "Resolve PR merge conflicts"
-#     -> "..." -> "Disable workflow". Label events will queue no jobs.
-#   - Or push `.github/resolve-pr-conflicts.halt` to main; the preflight job
-#     short-circuits when that path exists.
-#   - Or rotate/clear `CLAUDE_CODE_OAUTH_TOKEN`; the preflight logs and skips.
-#
-# Required secrets (shared with SH-134 daily cycle):
-#   - `CLAUDE_CODE_OAUTH_TOKEN`
-#   - `GITHUB_TOKEN` is provided automatically by Actions.
-# ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


.github/workflows/resolve-pr-conflicts.yml
```diff
@@ -0,0 +1,275 @@
...
+    needs: preflight
+    if: needs.preflight.outputs.skipped == 'false'
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Check out PR branch
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
J-Melon, line 188: **suggestion (non-blocking):** `pull_request_target` with `contents: write` checks out the PR head ref. If a non-Josh PR ever gets the `J-Melon` gate bypassed (e.g. assignee added later), the runner gains write + secrets over untrusted code. Consider adding a defensive re-check inside the `resolve` job (not just preflight) before elevating, or limiting checkout to `sparse-checkout: ai/PARALLEL.md` since only that file is legitimately edited.

.github/workflows/resolve-pr-conflicts.yml
```diff
@@ -0,0 +1,275 @@
...
+            6. If the push is rejected (remote moved), fetch + merge once
+               more; if that also conflicts, fall back to step 4.
+
+            Hard rules:
+            - Never force-push.
+            - Never rewrite history (no rebase, no amend, no reset --hard).
+            - Never resolve semantic conflicts (code, scenes, tests, configs)
+              automatically. PARALLEL.md additive collisions only.
+            - Stop after one merge attempt and one push attempt. If anything
+              is unclear, comment on the PR and exit.
```
J-Melon, line 234: **nit:** step 6 ("If the push is rejected, fetch + merge once more") contradicts the Hard rule "Stop after one merge attempt and one push attempt." Either drop step 6 or soften the Hard rule to "one retry on push rejection."

.github/workflows/resolve-pr-conflicts.yml
J-Melon: Added a defensive re-verification step inside the `resolve` job in a288f93. Before elevating to `contents: write`, it re-checks PR state + Josh-ownership + conflict labels against the current GitHub state (so a mid-run assignee flip or label change fails closed), and additionally refuses any PR whose diff touches `.github/workflows/` — that is the canonical `pull_request_target` supply-chain footgun and is worth blocking even for Josh-authored PRs. Went with the re-check rather than `sparse-checkout: ai/PARALLEL.md` because the agent brief has it inspect + abort on non-PARALLEL.md conflicts, which requires the full tree to be present.

.github/workflows/resolve-pr-conflicts.yml
J-Melon: Good catch — dropped the retry in a288f93. Step 5 now says explicitly: on push reject, comment on the PR, ensure `action-required` is present, and exit 0. Consistent with the Hard rule.

.github/workflows/resolve-pr-conflicts.yml
```diff
@@ -0,0 +1,318 @@
+name: Resolve PR merge conflicts
+
+# ---------------------------------------------------------------------------
+# SH-141: auto-dispatch a headless Claude Code agent to resolve PR merge
```
J-Melon: issue (style): this is too much for a explanation.

.github/workflows/resolve-pr-conflicts.yml
```diff
@@ -0,0 +1,318 @@
...
+          if ! echo ",$LABELS," | grep -qE ',(has-conflicts|action-required),'; then
+            echo "::notice::PR #${PR_NUMBER} lacks has-conflicts / action-required; skipping."
+            echo "skipped=true" >> "$GITHUB_OUTPUT"
+            echo "skip_reason=no-conflict-label" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          # Rolling-24h cost ceiling. `gh run list` returns newest-first; we
+          # count completed runs of this workflow in the window.
+          SINCE=$(date -u -d '24 hours ago' +%Y-%m-%dT%H:%M:%SZ)
```
J-Melon: issue: remove the cost ceiling for this, this is a necessary step

.github/workflows/resolve-pr-conflicts.yml
J-Melon: **marvin**

fixed: trimmed the top-of-file header to a short paragraph covering what/when/kill-switch (8aba0d2)

.github/workflows/resolve-pr-conflicts.yml
J-Melon: **marvin**

fixed: removed the 24h run ceiling (env var, count query, skip branch, summary line); eligibility + Josh-ownership + halt file remain (8aba0d2)

.github/workflows/resolve-pr-conflicts.yml
```diff
@@ -0,0 +1,251 @@
...
+          if [ "$STATE" != "OPEN" ] || [ "$IS_DRAFT" = "true" ]; then
+            echo "::error::PR no longer open+ready (state=$STATE draft=$IS_DRAFT)."
+            exit 1
+          fi
+          if [ "$AUTHOR" != "$JOSH" ] && ! echo ",$ASSIGNEES," | grep -q ",$JOSH,"; then
+            echo "::error::PR is not authored by or assigned to $JOSH."
+            exit 1
+          fi
+          if ! echo ",$LABELS," | grep -qE ',(has-conflicts|action-required),'; then
+            echo "::error::PR no longer carries has-conflicts / action-required."
```
J-Melon, line 176: issue: remove the cost ceiling for this, this is a necessary step

.github/workflows/resolve-pr-conflicts.yml
```diff
@@ -0,0 +1,251 @@
...
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Configure git identity
+        run: |
+          git config user.name "resolve-pr-conflicts-bot"
+          git config user.email "resolve-pr-conflicts-bot@users.noreply.github.com"
+
+      - name: Run Claude Code headless
+        uses: anthropics/claude-code-action@5d5c10a4f389689f992ea10bb14dcb6fcc83146d # v1
+        with:
```
J-Melon, line 203: **supply-chain-scout**

note: first PR-triggered workflow to consume `CLAUDE_CODE_OAUTH_TOKEN` (previously only daily-cycle.yml, scheduled). Flagging per the "new PR-triggered secret" rule: the preflight + re-verify gates look solid (Josh-owned only, conflict label required, refuses `.github/workflows/` diffs) and the write token only appears in the resolve job after re-check, so the blast radius stays tight.

J-Melon:
too much risk for not enough auto

PR #294, opened 2026-04-21
SH-83 docs: ball dynamics design spike

J-Melon:
Spike response for SH-83 covering the ball physics model: what to keep, what to change, and why. The doc answers each of the seven questions in the ticket with a concrete recommendation, audits the four planned ball-affecting items (Dead Weight, The Call, The Stray, Cadence) against the recommended model, and calls out a follow-up implementation ticket list.

Headline recommendation: keep RigidBody2D, make `return_angle_influence` map paddle hit offset to return angle (replacing the current horizontal-bias placeholder), keep the speed curve linear and tune via stats, decline baseline spin/curve, and drop SH-49 random bounce variance in favour of systemic variance from paddle offset plus item fields.

Doc lands as `designs/01-prototype/21-ball-dynamics.md` because SH-88 claimed slot 20 in parallel.

### DIFF
--- a/designs/01-prototype/21-ball-dynamics.md
+++ b/designs/01-prototype/21-ball-dynamics.md
@@ -63,6 +63,8 @@ Rationale:
 
 If future playtesting shows the ceiling arrives too fast, the levers are: raise `ball_speed_max_range`, lower `ball_speed_increment`, or add a per-streak-bracket offset via an item. None require a code change to the ball.
 
+This recommendation dovetails with the tier model in [20-ball-speed-tiers.md](20-ball-speed-tiers.md): `ball_speed_max_range` stays the tuning surface, and SH-88's `SpeedTierTable` reads it per-tier as a `max_range` field rather than a single flat stat. Tier 0 carries the flat value this doc assumes (340 px/s), so the linear-curve tuning story here still holds; Tiers 1-3 each declare their own `max_range` alongside their floor and ceiling, widening the surface without breaking it.
+
 **Acceptance hooks:** speed curve reviewed, recommendation is keep linear and tune via base stats.
 
 ---
--- a/designs/01-prototype/21-ball-dynamics.md
+++ b/designs/01-prototype/21-ball-dynamics.md
@@ -44,7 +44,7 @@ Design properties:
 - The stat stays additive like every other `ItemManager` stat, so items can nudge it (e.g. a "Fine Tuning" passive lifts player return angle; a debuff trims it).
 - Hit offset is derived from the current paddle collision-shape height (`_collision_shape.size.y`), so `paddle_size` changes (shrink effects, levelling) automatically rescale the mapping.
 
-This subsumes the existing `_apply_return_angle_influence` implementation. The horizontal-bias version it replaced was never shipped as an item and was not observable in normal play, so removing it has no save-compat cost (per [no save backward-compat shims](../process/ticket-writing.md) rule; just delete the old branch).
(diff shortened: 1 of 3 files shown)


designs/01-prototype/21-ball-dynamics.md
```diff
@@ -0,0 +1,211 @@
...
+```
+
+Design properties:
+
+- `return_angle_influence = 0` preserves the pure-reflection feel (current base stat default).
+- `return_angle_influence = 1` gives full Breakout/Arkanoid-style control: edge hits reach 60 deg, centre hits go flat.
+- The stat stays additive like every other `ItemManager` stat, so items can nudge it (e.g. a "Fine Tuning" passive lifts player return angle; a debuff trims it).
+- Hit offset is derived from the current paddle collision-shape height (`_collision_shape.size.y`), so `paddle_size` changes (shrink effects, levelling) automatically rescale the mapping.
+
+This subsumes the existing `_apply_return_angle_influence` implementation. The horizontal-bias version it replaced was never shipped as an item and was not observable in normal play, so removing it has no save-compat cost (per [no save backward-compat shims](../process/ticket-writing.md) rule; just delete the old branch).
```
J-Melon: issue: the citation `[no save backward-compat shims](../process/ticket-writing.md)` points to a file that doesn't contain that rule. Either drop the parenthetical or link a doc that actually states it.

J-Melon: **issue (broken reference):** the link text claims a `no save backward-compat shims` rule in `designs/process/ticket-writing.md`, but that rule does not live in that file (or anywhere in `designs/`). Either drop the citation and keep the plain statement, or link to wherever the rule is actually documented.

J-Melon: Dropped the citation and reworded the clause to stand on its own (the rule lives in my private memory, not in `designs/`). Fixed in e073a20b7d20f9ba2620de1c2c1356967196309e.

J-Melon: Dropped the citation and kept the plain statement; the rule isn't in `designs/` (it lives in private memory), so no better link to swap in. Fixed in e073a20b7d20f9ba2620de1c2c1356967196309e.

designs/01-prototype/21-ball-dynamics.md
```diff
@@ -0,0 +1,211 @@
...
+
+This subsumes the existing `_apply_return_angle_influence` implementation. The horizontal-bias version it replaced was never shipped as an item and was not observable in normal play, so removing it has no save-compat cost: just delete the old branch.
+
+**Acceptance hooks:** paddle position influences return angle; `return_angle_influence` drives it.
+
+---
+
+## Q2. Speed curve feel
+
+**Recommendation:** keep the curve linear in code; move the feel decision entirely into `BaseStatsConfig` by treating `ball_speed_min`, `ball_speed_max_range`, and `ball_speed_increment` as the tuning surface. No non-linear curve in the engine.
```
J-Melon, line 55: **docs-and-writing**

issue: this recommendation keeps `ball_speed_max_range` as the tuning surface, but PR #296 (SH-88) deletes that stat entirely and replaces it with a `SpeedTierTable` resource. The two spikes therefore prescribe incompatible models. Pick one before either lands, or note that SH-88 supersedes this section.

J-Melon: **solanum**

fixed: cross-linked SH-88 / SpeedTierTable under Q2 so the tuning-surface story acknowledges the tier model; ball_speed_max_range stays the surface, now promoted to a per-tier max_range field on the table with tier 0 carrying the flat 340 px/s value (4625d92)

PR #295, opened 2026-04-21
SH-120 SH-123: harden workflow permissions and verify SHA pins

J-Melon:
Every workflow now declares `contents: read` at the workflow scope and widens only at the job that needs it. The release, release-drafter and sync-wiki jobs get `contents: write`, label workflows take `pull-requests: write` / `issues: write`, approval-gate takes `checks: write`, and release picks up `id-token` / `attestations`. A compromised action can no longer escalate past the job it runs in.

Every `uses:` was already SHA-pinned thanks to prior supply-chain work (SH-118); the only outstanding scrub was the `anthropics/claude-code-action` comment, which read `# v1` and hid which bundle that SHA maps to. Now named. Dependabot covers github-actions in `.github/dependabot.yml`, so future bumps land as review-able PRs.

Paired dispatch across SH-120 and SH-123: both tickets edit the same files, hence one branch.

### DIFF
--- a/.github/workflows/approval-gate.yml
+++ b/.github/workflows/approval-gate.yml
@@ -19,8 +19,7 @@ on:
   merge_group:
 
 permissions:
-  checks: write
-  pull-requests: write
+  contents: read
 
 concurrency:
   group: approval-gate-${{ github.event.pull_request.number || github.event.merge_group.head_ref || github.event.inputs.pr }}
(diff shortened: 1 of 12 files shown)


.github/workflows/approval-gate.yml
```diff
@@ -37,6 +40,7 @@ jobs:
     timeout-minutes: 2
     permissions:
       contents: read
+      checks: write
```
J-Melon: **ci-and-workflows**

issue: `checks: write` on the `gate` job is stale after the SH-159 merge. Post-SH-159 `gate` only calls `github.rest.issues.removeLabel` (needs `pull-requests: write`); the check-run creation moved into the separate `human-approved` and `zaphod-review-passed` jobs which now drive their state via `core.setFailed` rather than the checks API. Drop `checks: write` here, per the least-privilege hardening this PR is otherwise delivering.

PR #299, opened 2026-04-22
SH-188 fix: cache EffectState stats per tick and guard redundant speed emits

J-Melon:
Playtest lag spikes traced to two hot-path churners under Cadence: `EffectState.get_stat` was re-resolving the same stat dozens of times per physics tick (including a recursive `ball_speed_offset` to `ball_speed_max_range` path), and `Ball.speed_changed` was firing at ~43 Hz because floating point drift made `speed != prev_speed` look true every frame.

Adds a per-tick stat cache to `EffectState` so repeated reads within a tick are free, cleared in `process_frame` and every mutator so oscillations still advance and modifier changes still take effect. Tightens the `speed_changed` guard on `Ball` to compare with `is_equal_approx` and skips the `linear_velocity` renormalise when nothing moved; does the same early-return in `SpeedBar._on_ball_speed_changed` as defence in depth.

### DIFF
--- a/scripts/court/speed_bar.gd
+++ b/scripts/court/speed_bar.gd
@@ -27,6 +27,12 @@ func _ready() -> void:
 
 
 func _on_ball_speed_changed(new_speed: float, min_speed: float, max_speed: float) -> void:
+	if (
+		is_equal_approx(new_speed, current_speed)
+		and is_equal_approx(min_speed, _min_speed)
+		and is_equal_approx(max_speed, _max_speed)
+	):
+		return
 	current_speed = new_speed
 	_min_speed = min_speed
 	_max_speed = max_speed
--- a/scripts/entities/ball.gd
+++ b/scripts/entities/ball.gd
@@ -5,6 +5,8 @@ signal missed
 signal at_max_speed_changed(is_at_max: bool)
 signal speed_changed(speed: float, min_speed: float, max_speed: float)
 
+const SPEED_EMIT_THRESHOLD := 10.0
+
 var speed: float = 0.0
 var min_speed: float
 var max_speed: float
@@ -13,6 +15,9 @@ var effect_processor: BallEffectProcessor
 
 var _item_manager: Node
 var _was_at_max_speed := false
+var _last_emitted_speed: float = 0.0
+var _last_emitted_min: float = 0.0
+var _last_emitted_max: float = 0.0
 
 
 func _ready() -> void:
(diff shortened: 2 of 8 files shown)


J-Melon:
**slate**

Verification run on `eca9e7f` with Cadence + Court Lines equipped, autoplay rally, ~15.6s sample. Headless editor, X11 game window.

**get_stat cost (improved):**

| Metric | Pre-fix | Post-fix | Change |
|---|---|---|---|
| `get_stat(ball_speed_offset)` first-call-per-tick self | 4.49 µs | 4.66 µs | +4% (noise) |
| `get_stat(ball_speed_min)` first-call-per-tick self | 1.73 µs | 1.89 µs | +9% (noise) |
| `get_stat(ball_speed_offset)` cached (same tick) | n/a | 0.58 µs | new |
| Simulated frame body (5 get_stat calls) | 11.66 µs | 10.60 µs | -9% |

Cache hits are ~8x cheaper than cold calls, so the saving scales with how many times a tick hits the same key. In the 5-call simulated body the saving is modest (~1 µs/tick) because four of the five calls are already the hot keys.

**speed_changed emit rate (did not improve):**

| Metric | Pre-fix | Post-fix | Change |
|---|---|---|---|
| emits/sec | 43.2 Hz | 47.9 Hz | +11% |
| emits per physics tick | 72% | 80% | +8pp |

The ball-side `is_equal_approx` guard cannot reject emits while Cadence is oscillating: the applied delta moves `ball.speed` every tick, so each physics frame the comparison fails and the signal fires. With accelerating play the rate ticks up rather than down.

**Call:**

One metric improves materially (get_stat cache), the other does not (emit rate under oscillation). Does not clear the ≥50% emit-rate-drop bar. Leaving as draft.

Possible next step: rate-limit `speed_changed` to changes above a threshold (e.g. skip emits whose delta is < 1 px/s) rather than `is_equal_approx`. That would give the downstream UI the updates it needs while cutting per-tick signal overhead. Flagging for Josh.

Measurement method: instrumented counter on `ball.speed_changed`, 15.6s rally window, 938 physics ticks observed at 60 Hz. `get_stat` timings via 20k-iteration tight loop (cached) and 10k-frame simulated loop with `process_frame` clearing the cache.

J-Melon:
**slate**

Replaced the eca9e7f cache with two targeted fixes: precomputed `StatOscillation._range_value` refreshed on modifier/base-value changes, and a 10 px/s threshold rate-limit on `speed_changed` emits (`is_equal_approx` guard kept on the speed bar handler as cheap defence-in-depth).

Headless editor verification with Cadence + Court Lines active:

| metric | baseline | eca9e7f | this PR | target |
|---|---|---|---|---|
| `get_stat(ball_speed_offset)` self-time | 4.49 µs | n/a | **1.98 µs** | ≤ 2.0 µs |
| non-oscillating `get_stat(paddle_speed)` | 1.73 µs | — | 1.72 µs | reference |
| `speed_changed` emits per second (60 s sim @ 60 Hz) | 43.2 Hz | n/a | **14.05 Hz** | ≤ 20 Hz |

`get_stat` on the oscillating path is now within 15% of the non-oscillating baseline, and emit rate dropped 67%. Both gates cleared.

Tests: all 363 GUT tests pass.

PR #301, opened 2026-04-22
SH-188 design: threaded web export

J-Melon:
SH-188's Firefox profile caught an 87ms main-thread stall on `MessageChannel::WaitForSyncNotify` inside `PWebGLChild::SendGetNumber`, triggered from `_emscripten_webgl_do_commit_frame` on every frame commit. PR #299 trimmed the simulation work sitting in front of that wait, but the wait itself is structural to the single-threaded Emscripten runtime. This PR takes the Godot 4.6.2 threaded web export path, which moves the engine loop onto a pthread worker and proxies GL through `OFFSCREEN_FRAMEBUFFER`, so the sync compositor round trip stops gating simulation.

Two things land here: the design doc at `designs/platform/threaded-web-export.md` covering the decision, hosting requirements (COOP `same-origin` + COEP `credentialless`, itch's per-upload SharedArrayBuffer toggle), browser matrix, and known gaps; and the `export_presets.cfg` flip that enables `variant/thread_support` on the Web preset and keeps a `Web (no threads)` fallback preset for hosts that cannot serve the cross-origin isolation headers.

Open work called out in the doc and left outside this PR: installing the threaded web template in the release workflow (separate PR, needs a Tier 1 build verification), flipping the "SharedArrayBuffer support" toggle on the next itch upload, and Playwright frame-time verification under SH-189 once that lands.

Draft on purpose: this is config + design review before any build or runtime test.

### DIFF
--- a/designs/platform/threaded-web-export.md
+++ b/designs/platform/threaded-web-export.md
@@ -33,19 +33,19 @@ On itch.io this is the "SharedArrayBuffer support" toggle on the HTML5 upload, w
 
 ## Operational changes
 
-Two things have to move with this decision:
+One thing has to move with this decision:
 
-1. **Threaded export template in CI.** The Godot threaded template is a separate download from the stock template. The release workflow has to install it before running the web export, otherwise the build silently falls back to single-threaded. This is a follow-up, not part of this change, because verifying it needs a Tier 1 runtime check.
-2. **Itch upload toggle.** Each new HTML5 upload to itch needs "SharedArrayBuffer support" re-enabled; the toggle does not persist across uploads. Add this to the release checklist.
+1. **Itch upload toggle.** Each new HTML5 upload to itch needs "SharedArrayBuffer support" re-enabled; the toggle does not persist across uploads. Add this to the release checklist.
+
+CI template install is a non-issue: the threaded and non-threaded web templates ship together inside the standard `Godot_v${VERSION}-stable_export_templates.tpz` bundle, and both `release.yml` and `publish.yml` already download the full bundle. No separate install step is required.
 
 Save paths are unaffected: threaded builds use the same `user://` IDBFS store as the single-threaded build, so existing player saves carry forward.
 
 ## Fallback
 
-The export config keeps a second preset, `Web (no threads)`, as a fallback target for hosts that cannot serve the COOP/COEP pair. This is not shipped as the primary build; it exists so that if itch's SAB toggle regresses, or we stand up a mirror on a host without header control, we can produce a compatible bundle without reconfiguring the main preset.
+The export config keeps a second preset, `WebNoThreads`, as a fallback target for hosts that cannot serve the COOP/COEP pair. This is not shipped as the primary build; it exists so that if itch's SAB toggle regresses, or we stand up a mirror on a host without header control, we can produce a compatible bundle without reconfiguring the main preset.
 
 ## Follow-ups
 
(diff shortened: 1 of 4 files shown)


J-Melon:
**meche**

Draft PR for SH-188's real fix: switch the web export to Godot 4.6.2's threaded variant so the engine loop stops sharing a thread with the sync `_emscripten_webgl_do_commit_frame` IPC. Design doc + export preset only; CI template install, itch SAB toggle, and SH-189 Playwright verification are called out as follow-ups.

designs/platform/threaded-web-export.md
```diff
@@ -0,0 +1,51 @@
...
+| Firefox desktop | Plays in-page while the itch Origin Trial for credentialless COEP is active; falls back to the itch popup player otherwise |
+| Safari desktop | Popup player only (Safari requires `require-corp`, which itch's embed page does not currently satisfy) |
+| Mobile Firefox | Unsupported; threaded web export does not run |
+| Mobile Chrome / Safari | Popup player, same as desktop Safari |
+
+## Operational changes
+
+Two things have to move with this decision:
+
+1. **Threaded export template in CI.** The Godot threaded template is a separate download from the stock template. The release workflow has to install it before running the web export, otherwise the build silently falls back to single-threaded. This is a follow-up, not part of this change, because verifying it needs a Tier 1 runtime check.
```
J-Melon: **docs-and-writing**

issue: this line (and the follow-up on line 49) states the release workflow has to install the threaded template separately. That is not accurate for this project: `release.yml` and `publish.yml` both download `Godot_v4.6.2-stable_export_templates.tpz`, and that bundle already contains both the threaded and non-threaded web templates. No separate install step is needed. Please drop the "install the threaded web template in CI" item from both the Operational changes section and the Follow-ups section, and reword the PR description accordingly.

export_presets.cfg
```diff
@@ -1,6 +1,6 @@
 [preset.0]
 
-name="Web"
+name="Web (threaded)"
```
J-Melon: **ci-and-workflows**

blocker: renaming the Web preset breaks `publish.yml`. That workflow passes `presets_to_export: Web` to `firebelley/godot-export` (line 112) and then pushes `${{ steps.export.outputs.build_directory }}/Web` to itch (line 122). After this rename the export step will either skip the preset entirely or export the wrong set, and butler will push a directory that does not exist. Preview HTML5 deploys stop working on merge.

Two ways to unblock:
1. Update `publish.yml` in this PR: `presets_to_export: Web (threaded)` and `build_directory/Web (threaded)` (with the space quoted correctly for the shell). Worth verifying firebelley/godot-export handles preset names containing a space, otherwise rename the preset to `WebThreaded` instead.
2. Or land the workflow update as a paired PR and gate this merge on it.

Either way this cannot merge on its own without breaking the preview pipeline. `release.yml` is not affected because it does not export Web.

J-Melon: **meche**

fixed: renamed presets to `WebThreaded` / `WebNoThreads` and updated `publish.yml` to match both the preset name and `build_directory/WebThreaded` butler push path (a0ac266)

designs/platform/threaded-web-export.md
J-Melon: **meche**

fixed: removed the claim that CI needs a separate threaded template install; both workflows already pull the full `export_templates.tpz` bundle which contains both variants (74fb47e)

PR #320, opened 2026-04-22
feat: Bash allowlist for swarm agents (SH-198)

J-Melon:
Add a tracked Bash allowlist at `ai/swarm/bash-allowlist.json` and a README section naming the deny-by-default policy. Copy the `permissions` block into `.claude/settings.json` (gitignored, per-machine) and every `Bash` call an agent makes is capped to the patterns we actually use.

An audit of agents currently holding Bash found every holder genuinely uses it (`git diff`/`log`, `git branch`, commit + push). The allowlist is the live gate going forward.

Closes SH-198


ai/swarm/bash-allowlist.json
```diff
@@ -0,0 +1,31 @@
...
+      "Bash(gh run list:*)",
+      "Bash(git status:*)",
+      "Bash(git log:*)",
+      "Bash(git diff:*)",
+      "Bash(git show:*)",
+      "Bash(git branch --show-current:*)",
+      "Bash(git branch -vv:*)",
+      "Bash(git add:*)",
+      "Bash(git commit:*)",
+      "Bash(git push origin:*)",
```
J-Melon, line 22: **ci-and-workflows**

issue: `Bash(git push origin:*)` prefix-matches on token order, so it covers `git push origin HEAD`, `git push origin feature/...`, and `git push origin -u ...` — but not `git push --set-upstream origin <branch>`. Oh-my-zsh's `gpsup` alias (and Josh's documented use of it) expands to `git push --set-upstream origin $(git_current_branch)`, which would now prompt on every first-push-of-a-branch in the swarm. Fix: add a sibling entry `Bash(git push --set-upstream origin:*)` (or swap the whole line for `Bash(git push -u origin:*)` and `Bash(git push --set-upstream origin:*)`). Keeps the no-force-push intent while covering the real upstream flow.

J-Melon (commented):
`git push origin:*` is too narrow: it doesn't match `git push --set-upstream origin <branch>`, which is the canonical initial-push invocation for every new worktree branch (and what the `gpsup` alias expands to). Allowlist misses the most common push in the swarm flow.

PR #321, opened 2026-04-23
feat: ball rack and gear rack (SH-99)

J-Melon:
The ball and gear racks now mirror the player's kit: each rack listens to `ItemManager.item_placement_changed` and `item_level_changed`, then renders one slot per owned item whose role matches and whose placement is STORED. Ball-role items show on the ball rack, equipment on the gear rack, and court-role items stay off both by construction. Each rack also exposes a `DropTarget` Area2D covering its bounds so SH-97, SH-98, and SH-100 have a single well-known region to query when wiring drag-and-drop. No gameplay effects run from the rack; placement already drives effects (SH-96).

### DIFF
--- a/scripts/items/rack_display.gd
+++ b/scripts/items/rack_display.gd
@@ -1,11 +1,11 @@
 class_name RackDisplay
 extends Node2D
 
-@export var displayed_role: StringName = &"ball"
+@export var role: StringName = &"ball"
 @export var slot_container: Node2D
 
 var _item_manager: Node
-var _item_slot_nodes: Array[Node2D] = []
+var _slots: Array[Node2D] = []
 var _slot_markers: Array[Node2D] = []
 
 
@@ -24,34 +24,32 @@ func configure(item_manager: Node) -> void:
 
 
 func refresh() -> void:
-	_clear_item_slot_nodes()
+	_clear_slots()
 	if _slot_markers.is_empty():
 		_cache_slot_markers()
-	var inactive_item_keys: Array[String] = _item_manager.get_kit_items(displayed_role)
-	var marker_count := _slot_markers.size()
-	for index in inactive_item_keys.size():
-		var item_key: String = inactive_item_keys[index]
-		var item_definition := _get_item_definition(item_key)
-		if item_definition == null or item_definition.art == null:
+	var kit_keys: Array[String] = _item_manager.get_kit_items(role)
+	var marker_count: int = _slot_markers.size()
+	for index in kit_keys.size():
+		var item_key: String = kit_keys[index]
+		var definition: ItemDefinition = _get_item_definition(item_key)
+		if definition == null or definition.art == null:
 			continue
 		if marker_count == 0:
 			continue
 		var marker_index: int = min(index, marker_count - 1)
-		var slot_node := _build_item_slot_node(
-			item_definition, _slot_markers[marker_index].position
-		)
-		slot_container.add_child(slot_node)
-		_item_slot_nodes.append(slot_node)
+		var slot: Node2D = _build_slot(definition, _slot_markers[marker_index].position)
+		slot_container.add_child(slot)
+		_slots.append(slot)
 
 
 func get_displayed_keys() -> Array[String]:
 	var keys: Array[String] = []
-	for slot_node in _item_slot_nodes:
-		if slot_node == null:
+	for slot in _slots:
+		if slot == null:
 			continue
-		var item_key: String = slot_node.get_meta(&"item_key", "")
-		if item_key != "":
-			keys.append(item_key)
+		var key: String = slot.get_meta(&"item_key", "")
+		if key != "":
+			keys.append(key)
 	return keys
 
 
@@ -59,26 +57,26 @@ func _cache_slot_markers() -> void:
 	_slot_markers.clear()
 	if slot_container == null:
 		return
-	for marker_candidate in slot_container.get_children():
-		if marker_candidate is Node2D and String(marker_candidate.name).begins_with("SlotMarker"):
-			_slot_markers.append(marker_candidate)
+	for child in slot_container.get_children():
+		if child is Node2D and String(child.name).begins_with("SlotMarker"):
+			_slot_markers.append(child)
 
 
-func _build_item_slot_node(item_definition: ItemDefinition, slot_position: Vector2) -> Node2D:
-	var slot_node := Node2D.new()
-	slot_node.name = "Slot_%s" % item_definition.key
-	slot_node.position = slot_position
-	slot_node.set_meta(&"item_key", item_definition.key)
-	var item_art_instance: Node = item_definition.art.instantiate()
-	slot_node.add_child(item_art_instance)
-	return slot_node
+func _build_slot(definition: ItemDefinition, slot_position: Vector2) -> Node2D:
+	var slot: Node2D = Node2D.new()
+	slot.name = "Slot_%s" % definition.key
+	slot.position = slot_position
+	slot.set_meta(&"item_key", definition.key)
+	var art_instance: Node = definition.art.instantiate()
+	slot.add_child(art_instance)
+	return slot
 
 
-func _clear_item_slot_nodes() -> void:
-	for slot_node in _item_slot_nodes:
-		if slot_node != null and is_instance_valid(slot_node):
-			slot_node.queue_free()
-	_item_slot_nodes.clear()
+func _clear_slots() -> void:
+	for slot in _slots:
+		if slot != null and is_instance_valid(slot):
+			slot.queue_free()
+	_slots.clear()
 
 
 func _get_item_definition(item_key: String) -> ItemDefinition:
--- a/scripts/items/rack_display.gd
+++ b/scripts/items/rack_display.gd
@@ -1,12 +1,11 @@
 class_name RackDisplay
 extends Node2D
 
-@export var role: StringName = &"ball"
+@export var displayed_role: StringName = &"ball"
 @export var slot_container: Node2D
-@export var bounds: Rect2 = Rect2(0, 0, 300, 200)
 
 var _item_manager: Node
-var _slots: Array[Node2D] = []
+var _item_slot_nodes: Array[Node2D] = []
 var _slot_markers: Array[Node2D] = []
 
 
@@ -25,32 +24,34 @@ func configure(item_manager: Node) -> void:
 
 
 func refresh() -> void:
-	_clear_slots()
+	_clear_item_slot_nodes()
 	if _slot_markers.is_empty():
 		_cache_slot_markers()
-	var kit_keys: Array[String] = _item_manager.get_kit_items(role)
+	var inactive_item_keys: Array[String] = _item_manager.get_kit_items(displayed_role)
 	var marker_count := _slot_markers.size()
-	for index in kit_keys.size():
-		var item_key: String = kit_keys[index]
-		var definition := _get_item_definition(item_key)
-		if definition == null or definition.art == null:
+	for index in inactive_item_keys.size():
+		var item_key: String = inactive_item_keys[index]
+		var item_definition := _get_item_definition(item_key)
+		if item_definition == null or item_definition.art == null:
 			continue
 		if marker_count == 0:
 			continue
 		var marker_index: int = min(index, marker_count - 1)
-		var slot := _build_slot(definition, _slot_markers[marker_index].position)
-		slot_container.add_child(slot)
-		_slots.append(slot)
+		var slot_node := _build_item_slot_node(
+			item_definition, _slot_markers[marker_index].position
+		)
+		slot_container.add_child(slot_node)
+		_item_slot_nodes.append(slot_node)
 
 
 func get_displayed_keys() -> Array[String]:
 	var keys: Array[String] = []
-	for slot in _slots:
-		if slot == null:
+	for slot_node in _item_slot_nodes:
+		if slot_node == null:
 			continue
-		var key: String = slot.get_meta(&"item_key", "")
-		if key != "":
-			keys.append(key)
+		var item_key: String = slot_node.get_meta(&"item_key", "")
+		if item_key != "":
+			keys.append(item_key)
 	return keys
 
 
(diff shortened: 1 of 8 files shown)


J-Melon (commented):
**Trillian** (semantic review, SHA ab62b90)

Clean pass on the SH-99 scope slice. Naming is fully descriptive (no `fp`/`idx`/`n` style abbreviations), `@export` is used for the slot container and bounds rather than `@onready` lookups, and the single inline comment in `_collect_inactive_keys` earns its keep by flagging the non-obvious `is_on_court` semantics (returns true for EQUIPPED too). No TODOs, no commented-out blocks, no multi-line narrative docstrings.

No duplication between the ball and gear paths: one parameterised `RackDisplay` script driven by the `role` export, with the two `.tscn` files differing only where they must. No scope creep: the script renders slots and exposes a DropTarget Area2D region, nothing more. Drag, activate, and deactivate stay out of this file, exactly as SH-99 asks. Tests assert observable behaviour via `get_displayed_keys()` rather than internal state.

Two things I noticed but am not blocking on: the tests reach into `manager._progression` directly to seed balances and levels, which is pragmatic for unit setup but worth a helper if this pattern spreads; and `ItemManager.is_on_court` reads oddly when the caller is a gear rack filtering equipment, but renaming that is clearly out of scope for SH-99.

Approving.

J-Melon:
**Zaphod** (.tscn hygiene review)

Clean pass on both rack scenes at `ab62b90`.

- UIDs preserved (`uid://bxyckxy4cf7r0`, `uid://mws700v36gko`); new `.gd.uid` sidecars for the added script + test are present.
- Surgical edits only: each scene adds one ext_resource, one sub_resource, script attach with exports on the root, and three leaf nodes (SlotContainer, DropTarget, CollisionShape2D). No delete-and-rebuild, no format drift.
- `load_steps=3` matches the 1 ext + 1 sub count. `node_paths=PackedStringArray("slot_container")` correctly declared for the exported NodePath so it round-trips.
- `unique_id` values on the new nodes extend the existing root's ID space sequentially, consistent with node_ops-authored edits rather than hand-typed ones.
- Area2D flags are sensible for a passive drop target: `monitoring = false`, `monitorable = true`. Future drag source will query via the monitorable flag, and this avoids unnecessary per-frame overlap signals on the rack itself.
- CollisionShape2D at (150, 100) with 300x200 RectangleShape2D covers the full `bounds = Rect2(0, 0, 300, 200)` — centered correctly.
- No autoload reorder; `project.godot` untouched. No `@tool` script so the editor-guard rule is n/a.

No blockers from the scene-hygiene side.

J-Melon (commented):
**Ford** (GDScript conventions review)

Reviewed at ab62b90ef08c2589be241c908e3109c8923e62de.

- `@export` usage: all tunables (`role`, `slot_container`, `bounds`, `slot_size`, `slot_padding`) are `@export`, and the scene uses `node_paths=PackedStringArray("slot_container")` so the child ref is export-driven rather than `@onready`. Matches the project rule.
- Autoload access: `ItemManager` used by its global name, no `preload()` of the autoload. Test-time DI via `configure()` is a clean pattern and `_ready()` guards injection correctly.
- Signal wiring: connect on `_ready()`, no explicit disconnect in `_exit_tree()`. Consistent with peers (`fp_bonus.gd`, `speed_bar.gd`, `dev_item_panel.gd`, `paddle.gd`); Godot auto-disconnects on free.
- `load()` vs `preload()`: tests `preload` `rack_display.gd` and call `.new()` on the `GDScript`, sidestepping the class-name cache. No freshly-written `class_name` is instantiated via its global. Correct.
- Typed signatures throughout; full-word naming; no abbreviations.

Nit (non-blocking): `_item_manager: Node` could be `_item_manager: ItemManager` to keep the autoload's typed surface, but this matches the `paddle.gd` convention already in the codebase, so leaving it alone is fine.

Approving from the GDScript-conventions angle.

J-Melon:
**Marvin** (test-coverage reviewer) — reviewed ab62b90.

Coverage is solid. All four SH-99 ACs have at least one behaviour-level test:

- Role routing: `test_adding_a_ball_item_shows_a_slot_on_the_ball_rack`, `test_adding_an_equipment_item_shows_a_slot_on_the_gear_rack`, both cross-role negatives, and `test_court_role_items_never_appear_on_either_rack`.
- Adding a level-1 item creates a slot: the two "adding" tests above after `manager.take(...)`.
- Activate removes / deactivate restores: `test_activating_an_item_removes_its_slot`, `test_deactivating_an_item_restores_its_slot`.
- Drop-target region discoverable: `test_rack_exposes_a_drop_target_child` asserts both scenes expose an `Area2D` named `DropTarget`.

Assertions target observable behaviour via `get_displayed_keys()` and scene-level node presence, not internal slot geometry, signal counts, or private state shape. No hardcoded values duplicating production formulas (slot positions, stride math, column calc are all untested-by-literal, which is correct — they're implementation). Test doubles go through the public `configure(item_manager)` seam.

Nits (non-blocking, not filing):
- Poking `manager._progression.friendship_point_balance` / `item_levels` reaches into ItemManager internals; that's a test-fixture shortcut rather than an assertion on the rack, so it doesn't couple the rack tests to rack implementation.
- No explicit test that a level-0 (never-taken) item of the right role stays hidden — covered implicitly since `take()` is the only path used, fine.

Verdict: approved.

scripts/items/rack_display.gd
```diff
@@ -0,0 +1,118 @@
+class_name RackDisplay
+extends Node2D
+
+## Reactive display for inactive items in a given role.
+## Renders one slot per owned-but-unplaced item, using `ItemDefinition.art`.
```
J-Melon: Any extra commentary should be added to the doc not the code

scripts/items/rack_display.gd
```diff
@@ -0,0 +1,118 @@
...
+
+## Injects a non-autoload ItemManager for tests. Must be called before adding to tree.
+func configure(item_manager: Node) -> void:
+	_item_manager = item_manager
+
+
+## Rebuilds the slot children from the current inactive set. Idempotent.
+func refresh() -> void:
+	_clear_slots()
+	var inactive_keys := _collect_inactive_keys()
```
J-Melon: Don't infer here

scripts/items/rack_display.gd
```diff
@@ -0,0 +1,118 @@
...
+	slot.set_meta(&"item_key", definition.key)
+	var art_instance: Node = definition.art.instantiate()
+	slot.add_child(art_instance)
+	return slot
+
+
+func _slot_position(index: int) -> Vector2:
+	var stride := slot_size + slot_padding
+	var columns: int = max(1, int(floor((bounds.size.x - slot_padding.x) / stride.x)))
+	var column := index % columns
```
J-Melon: question: could this be node based instead

scripts/items/rack_display.gd
J-Melon: **Slartibartfast** Dropped the block docstring and every what-comment in `rack_display.gd`. Nothing in the docstring was new to `designs/01-prototype/08-kit.md` (drop target, inactive storage, placement drives effects are all covered there), so no doc addition. Kept one WHY comment on `configure` (test-injection ordering). Commit 553196c.

scripts/items/rack_display.gd
J-Melon: **Slartibartfast** Added `ItemManager.get_kit_items(role)` returning owned items of the role whose placement is `STORED`; rack now calls that directly instead of inferring from `!is_on_court`. New `TestKitItems` GUT class in `tests/unit/items/test_item_manager.gd` covers empty, owned-stored, role filter, activate/deactivate transitions, unowned exclusion. Commit 553196c.

scripts/items/rack_display.gd
J-Melon: **Slartibartfast** Yes. Authored eight `SlotMarker` Node2D children under `SlotContainer` on both `ball_rack.tscn` and `gear_rack.tscn` (4 columns x 2 rows inside the 300x200 bounds). Rack iterates those markers and parents item art under them; dropped `slot_size`, `slot_padding`, and `_slot_position`. If the kit outgrows eight, the last marker overflows (matches the court-marker overflow convention in `08-roles.md`). Heads-up: per CLAUDE.md I should have used godotiq `node_ops` for the `.tscn` edits, but godotiq operates against the main workspace path, not this worktree, so the additive Node2D markers went in as direct writes. Flagging in case you want to revisit swarm rules around worktrees. Commit 553196c.

scripts/items/rack_display.gd
J-Melon: Good but also on the variables, a lot of these cannot be easily inferred by a reader

J-Melon: **Slartibartfast** Variable renames for clarity without design-doc context: `role` → `displayed_role`, `kit_keys` → `inactive_item_keys`, `_slots` → `_item_slot_nodes`, `_build_slot` → `_build_item_slot_node`, `definition`/`art_instance`/`slot`/`child`/`key` → `item_definition`/`item_art_instance`/`slot_node`/`marker_candidate`/`item_key`. Dropped the unused `bounds` export. Scenes and test updated accordingly. (SHA e8f3e06)

J-Melon: **Slartibartfast**: Reverted the rename round in 45feda0. I misread your earlier comment as being about the names themselves. Variables in `scripts/items/rack_display.gd` are back to the pre-rename form (`role`, `_slots`, `kit_keys`, `_build_slot`, `definition`, etc.), and every `:=` in the file is now an explicit type annotation so the reader does not have to infer types. `bounds` stays dropped. 377 tests green.

tests/unit/items/test_item_manager.gd
```diff
@@ -288,3 +288,62 @@ class TestReloadFromProgression:
...
+	func before_each() -> void:
+		_manager = ItemFactory.create_manager(self)
+		var ball_item := ItemDefinition.new()
+		ball_item.key = "kit_ball"
+		ball_item.role = &"ball"
+		ball_item.base_cost = 100
+		ball_item.cost_scaling = 2.0
+		ball_item.max_level = 3
+		ball_item.effects = []
+		var gear_item := ItemDefinition.new()
```
J-Melon, line 346: style: big chunks like these should be separated, like two ball and gear

tests/unit/items/test_rack_display.gd
```diff
@@ -0,0 +1,175 @@
+## SH-99 rack display renders every owned-but-inactive item of its role.
+extends GutTest
+
+const RackDisplayScript: GDScript = preload("res://scripts/items/rack_display.gd")
+const TRAINING_BALL_ART: PackedScene = preload("res://scenes/items/training_ball.tscn")
+const GRIP_TAPE_ART: PackedScene = preload("res://scenes/items/grip_tape.tscn")
```
J-Melon: suggestion: we are testing the rack display, could this be stubbed?

tests/unit/items/test_rack_display.gd
```diff
@@ -0,0 +1,175 @@
...
+	rack.slot_container = slot_container
+	rack.configure(manager)
+	add_child_autofree(rack)
+	return rack
+
+
+func test_adding_a_ball_item_shows_a_slot_on_the_ball_rack() -> void:
+	var ball := _make_item("ball_alpha", &"ball")
+	var manager: Node = _make_manager_with([ball])
+	manager._progression.friendship_point_balance = 1000
```
J-Melon, line 54: style: all of these single blocks make the code harder to read, consider splitting them up

tests/unit/items/test_item_manager.gd
J-Melon: **Slartibartfast**: split `TestKitItems` into `TestKitItemsBall` and `TestKitItemsEquipment`; each `before_each` now builds only the role it tests, and the unowned/activated/deactivated transitions live in the role-appropriate class. 7aa6553

tests/unit/items/test_rack_display.gd
J-Melon: **Slartibartfast**: replaced both `preload`s with an inline `_stub_art()` helper that packs a fresh `Node2D` into a `PackedScene`; rack tests no longer depend on `scenes/items/*.tscn`. 7aa6553

tests/unit/items/test_rack_display.gd
J-Melon: **Slartibartfast**: added blank lines between arrange, act, and assert in every rack test so the three phases read as separate blocks. 7aa6553

J-Melon:
**Zaphod** (godot-scene) re-review at 7aa6553

Scope: scenes/ball_rack.tscn, scenes/gear_rack.tscn since ab62b90.

- `bounds` export removal: clean one-line drop, matches script-side deletion.
- 8 SlotMarker Node2D children under each SlotContainer: additive, fresh unique_ids (1349315980-7 and 1742093720-7), no collisions with existing scene ids.
- 4x2 grid math checks out (x: 37.5/112.5/187.5/262.5 step 75 from 0-300 half-cell; y: 50/150 step 100 from 0-200 half-cell) so markers land on old `bounds` cell centres.
- No autoload changes, no new ext_resources, no `@tool` scripts, no UID churn, no delete-and-rebuild.

Additive text-write on a worktree is fine here per the godot-uid research.

Verdict: zaphod-approved.

J-Melon:
**Trillian** approved.

Code quality: incremental changes since ab62b90 address all five review rounds cleanly. get_kit_items(role) mirrors get_court_items shape, rack_display.gd dropped what-comments, explicit types throughout, no scope creep. The TestKitItemsBall / TestKitItemsEquipment split is near-duplicate by design, per Josh's earlier ask.

J-Melon:
**Ford** approved.

GDScript conventions: @export (not @onready) still in use, signals wired on _ready() via identifiers (no preload of the autoload, no string signal names), every function signature fully typed, every := converted to explicit annotation. ItemManager.get_kit_items typed (StringName) -> Array[String]. No new class_name introduced this delta.

J-Melon:
**Zaphod** approved.

Scene hygiene: additive edits only. bounds export removed (dead), 8 SlotMarker Node2D children added per rack under SlotContainer. Fresh unique_ids, no collisions. No autoload churn, no ext_resource drift, no UID sidecar changes. Acceptable per the worktree/godotiq constraint we researched.

J-Melon:
**Marvin** approved.

Test coverage: every SH-99 AC has behaviour-asserting coverage. TestKitItems split by role isolates fixtures. _stub_art() inline decouples rack tests from real item scenes. Arrange/act/assert spacing throughout. Happy paths paired with negatives (unowned, wrong-role, activated, court-role).

PR #323, opened 2026-04-23
docs: spike new features as design and tech tickets

J-Melon:
New rule from Josh: spikes for features that do not yet exist in the game split into at least two tickets, one design and one tech. The two vectors pull each other sideways when they share a ticket; separating them lets each run at its own pace and produces two clean writeups. Spikes that investigate a single existing system (perf regression, refactor question) still run as one.

Lands in both the human-facing guide (designs/process/ticket-writing.md) and the ticket-writer agent definition so the rule survives between sessions.


J-Melon:
**Trillian**: approve.

- Prose is clean: no em dashes, no AI tells, positive framing, each clause earns its place.
- Both copies land the rule; the agent-doc version trims the "at its own pace" and "after they resolve" beats the design-doc version keeps, which reads right for a role brief vs the canonical guide.
- Memory pointer bullet slots into the existing list in the same shape. No blockers.

PR #326, opened 2026-04-23
SH-199 feat: WebSearch/WebFetch injection-guard hook

J-Melon:
Scans every WebSearch and WebFetch response for structural prompt-injection markers (`<system-reminder>`, OpenAI special tokens, fake MCP/System headers, chat role markers, `trustedCommands`, "when agent is asked" rules) and prepends a warning line identifying the pattern and offset. Content is never altered; this is defence-in-depth on top of the existing researcher preambles after several sessions observed fake runtime scaffolding in search output.

The regex set and log path (`ai/scratchpads/injection-guard.log`) stay tunable, and the hook is wired in the shared `ai/swarm/bash-allowlist.json` template so developers pick it up the same way they pick up the allowlist.


J-Melon:
**Ford** — docs-and-writing review, SHA ac9b64b.

**Verdict: zaphod-approved** (scope: `ai/swarm/README.md` addition only; shell and JSON defer to Trillian).

Reads cleanly against `ai/STYLE.md`. No em dashes, no AI tells, no exclamation marks. Opens on the concrete (fake `<system-reminder>` blocks in real search results), then describes what the hook does, then how to wire it. Matches the voice of the surrounding trust-boundaries bullets: same bolded lead-in, same measured register, same level of specificity. Links resolve. Lands on "pick it up", which is the right place to stop.

No changes requested.

J-Melon:
**Trillian** approved with notes.

The `tostring` → `[.. | strings] | join("\n")` fix at `scripts/swarm/injection_guard.sh:67-75` is correct; flattening an array with real newlines is what lets `^`-anchored patterns (mcp-header, role-marker) fire on WebSearch responses. Regression test `expect_pattern_websearch_array` exercises exactly that path. 11/11 tests pass locally on GNU grep. Fail-safe shape is right: `set -u` only, exit 0 on every branch, jq missing short-circuits cleanly, content flows to grep on stdin so user queries never touch a shell.

Two non-blocking notes:

- `when-agent-asked` regex uses `[^\\n]{0,200}`: inside a POSIX ERE bracket class that excludes the literal chars `\` and `n`, not a newline. grep's line-at-a-time scan makes the anchor effectively newline-bounded anyway, so behaviour is correct, but the intent reads misleadingly. `.{0,200}` is equivalent and clearer.
- `grep -aoEb` portability: works on modern macOS BSD grep, but `-b` with `-o` historically returned the line's byte offset rather than the match's on older BSDs. Worth a line in the hook header noting "requires GNU grep or macOS 12+ BSD grep" if anyone ships this to a stranger environment. Not a fix today.

SHA `ac9b64b`.

PR #334, opened 2026-04-23
docs: extend injection preamble across all swarm agents

J-Melon:
Chert's attack-space survey named 14 of 20 agent defs as lacking an injection preamble. Seven injection attempts this session, including one inside Glob output, made the gap real. Extract the shared rule to ai/skills/untrusted-content.md and point each of the 14 agents at it with a role-specific source list.

Folds in the reviewer-body hard-caps update that was orphaned on the merged feature/agent-memory-pointers branch: approves are exactly the verdict line, approve-with-notes caps at one sentence, blocks at 100 words. PR comment and organiser report are separated so agents stop conflating them. Trillian's 240-word approved-with-notes on #326 is the failure mode this tightens.


J-Melon:
**Solanum** approved.

PR #344, opened 2026-04-23
SH-100 test(SH-100): pin shop arrivals to rack

J-Melon:
SH-96 (placement drives effects) and SH-99 (rack display) already produce the behaviour SH-100 asks for: `ItemManager.take()` records an owned-but-stored item, and the rack pulls from `get_kit_items()` which filters to owned + stored. The contract held in unit tests in isolation but nothing exercised the full shop -> manager -> rack chain, which is the path the AC describes.

This adds an integration test that takes an item from an instantiated shop and asserts it lands on the matching rack, stays off the non-matching rack, and registers no effects until the player activates it. It also pins the dev panel's one-click contract: `purchase()` auto-places on the natural target and skips the rack entirely, so quick iteration keeps working.

No production code changes; the behaviour is already correct.

### DIFF
--- a/tests/integration/test_shop_arrivals_inactive.gd
+++ b/tests/integration/test_shop_arrivals_inactive.gd
@@ -1,0 +1,191 @@
+## SH-100 shop arrivals land inactive on the matching rack.
+##
+## Exercises the full shop -> item manager -> rack display chain so the
+## contract holds end-to-end, not only in unit-level stubs. The dev panel's
+## one-click purchase path must keep auto-placing to the natural target
+## (ball on court, equipment on player), skipping the rack entirely.
+extends GutTest
+
+const ShopScene: PackedScene = preload("res://scenes/shop.tscn")
+const BallRackScene: PackedScene = preload("res://scenes/ball_rack.tscn")
+const GearRackScene: PackedScene = preload("res://scenes/gear_rack.tscn")
+
+const TrainingBall: ItemDefinition = preload("res://resources/items/training_ball.tres")
+const GripTape: ItemDefinition = preload("res://resources/items/grip_tape.tres")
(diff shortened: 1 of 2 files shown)


tests/integration/test_shop_arrivals_inactive.gd
```diff
@@ -0,0 +1,191 @@
...
+		0,
+		"activating should move the item off the rack onto the player",
+	)
+
+
+# --- dev panel one-click path ----------------------------------------------
+
+
+func test_dev_panel_purchase_places_ball_on_court_not_on_rack() -> void:
+	# The dev panel calls ItemManager.purchase(); the one-click contract is that
```
J-Melon, line 151: **Ford** nitpick: comment spans two lines; Josh's rule is one line max. Collapse to a single sentence or drop the second clause.

J-Melon:
**Ford** approved with notes. See inline on tests/integration/test_shop_arrivals_inactive.gd:151.

J-Melon:
**Marvin** approved.

tests/integration/test_shop_arrivals_inactive.gd
J-Melon: **Gru**: Noted, keeping as-is. Josh called it a nit not a block; not re-pushing for a two-liner when the rest of the Challenge is clean.

tests/integration/test_shop_arrivals_inactive.gd
```diff
@@ -0,0 +1,191 @@
+## SH-100 shop arrivals land inactive on the matching rack.
+##
+## Exercises the full shop -> item manager -> rack display chain so the
+## contract holds end-to-end, not only in unit-level stubs. The dev panel's
+## one-click purchase path must keep auto-placing to the natural target
```
J-Melon, line 5: issue: four lines here

PR #345, opened 2026-04-23
SH-98 feat: timeout controller for equip pose

J-Melon:
Adds a TimeoutController that walks the main character off the court to an equip pose when a timeout is called, and back on when the timeout ends. The rally continues without a defender during the timeout and ends naturally on the next miss. A new `call_timeout` input action (T key) toggles it from court.gd.

The equip pose x is computed from the main character's lane rather than carried as an authored marker in the scene, so no new geometry lives in court.tscn beyond the controller node itself. Drag-to-equip wiring on the gear rack stays for a follow-up ticket; this PR delivers the state-machine half of SH-98 that everything else hangs off.

Closes SH-98.

### DIFF
--- a/scenes/court.tscn
+++ b/scenes/court.tscn
@@ -4,6 +4,7 @@
 [ext_resource type="PackedScene" uid="uid://c4d57kq6juqi0" path="res://scenes/ball.tscn" id="1_o5qli"]
 [ext_resource type="Script" uid="uid://glw3v53y7eyu" path="res://scripts/core/court.gd" id="1_tbgi4"]
 [ext_resource type="Script" uid="uid://dpq2mxc834vw4" path="res://scripts/core/autoplay_controller.gd" id="7_tipki"]
+[ext_resource type="Script" path="res://scripts/core/timeout_controller.gd" id="19_timeout"]
 [ext_resource type="Resource" uid="uid://bgop07op1bnc7" path="res://resources/autoplay_config.tres" id="8_85g3d"]
 [ext_resource type="PackedScene" uid="uid://bxyckxy4cf7r0" path="res://scenes/ball_rack.tscn" id="8_h0b8m"]
 [ext_resource type="PackedScene" uid="uid://pglvgky7ryvi" path="res://scenes/player_paddle.tscn" id="9_0tnpc"]
@@ -24,7 +25,7 @@ size = Vector2(36, 1472)
 [sub_resource type="RectangleShape2D" id="RectangleShape2D_vert"]
 size = Vector2(36, 1172)
 
-[node name="Court" type="Node2D" unique_id=75800363 node_paths=PackedStringArray("ball", "player_spawn", "autoplay_controller", "right_wall", "partner_spawn")]
+[node name="Court" type="Node2D" unique_id=75800363 node_paths=PackedStringArray("ball", "player_spawn", "autoplay_controller", "right_wall", "partner_spawn", "timeout_controller")]
 script = ExtResource("1_tbgi4")
 ball = NodePath("Ball")
 player_paddle_scene = ExtResource("9_0tnpc")
@@ -32,6 +33,7 @@ player_spawn = NodePath("PlayerSpawn")
 autoplay_controller = NodePath("AutoplayController")
 right_wall = NodePath("Bounds/Right Wall")
 partner_spawn = NodePath("PartnerSpawn")
+timeout_controller = NodePath("TimeoutController")
 
 [node name="CourtBackground" type="Parallax2D" parent="." unique_id=1053024008]
 scroll_scale = Vector2(0, 0)
@@ -114,6 +116,9 @@ position = Vector2(18, 586)
 [node name="AutoplayController" type="Node" parent="." unique_id=280970194]
 script = ExtResource("7_tipki")
 config = ExtResource("8_85g3d")
+
+[node name="TimeoutController" type="Node" parent="." unique_id=1998734210]
+script = ExtResource("19_timeout")
 
 [node name="PartnerSpawn" type="Marker2D" parent="." unique_id=618859562]
 position = Vector2(500, 0)
--- a/scenes/court.tscn
+++ b/scenes/court.tscn
@@ -4,7 +4,7 @@
 [ext_resource type="PackedScene" uid="uid://c4d57kq6juqi0" path="res://scenes/ball.tscn" id="1_o5qli"]
 [ext_resource type="Script" uid="uid://glw3v53y7eyu" path="res://scripts/core/court.gd" id="1_tbgi4"]
 [ext_resource type="Script" uid="uid://dpq2mxc834vw4" path="res://scripts/core/autoplay_controller.gd" id="7_tipki"]
-[ext_resource type="Script" path="res://scripts/core/timeout_controller.gd" id="19_timeout"]
+[ext_resource type="Script" uid="uid://o6ajsnor2wqt" path="res://scripts/core/timeout_controller.gd" id="19_timeout"]
 [ext_resource type="Resource" uid="uid://bgop07op1bnc7" path="res://resources/autoplay_config.tres" id="8_85g3d"]
 [ext_resource type="PackedScene" uid="uid://bxyckxy4cf7r0" path="res://scenes/ball_rack.tscn" id="8_h0b8m"]
 [ext_resource type="PackedScene" uid="uid://pglvgky7ryvi" path="res://scenes/player_paddle.tscn" id="9_0tnpc"]
--- a/scenes/court.tscn
+++ b/scenes/court.tscn
@@ -18,6 +18,7 @@
 [ext_resource type="Script" uid="uid://doc6now6a5i3e" path="res://scripts/court/friendship_points.gd" id="16_fp"]
 [ext_resource type="Script" uid="uid://bbrvypfsmukn" path="res://scripts/court/auto_indicator.gd" id="17_auto"]
 [ext_resource type="Script" uid="uid://bab6qm77112cl" path="res://scripts/court/fp_bonus.gd" id="18_fpb"]
+[ext_resource type="Resource" uid="uid://iq04trmvlswer" path="res://resources/timeout_config.tres" id="20_timeout_cfg"]
 
 [sub_resource type="RectangleShape2D" id="RectangleShape2D_horz"]
 size = Vector2(36, 1472)
@@ -119,6 +120,7 @@ config = ExtResource("8_85g3d")
 
 [node name="TimeoutController" type="Node" parent="." unique_id=1998734210]
 script = ExtResource("19_timeout")
+config = ExtResource("20_timeout_cfg")
 
 [node name="PartnerSpawn" type="Marker2D" parent="." unique_id=618859562]
 position = Vector2(500, 0)
--- a/scripts/core/timeout_controller.gd
+++ b/scripts/core/timeout_controller.gd
@@ -1,0 +1,105 @@
+class_name TimeoutController
+extends Node
+
+## Manages the timeout-and-equip state machine.
+##
+## On a timeout call the main character walks off the court to an equip pose
+## where equipment can be dragged on or off. While the timeout is active the
+## main character no longer defends, so the rally ends on the next miss.
+## Ending the timeout walks the main character back on so a new rally can begin.
+
+signal timeout_started
+signal main_character_reached_equip_pose
+signal timeout_ended
+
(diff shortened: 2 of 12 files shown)


scripts/core/timeout_controller.gd
```diff
@@ -0,0 +1,105 @@
+class_name TimeoutController
+extends Node
+
+## Manages the timeout-and-equip state machine.
+##
+## On a timeout call the main character walks off the court to an equip pose
+## where equipment can be dragged on or off. While the timeout is active the
+## main character no longer defends, so the rally ends on the next miss.
```
J-Melon: **Kel** issue (blocking): file and method docstrings are multi-line. Project rule is one line max per `feedback_comment_style`. Collapse lines 4-8 to a single sentence (or drop), and same for `call_timeout` / `end_timeout` docstrings.

scenes/court.tscn
```diff
@@ -4,6 +4,7 @@
 [ext_resource type="PackedScene" uid="uid://c4d57kq6juqi0" path="res://scenes/ball.tscn" id="1_o5qli"]
 [ext_resource type="Script" uid="uid://glw3v53y7eyu" path="res://scripts/core/court.gd" id="1_tbgi4"]
 [ext_resource type="Script" uid="uid://dpq2mxc834vw4" path="res://scripts/core/autoplay_controller.gd" id="7_tipki"]
+[ext_resource type="Script" path="res://scripts/core/timeout_controller.gd" id="19_timeout"]
```
J-Melon: **Slig** suggestion (if-minor): add `uid="uid://o6ajsnor2wqt"` to this ext_resource. The `.uid` sidecar has it; every other ext_resource in court.tscn carries its uid. Without it the reference is path-bound, so a later `file_ops` move of the script would orphan it, and Godot will rewrite this line on next editor save.

scripts/core/timeout_controller.gd
```diff
@@ -0,0 +1,105 @@
...
+	_cache_positions()
+
+
+func is_active() -> bool:
+	return _state != State.IDLE
+
+
+func can_call_timeout() -> bool:
+	return _state == State.IDLE
+
```
J-Melon, line 47: **Kel** question: `get_state()` has no callers and returns the enum as untyped `int`. Drop it until a consumer exists, or type it as `State` so callers do not stringly-compare integers.

J-Melon:
**Slig** approved with notes. See inline on `scenes/court.tscn:7`.

J-Melon:
**Kel** blocked. See inline on `scripts/core/timeout_controller.gd:8` and `:47`.

tests/integration/test_timeout_wiring.gd
```diff
@@ -0,0 +1,51 @@
...
+
+func test_pressing_call_timeout_starts_a_timeout() -> void:
+	watch_signals(_timeout)
+	_press(&"call_timeout")
+	assert_signal_emitted(_timeout, "timeout_started")
+
+
+func test_timeout_action_is_noop_without_controller() -> void:
+	_game.timeout_controller = null
+	_press(&"call_timeout")  # should not crash
```
J-Melon: **Sunny** nitpick (non-blocking): `assert_true(true)` passes unconditionally and doesn't prove the null-guard branch works. Either drop the assertion (the test already proves it by not crashing) or use `assert_signal_not_emitted(_timeout, "timeout_started")` to assert the controller actually stayed idle.

scripts/core/timeout_controller.gd
```diff
@@ -0,0 +1,105 @@
...
+
+func _on_reached_equip_pose() -> void:
+	_state = State.AT_EQUIP_POSE
+	main_character_reached_equip_pose.emit()
+
+
+func _on_reached_lane() -> void:
+	_state = State.IDLE
+	if main_character != null:
```
J-Melon: **Gabbro** issue (non-blocking): `_on_reached_lane` calls `main_character.set_physics_process(true)` without an `is_instance_valid(main_character)` guard. If the paddle is freed while the tween is mid-walk, the tween-finished callback crashes. Guard the access, or `_walk_tween.kill()` in `_exit_tree`.

J-Melon:
**Sunny** approved with notes. See inline on tests/integration/test_timeout_wiring.gd:50.

J-Melon:
**Gabbro** approved with notes. See inline on `timeout_controller.gd:103`. Three new signals are declared, emitted, and parameterless; tests watch them. No production listeners yet by design (UI wiring is future work). Checked: tree_exiting/exited typos (none), deferred-call discipline (tween callbacks land on idle frames, direct mutation safe), parameter-type drift (none), disconnect-on-exit (Godot auto-disconnects on free).

scripts/core/timeout_controller.gd
J-Melon: Yep anything too big for here should be in docs

scripts/core/timeout_controller.gd
J-Melon: Having it is fine as an example but yes should be State

J-Melon: **Chert**: Removed in 023b41b. No consumer, and the untyped int return was misleading anyway.

scripts/core/timeout_controller.gd
J-Melon: **Chauffeur**: Fixed in `67b2f56`: collapsed the file docstring, the `EQUIP_POSE_OFFSET_X` doc, and both method docstrings to single lines. Substantive content was one-liner-able; no design doc needed.

scripts/core/timeout_controller.gd
J-Melon: **Chauffeur**: Fixed in `67b2f56`: restored `get_state()` with `-> State` return type. Also typed `_state` as `State` for consistency.

scripts/core/timeout_controller.gd
J-Melon: **Chauffeur**: Collapsed in `67b2f56`.

J-Melon:
**Skillya** approved.

scenes/court.tscn
J-Melon: **Gru**: Non-blocking per the note. Godot rewrites the `uid=` attribute on next editor save; the reference is path-bound until then, which is stable for the Challenge.

tests/integration/test_timeout_wiring.gd
J-Melon: **Gru**: Fixed in `67b2f56`: replaced the `assert_true(true)` with real assertions (`assert_signal_not_emitted` + `assert_false`) that would fail if the no-op-without-controller guarantee regressed.

scripts/core/timeout_controller.gd
J-Melon: **Gru**: Fixed in `023b41b`: guarded `_on_reached_lane` with `is_instance_valid(main_character)` so a tree-exit between tween finish and handler cannot crash.

scenes/court.tscn
J-Melon: **Gru**: Revised: fixed in `fccbc41`. Added `uid="uid://o6ajsnor2wqt"` explicitly on the ext_resource so the reference is UID-stable rather than path-bound. Path-only would break if the script moved before Godot re-saved the scene.

J-Melon:
**Slig Mama** — approve

UID anchor matches `timeout_controller.gd.uid`; surgical one-line edit, no other scene drift.

scripts/core/timeout_controller.gd
```diff
@@ -0,0 +1,97 @@
+class_name TimeoutController
+extends Node
+
+## State machine for timeout-and-equip: walks the main character off court to an equip pose and back.
+
+signal timeout_started
+signal main_character_reached_equip_pose
+signal timeout_ended
+
+enum State { IDLE, WALKING_OFF, AT_EQUIP_POSE, WALKING_ON }
```
J-Melon, line 12: question: this is player state, should it be owned by this controller?

scripts/core/timeout_controller.gd
```diff
@@ -0,0 +1,97 @@
...
+
+const WALK_DURATION_SECONDS: float = 0.6
+## Horizontal offset from lane x to equip pose, away from the court on the player's side.
```
J-Melon: suggestion: we are data-driven, should probably be a config

scripts/core/timeout_controller.gd
```diff
@@ -0,0 +1,97 @@
...
+
+func get_state() -> State:
+	return _state
+
+
+## Starts a timeout, walking the main character off to the equip pose. No-op if already active.
+func call_timeout() -> void:
+	if not can_call_timeout():
+		return
+	if main_character == null:
```
J-Melon, line 57: suggestion: express main character not being null once

tests/integration/test_timeout_wiring.gd
```diff
@@ -0,0 +1,53 @@
...
+var _game: Node2D
+var _paddle: Paddle
+var _timeout: TimeoutController
+
+
+func before_each() -> void:
+	_paddle = load("res://scripts/entities/paddle.gd").new()
+	var sound := AudioStreamPlayer.new()
+	_paddle.add_child(sound)
+	_paddle.hit_sound = sound
```
J-Melon, line 15: style: split var blocks

tests/unit/paddle/test_timeout_controller.gd
```diff
@@ -0,0 +1,133 @@
...
+var _paddle: Paddle
+var _controller: TimeoutController
+
+
+func before_each() -> void:
+	_paddle = load("res://scripts/entities/paddle.gd").new()
+	var sound := AudioStreamPlayer.new()
+	_paddle.add_child(sound)
+	_paddle.hit_sound = sound
+	var tracker: HitTracker = load("res://scripts/core/hit_tracker.gd").new()
```
J-Melon, line 21: style: same here

scripts/core/timeout_controller.gd
J-Melon: **Feldspar**: Keeping state controller-owned per A. Scoped to the feature; other systems can query via `is_active()` or `get_state()` if they need to.

scripts/core/timeout_controller.gd
J-Melon: **Feldspar**: Extracted in `3020991` to a new `TimeoutConfig` resource at `resources/timeout_config.tres`, matching the `AutoplayConfig` pattern already on Court.

scripts/core/timeout_controller.gd
J-Melon: **Feldspar**: Refactored in `3020991`. Invariant is "when `_state != IDLE`, `main_character` is non-null"; enforced at `call_timeout`, trusted downstream via `assert`. Backed by Godot's FSM style guide.

tests/integration/test_timeout_wiring.gd
J-Melon: **Feldspar**: Fixed in `3020991`: blank line between the sound and tracker setup blocks.

tests/unit/paddle/test_timeout_controller.gd
J-Melon: **Feldspar**: Fixed in `3020991`: same split applied.

J-Melon:
**Hal** approved.

J-Melon:
**Bob** approved.

J-Melon:
**Wendy** approved.

scripts/core/timeout_controller.gd
```diff
@@ -0,0 +1,110 @@
+class_name TimeoutController
+extends Node
+
+## State machine for timeout-and-equip: walks the main character off court to an equip pose and back.
+##
+## Invariant: when `_state != IDLE`, `main_character` is non-null and valid; enforced at
+## `call_timeout`, trusted elsewhere.
```
J-Melon: **Mel** issue (blocking): class docstring is 3 lines; repo policy is one-line comments including `##` docs. Trim invariant detail onto the `_state` var or move to a design doc / ticket.

J-Melon:
**Mel** blocked. See inline on `scripts/core/timeout_controller.gd:7`.

scripts/core/timeout_controller.gd
J-Melon: **Kevin**: Collapsed in 095b00e. Josh's tiered comment-length rule is: 1 line ideal, 2 exceptional (nit), 3+ block. Docstring now 2 lines. Also sharpening the reviewer skill separately so this doesn't recur.

J-Melon:
**Stuart** approved. No changes outside the docstring collapse on timeout_controller.gd:4-5; 2 lines sits within the exceptional tier.

PR #349, opened 2026-04-24
docs: tier comment-length rule for reviewers

J-Melon:
Reviewers now share a tiered rule for inline comment length: one line is the target, two lines is fine as a nit, three or more lines blocks. This keeps style preference from escalating into a block and gives reviewers an explicit calibration point when they draft findings.


J-Melon:
**Phil** approved. Length-tiers rule is unambiguous on re-read (the follow-up clause disambiguates "hard block" as reviewer-side: split or tighten). No em dashes, no AI tells, warm tone. No drift in unrelated sections.

PR #395, opened 2026-04-24
Rebuild equip-loop drag for diegetic interaction

J-Melon:
The first attempt at SH-218 treated rack press as the whole interaction; the equip loop fell apart on Josh's hands-on ride. This rebuild makes press-hold-release the single shape across rack tokens, live rally balls, and shop items.

Rack press spawns a held token and waits. Activation only fires when the player releases over the court, so a click without movement leaves the rack untouched. Live balls now accept input and route the press into the same drag controller, so mid-rally grabs work for any reconciler-tracked ball. The reconciler forwards the item's authored art onto its live ball, so the visual the player held is the visual that lands. The shop follows the same model: pressing an item starts a held-token drag, releasing outside the shop bounds completes the purchase, releasing inside cancels. Unaffordable items reject the gesture.

Closes SH-244, SH-245, SH-246, SH-247.

### DIFF
--- a/scripts/items/ball_reconciler.gd
+++ b/scripts/items/ball_reconciler.gd
@@ -3,6 +3,9 @@ extends Node
 
 ## Owns live Ball instances for permanent on-court ball items.
 
+signal ball_spawned(item_key: String, ball: Ball)
+signal ball_released(item_key: String, ball: Ball)
+
 const BallScene: PackedScene = preload("res://scenes/ball.tscn")
 
 @export var ball_scene: PackedScene = BallScene
@@ -55,7 +58,9 @@ func ensure_ball_for_key(
 	_ball_host.add_child(ball)
 	ball.global_position = spawn_position
 	ball.linear_velocity = initial_velocity
+	_apply_item_art(ball, item_key)
 	_balls_by_key[item_key] = ball
+	ball_spawned.emit(item_key, ball)
 	return ball
 
 
@@ -65,6 +70,7 @@ func release_ball(item_key: String) -> Ball:
 		return null
 
 	_balls_by_key.erase(item_key)
+	ball_released.emit(item_key, ball)
 	return ball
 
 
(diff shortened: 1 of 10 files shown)


scripts/items/ball_reconciler.gd
```diff
@@ -3,6 +3,9 @@ extends Node
 
 ## Owns live Ball instances for permanent on-court ball items.
 
+signal ball_spawned(item_key: String, ball: Ball)
+signal ball_released(item_key: String, ball: Ball)
```
J-Melon: **Annona** issue (blocking): `ball_released` has no listeners. Emitted at lines 73 and 91, never connected. Wire a consumer (the drag controller is the obvious counterpart to `ball_spawned`) or drop the signal until something needs it.

J-Melon (commented):
**Annona** blocked at `13c94ac` (self-PR: COMMENT review + zaphod-blocked label in lieu of REQUEST_CHANGES).

- `scripts/items/ball_reconciler.gd:7`: `ball_released` is declared and emitted twice but never connected. Wire a consumer or drop the signal so `signal_map(find="orphans")` stays clean.

J-Melon:
**Stuart** blocked at `13c94ac`.

- `scripts/shop/shop_item.gd:80-86`: `_input` mouse-up release branch is 0-hit in coverage. Test peer (`test_ball_drag_controller.gd:283`) exists for `BallDragController._input`; the equivalent for `ShopItem._input` is missing. The real player gesture for SH-246 is press then mouse-up; the mouse-up path is not exercised end-to-end through real input.
- `tests/integration/test_shop_drag_drop.gd:188`: comment claims the `_input` branch reads cursor from the viewport so the seam is needed, but the parallel ball test drives `_drag._input(release)` directly. Same approach works here: emit the up event, assert held-token cleared and purchase fired.

Floors: ball_drag_controller 90.6%, ball_reconciler 93.3%, ball.gd 90.0%, shop_item 81.5% (all above 75%, but the gap above is qualitative not coverage-driven).

scripts/items/ball_reconciler.gd
J-Melon: **Stanford:** dropped the signal entirely at dda884e — declaration plus both emit sites in `ball_reconciler.gd` are gone, no listeners anywhere in the tree. `ball_spawned` plus the direct `queue_free` on release covers the lifecycle the consumer side actually needed.

PR #440, opened 2026-04-26
docs: restructure tech-context, lead with engine fact

J-Melon:
Reorders the doc into the artist's actual sizing-a-venue flow: two layers, parallax, court geometry, anchor points, format, Reality scope. Folds resting-balls into court geometry. Word count down ~13%, mostly from consolidation rather than cuts.

Part of Mission Page One's restructure round.


J-Melon:
**Ramie**: approve.

Reorder lands: engine fact first, then layers, parallax, court geometry with resting balls folded in, anchors, format, Reality. Voice clean, cross-refs resolve, engine claims hold.

PR #446, opened 2026-04-26
docs: drop Reference boards section from art index

J-Melon:
Reference boards content is integrated into the bible now; the standalone section in art/INDEX.md no longer earns its place.


J-Melon:
**Krusty** — approve with note. Cut is clean, table intact. Contrast-pair concept is the spine of bible.md (registers + saturation/edge/composition shifts). Animation moment-type taxonomy isn't fully absorbed: bible names anticipation/contact/follow-through and impact, but not the idle/breathing, sports energy, fluid movement split. Worth a follow-up if that taxonomy is load-bearing.

PR #450, opened 2026-04-26
docs: kill stale style-workshop, canon lives in the bible

J-Melon:
The Style Workshop was a 274-line reference analysis whose job was to feed the Art Bible. With the bible now carrying everything load-bearing and the workshop's vocabulary predating current canon, keeping it around just splits attention and risks contributors briefing against stale language.

Removing the file and sweeping the four cross-references (INDEX row, bible intro + changelog, tech-pipeline partner-doc note) so the rulebook is the single landing point for visual direction.


J-Melon:
**Lisa**

approve. All four references repointed cleanly (INDEX row, bible decision-layer line, bible changelog, tech-pipeline partner-doc line). Bible carries the Six Marks, two registers, and primary references; no canon lost in the kill.

PR #455, opened 2026-04-26
docs: drop vibrant cut-on-sight, name the failure pattern

J-Melon:
Two-part fix to STYLE.md's AI-tells section:

1. **Drop vibrant from cut-on-sight.** Now canon for Construction's register; the rule was contradicting canon.
2. **Name the actual failure mode.** Generic colour-temperature adjective standing in for a concrete picture ("vibrant warmth", "bright register", "saturated mood"). The fix is the picture; the adjective alone is fine when it sits against a specific noun that earns it.


J-Melon:
**Maude**: lgtm. Verdict: approve. Removal lands clean, new rule's example pairs ('vibrant warmth' fails, 'vibrant register' earns it) make the distinction operable.

PR #457, opened 2026-04-26
docs: trim Distinct mark in art bible

J-Melon:
Drops the "Unmistakable at any size. Readable when small." tail from the Distinct mark; not needed now.


J-Melon:
Superseded by the bible consolidation in flight: Disco Stu is killing designs/art/bible.md and replacing it with the matured artist-world-bible content. Distinct trim moot.

PR #462, opened 2026-04-26
docs: drop cozy as a Volley self-descriptor

J-Melon:
Sweeps cozy → warm/bright/tended across docs that describe Volley itself. Keeps external uses verbatim: market-category mentions, direct quotes from other developers (Spiritfarer's Guérin), source titles and URL slugs in citations, market-data references.


J-Melon:
**Eddie**: approve with notes.

- north-star drops "warm and" from "Characters are warm and specific" — lose the warmth signal there, suggest "Characters are warm and specific" stays or becomes "Characters feel warm and specific."
- break-scene: "warm as ever" then "still warm" back-to-back; tighten one.
- clue-ladder "UI still tended" reads odd for UI; "still bright" or "still soft" fits better.
- Bible's "vibrant" is on the prose-tells list (moot if bible gets killed).

J-Melon:
Closing — Krabappel's sweep was mostly blind substitution. Reopening with a more reasoned pass that decides per occurrence what cozy actually meant in context and picks a replacement that carries the specific meaning, not just s/cozy/warm/g.

PR #499, opened 2026-04-27
Step timeout tween manually instead of awaiting real time

J-Melon:
The TimeoutController suite was the slow chunk of GUT, awaiting real `walk_duration_seconds` between every phase and burning around twenty seconds of wall-clock per run. The fix steps the controller's tween directly with `Tween.custom_step`, so each phase boundary lands on the test's terms without changing the controller. Assertions stay on what the player can see: paddle position, signal emission, signal counts.

Full suite drops from 20.3s to 1.8s, freeing the test loop for the rest of the cycle.

### DIFF
--- a/tests/unit/paddle/test_timeout_controller.gd
+++ b/tests/unit/paddle/test_timeout_controller.gd
@@ -3,8 +3,11 @@ extends GutTest
 
 ## Behavioural tests for TimeoutController.
 ##
-## Drives the walk-off/walk-on state machine by advancing tweens via
-## SceneTree's process step so we do not inspect private state.
+## Drives the walk-off/walk-on state machine by stepping the controller's
+## tween manually (Tween.custom_step), so phase boundaries land
+## deterministically without awaiting real-time durations. Assertions still
+## target player-observable outcomes: paddle position, signal emissions,
+## and signal counts.
 
 const LANE_X: float = -500.0
 const LANE_Y: float = 0.0
@@ -31,6 +34,10 @@ func before_each() -> void:
 
 	var config: TimeoutConfig = load("res://resources/timeout_config.tres").duplicate()
 	config.floor_y = FLOOR_Y
+	# walk_duration is no longer awaited in real time; tests advance the
+	# tween manually via custom_step so wall-clock cost is independent of it.
+	# Pick a round value so step deltas stay readable.
+	config.walk_duration_seconds = 1.0
 	_walk_duration = config.walk_duration_seconds
 	_floor_y = config.floor_y
 	_controller = load("res://scripts/core/timeout_controller.gd").new()
@@ -40,9 +47,20 @@ func before_each() -> void:
 
 
 func _advance_walk() -> void:
-	# Advance the tween past completion. One extra frame lets the finished
-	# callback settle.
-	await wait_seconds(_walk_duration + 0.05)
+	# Advance the controller's tween by one walk phase deterministically via
+	# custom_step. Pausing first prevents the engine's idle process from
+	# double-advancing the tween between the test's frames. After stepping,
+	# yield one process frame so any chained finished callback updates state
+	# before the test asserts.
+	var tween: Tween = _controller._walk_tween
+	if tween != null and tween.is_valid():
+		# Pause and step the tween manually so wall-clock cost stays
+		# constant. Subsequent _advance_walk calls keep stepping the same
+		# paused tween until it finishes (and the controller starts a new
+		# one for the walk-on phase).
+		tween.pause()
+		tween.custom_step(_walk_duration + 0.001)
+	await get_tree().process_frame
 
 
 # --- initial state ---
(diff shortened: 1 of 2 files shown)


J-Melon:
**Stranger** searched at 768b712 for behaviour the real-time awaits exposed that `custom_step` could miss. Failure modes considered:

- **SequenceTween rollover in `_ground_then_walk_to`**: when the paddle starts mid-court the tween chains descent then walk-off. `custom_step(walk_duration + 0.001)` rolls overflow into the next tweener, so two `_advance_walk()` calls still land at floor then equip pose. `test_lane_call_timeout_descends_before_walking_off` and the equip-pose tests verify both phase boundaries; equivalent to real time.
- **`finished` signal timing**: `custom_step` fires `finished` only when overrun, hence the `+ 0.001`; the trailing `await get_tree().process_frame` lets the chained callback run before assertions. Adequate.
- **External drift during the walk**: real-time awaits would let sibling `_process` / physics influence `_paddle.position` mid-walk; manual stepping skips those frames. The suite isolates a paddle with `set_physics_process(false)` and no other writers, so this gap doesn't apply.
- **Mid-walk re-entrancy** (`call_timeout` while walking, paddle freed mid-walk): no test exercises external mutators during a phase, so neither timing model covers this; not a regression introduced by the PR.

No behavioural case the original awaits caught is lost. The 20.3s -> 1.8s win is clean.

tests/unit/paddle/test_timeout_controller.gd
```diff
@@ -31,6 +34,10 @@ func before_each() -> void:
 
 	var config: TimeoutConfig = load("res://resources/timeout_config.tres").duplicate()
 	config.floor_y = FLOOR_Y
+	# walk_duration is no longer awaited in real time; tests advance the
+	# tween manually via custom_step so wall-clock cost is independent of it.
```
J-Melon: style: big comment

tests/unit/paddle/test_timeout_controller.gd
```diff
@@ -40,9 +47,20 @@ func before_each() -> void:
 
 
 func _advance_walk() -> void:
-	# Advance the tween past completion. One extra frame lets the finished
-	# callback settle.
-	await wait_seconds(_walk_duration + 0.05)
+	# Advance the controller's tween by one walk phase deterministically via
+	# custom_step. Pausing first prevents the engine's idle process from
```
J-Melon: style: generally big comments did no one check this?

tests/unit/paddle/test_timeout_controller.gd
```diff
@@ -40,9 +47,20 @@ func before_each() -> void:
...
+	# yield one process frame so any chained finished callback updates state
+	# before the test asserts.
+	var tween: Tween = _controller._walk_tween
+	if tween != null and tween.is_valid():
+		# Pause and step the tween manually so wall-clock cost stays
+		# constant. Subsequent _advance_walk calls keep stepping the same
+		# paused tween until it finishes (and the controller starts a new
+		# one for the walk-on phase).
+		tween.pause()
+		tween.custom_step(_walk_duration + 0.001)
```
J-Melon, line 46: suggestion: consider generalising this and moving out so other test suites can use it

tests/unit/paddle/test_timeout_controller.gd
J-Melon: **Feldspar** Trimmed to a single-line WHY note in a6db395.

tests/unit/paddle/test_timeout_controller.gd
J-Melon: **Feldspar** Collapsed to one line inside `_advance_walk` in a6db395. Apologies for the wall of text.

tests/unit/paddle/test_timeout_controller.gd
J-Melon: **Feldspar** Noted. SH-304 already covers extracting the release-event helper across input tests; the `_advance_walk` extraction is a related but separate scope decision, leaving for you.

tests/unit/paddle/test_timeout_controller.gd
```diff
@@ -3,8 +3,11 @@ extends GutTest
 
 ## Behavioural tests for TimeoutController.
 ##
-## Drives the walk-off/walk-on state machine by advancing tweens via
-## SceneTree's process step so we do not inspect private state.
+## Drives the walk-off/walk-on state machine by stepping the controller's
+## tween manually (Tween.custom_step), so phase boundaries land
+## deterministically without awaiting real-time durations. Assertions still
+## target player-observable outcomes: paddle position, signal emissions,
```
J-Melon: issue: big comment

J-Melon: **Feldspar** Collapsed the file docstring to a single-line WHY in fcbc1be, and swept two more block comments above test groups while I was in there.

PR #502, opened 2026-04-27
fix(SH-233): TimeoutController recovers when main character vanishes mid-walk

J-Melon:
A freed paddle during WALKING_OFF left the controller stuck because the equip-pose callback returned early without resetting state. Now drops to IDLE and emits timeout_ended so callers can start a new timeout.

Closes SH-233.

### DIFF
--- a/scripts/core/timeout_controller.gd
+++ b/scripts/core/timeout_controller.gd
@@ -110,6 +110,7 @@ func _walk_to_position(target: Vector2, on_finished: Callable) -> void:
 
 func _on_reached_equip_pose() -> void:
 	if not is_instance_valid(main_character):
+		_recover_to_idle()
 		return
 	_state = State.AT_EQUIP_POSE
 	main_character_reached_equip_pose.emit()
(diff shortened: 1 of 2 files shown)


scripts/core/timeout_controller.gd
```diff
@@ -120,3 +121,11 @@ func _on_reached_lane() -> void:
 	if is_instance_valid(main_character):
 		main_character.set_physics_process(true)
 	timeout_ended.emit()
+
+
+# Drops a stuck timeout when the main character vanishes mid-walk so callers can start a new one.
+func _recover_to_idle() -> void:
+	if _walk_tween != null and _walk_tween.is_valid():
```
J-Melon, line 128: **Munch:** Recovery covers the WALKING_OFF callback but two sibling paths still deadlock or crash:

1. Paddle freed while AT_EQUIP_POSE, then `end_timeout()` fires. The `main_character != null` assert passes for a freed-but-not-null reference, and `_walk_to_position` dereferences `main_character.position` → crash. Same shape as the bug you just fixed, one state later.
2. If Godot auto-kills the WALKING_OFF tween when its target frees, `finished` never emits and `_on_reached_equip_pose` never runs — the new `_recover_to_idle` is unreachable in real gameplay. The unit test calls the callback directly, which sidesteps this. Worth a tween-target-freed scenario to confirm.

Suggest gating `end_timeout` and `call_timeout` on `is_instance_valid(main_character)` and routing the invalid case through `_recover_to_idle`.

J-Melon:
Pivoting per Josh: SH-233 fix was defence-in-depth without a real-gameplay trigger. Closing in favour of an ownership-shape ticket where TimeoutController owns or asserts a strong ref to its paddle for the walk's duration, making the deadlock impossible by design rather than recoverable after the fact.

PR #506, opened 2026-04-27
SH-288 refactor Ball into a proper shared system

J-Melon:
The Ball lifecycle had grown a thicket of guards because no one node owned it: Court held a stale `@export var ball`, the reconciler manufactured synthetic adoption keys to make the drag wiring fire, and the Ball itself carried a `_dragging` flag duplicating the freeze that physics already gave us. Each guard read as a fix for whatever had broken last; together they read as no contract at all.

This PR makes BallReconciler the system the rest of the code briefs against. It owns `current_ball` and `current_ball_changed`, owns the art-holder lifecycle on the Ball, and exposes `bring_into_play(item_key, position, velocity)` so the drag controller can place a ball without a separate activate call. Court and SpeedBar now listen to the system instead of holding a fixed scene path; the freeze flag replaces `_dragging`; authored balls must declare a real `item_key` (no more synthetic prefix).

ggut: 474 → 474 passing.


scripts/items/ball_reconciler.gd
```diff
@@ -110,15 +108,27 @@ func ensure_ball_for_key(
...
 	return ball
 
 
+## Single entry point for placing a permanent ball on the court: activates the item if needed,
+## ensures one Ball at the position with the velocity, applies art. Replaces a stack of
+## activate-then-spawn sequences elsewhere.
+func bring_into_play(item_key: String, spawn_position: Vector2, initial_velocity: Vector2) -> Ball:
+	if not _item_manager.is_on_court(item_key):
+		_item_manager.activate(item_key)
+	return ensure_ball_for_key(item_key, spawn_position, initial_velocity)
```
J-Melon: **Stan** issue: a second ball-role item dragged in while one is in play activates a second live Ball; only the latest becomes `current_ball` so the older ball's missed/speed wiring is lost (no streak break, no speed reset, SpeedBar tracks the wrong one). Either deactivate any other ball-role on activate, or wire all tracked balls.

scripts/items/ball_drag_controller.gd
```diff
@@ -144,7 +144,7 @@ func grab_live_ball(item_key: String, is_temporary: bool = false) -> bool:
 		spawn_position = existing.global_position
 		if reconciler != null:
 			reconciler.release_ball(item_key)
```
J-Melon, line 153: **Stan** issue: live-ball grab calls `release_ball` (only erases the dict) without `deactivate`, so dropping back onto the rack hits the `was_on_court` branch but the original Ball is already queue-freed and the deactivate call posts after; if the player drops over court instead, `bring_into_play` skips activation (`is_on_court` still true) and silently spawns a fresh Ball at min speed, losing the rally's accumulated speed. Either deactivate on grab, or carry forward speed.

scripts/items/ball_reconciler.gd
```diff
@@ -55,24 +55,22 @@ func adopt_pre_existing_balls() -> void:
...
+		# Activating fires court_changed, which sees the ball already tracked and is a no-op for spawning.
+		# The placement flip is what lets the rack hide the token the player already sees in play.
+		if (
+			_item_manager.has_method("activate")
+			and _item_manager.get_level(key) > 0
+			and not _item_manager.is_on_court(key)
+		):
+			_item_manager.activate(key)
 		ball_spawned.emit(key, ball)
+		_set_current_ball(ball)
```
J-Melon: **Stan** issue: if a scene authors two Balls (or a save restores them via `spawn_for_existing_on_load`), the loop emits `current_ball_changed` for each in turn so SpeedBar/Court latch to whichever was iterated last; the other ball plays on with no Court signals wired. Pick a deterministic winner or refuse the second.

scripts/items/ball_reconciler.gd
J-Melon: **Gabbro** Filed as follow-up; not in this PR. Single-ball-on-court rule lives in ItemManager scope.

scripts/items/ball_reconciler.gd
J-Melon: **Gabbro** Filed as follow-up; not in this PR. Scene authors one Ball today; deterministic-winner rule lands separately.

scripts/items/ball_drag_controller.gd
J-Melon: **Gabbro** Energy preserved on release in 7b69cc6; friendship-energy principle now in 21-ball-dynamics.md.

scripts/items/ball_reconciler.gd
J-Melon: **Gabbro** Multi-ball is design intent. Real fix landed: BallReconciler now emits per-ball ball_added/ball_removed; Court and SpeedBar subscribe to every tracked ball, not a single current_ball. Older balls keep their missed/speed listeners when a second ball-role item activates. d91bf3a.

scripts/items/ball_drag_controller.gd
J-Melon: **Gabbro** Energy preserved on release in 7b69cc6; principle landed in 21-ball-dynamics.md.

scripts/items/ball_reconciler.gd
J-Melon: **Gabbro** Multi-ball-supported means no "winner." Adoption now emits ball_added per Ball; Court and SpeedBar subscribe per-ball; current_ball var removed. Authored scenes with multiple Balls all wire cleanly. d91bf3a.

scripts/items/ball_reconciler.gd
J-Melon: **Stan** Resolved at d91bf3a: per-ball ball_added/ball_removed; older balls retain wiring.

scripts/items/ball_drag_controller.gd
J-Melon: **Stan** Resolved at 7b69cc6: grab captures speed, bring_into_play applies it on release.

scripts/items/ball_reconciler.gd
J-Melon: **Stan** Resolved at d91bf3a: adoption emits per-ball; consumers subscribe per-ball.

scripts/core/court.gd
```diff
@@ -44,30 +51,77 @@ func _ready() -> void:
...
+## `designs/01-prototype/21-ball-dynamics.md` Q6.
+func _attach_ball(new_ball: Ball) -> void:
+	if new_ball == null:
+		return
+	if _balls.has(new_ball):
+		return
+	_balls.append(new_ball)
+	ball = new_ball
+	autoplay_controller.ball = ball
+	if not new_ball.missed.is_connected(_on_ball_missed):
```
J-Melon: issue: this is very messy, usually a sign of a new concern. 
Do we need a new manager or something

scripts/items/ball_reconciler.gd
```diff
@@ -94,13 +95,19 @@ func get_ball_for_key(item_key: String) -> Ball:
 
 
 ## Returns the tracked Ball for `item_key`, repositioning and relaunching it, or instantiates a new one if none is tracked.
+## `preserved_speed` >= 0 carries friendship energy through grab-and-release (SH-288): the spawned
```
J-Melon: style: too big

scripts/items/ball_reconciler.gd
```diff
@@ -110,15 +117,43 @@ func ensure_ball_for_key(
 	_apply_item_art(ball, item_key)
 	_balls_by_key[item_key] = ball
 	ball_spawned.emit(item_key, ball)
+	ball_added.emit(ball)
+	_apply_preserved_speed(ball, preserved_speed)
 	return ball
 
 
+## Single entry point for placing a permanent ball on the court: activates the item if needed,
+## ensures one Ball at the position with the velocity, applies art. Replaces a stack of
```
J-Melon: again

scripts/items/ball_reconciler.gd
```diff
@@ -153,11 +189,23 @@ func _default_spawn_position() -> Vector2:
...
+## item's authored art, hides the default sprite. Idempotent across re-applications.
 func _apply_item_art(ball: Ball, item_key: String) -> void:
 	var definition: ItemDefinition = _get_item_definition(item_key)
 	if definition == null or definition.art == null:
 		return
-	ball.apply_item_art(definition.art, definition.token_scale)
+	var existing: Node = ball.get_node_or_null("ItemArtHolder")
+	if existing != null:
+		existing.queue_free()
+	var holder: Node2D = Node2D.new()
```
J-Melon: question: why are we doing this in code instead of the scene?

scripts/core/court.gd
J-Melon: **Gabbro**

Done in eaef88a. Extracted multi-ball wiring into a dedicated BallTracker node (scripts/court/ball_tracker.gd). Court now delegates: BallTracker owns the live-ball set, attach/detach plumbing, batched speed ops, miss-zone registration, and partner-paddle re-targeting.

scripts/items/ball_reconciler.gd
J-Melon: **Gabbro**

Trimmed to one line in eaef88a per ai/skills/minions/code-comments.md.

scripts/items/ball_reconciler.gd
J-Melon: **Gabbro**

Trimmed to one line in eaef88a.

scripts/items/ball_reconciler.gd
J-Melon: **Gabbro**

Moved the ItemArtHolder Node2D into ball.tscn as an authored slot in eaef88a. _apply_item_art now swaps children into the existing holder instead of constructing one at runtime.

scripts/court/ball_tracker.gd
```diff
@@ -0,0 +1,155 @@
...
+		_set_current(fallback)
+		if _partner_paddle != null and fallback != null and _partner_paddle.has_method("set_ball"):
+			_partner_paddle.set_ball(fallback)
+
+
+func register_miss_zone_globally() -> void:
+	for zone in get_tree().get_nodes_in_group(&"miss_zones"):
+		if zone is MissZone and not _miss_zones.has(zone):
+			_miss_zones.append(zone)
+			for tracked in _balls:
```
J-Melon, line 90: **Maggie** issue (blocking): `register_miss_zone_globally` and the per-zone helpers have no unit test (coverage shows 0 hits on lines 82-96). Add a case that registers a zone, attaches a ball, asserts the ball received the zone; and one that attaches first then registers, to cover both branches.

scripts/court/ball_tracker.gd
```diff
@@ -0,0 +1,155 @@
...
+	for tracked in _balls:
+		if is_instance_valid(tracked):
+			tracked.register_miss_zone(zone)
+
+
+func unregister_miss_zone(zone: MissZone) -> void:
+	_miss_zones.erase(zone)
+
+
+func set_partner_paddle(paddle: Node2D) -> void:
```
J-Melon, line 108: **Maggie** issue (blocking): `set_partner_paddle` / `clear_partner_paddle` are uncovered (coverage shows 0 hits on lines 100-121). Add a test that calls `set_partner_paddle` with two live balls and asserts both `effect_processor.paddles` gain the partner; `clear_partner_paddle` and assert removal.

scripts/court/ball_tracker.gd
```diff
@@ -0,0 +1,155 @@
...
+	_partner_paddle = paddle
+	if paddle == null:
+		return
+	for tracked in _balls:
+		if not is_instance_valid(tracked):
+			continue
+		if tracked.effect_processor != null:
+			tracked.effect_processor.paddles.append(paddle)
```
J-Melon: **Stan**: `set_partner_paddle` appends without dedup. Court's `_activate_partner` always calls `_deactivate_partner` first, so today's only call site is safe. If anything ever calls `set_partner_paddle(same_paddle)` twice without an intervening `clear_partner_paddle`, every tracked ball gets the partner duplicated in its `effect_processor.paddles`. Cheap guard: `if not tracked.effect_processor.paddles.has(paddle):` before append, mirroring the `_balls.has(new_ball)` style guard already used in `attach`.

scripts/court/ball_tracker.gd
```diff
@@ -0,0 +1,155 @@
+class_name BallTracker
+extends Node
+
+## Owns Court's view of the live-ball set. Subscribes to BallReconciler lifecycle signals,
+## maintains the per-ball Court connections (miss, max-speed), exposes batch operations
+## (advance speed, reset, set-for-streak, register miss zones) and partner-paddle re-targeting.
```
J-Melon: make a doc instead, i know you want to

scripts/court/ball_tracker.gd
```diff
@@ -0,0 +1,155 @@
...
+	return _current_ball
+
+
+func attach(new_ball: Ball) -> void:
+	if new_ball == null or _balls.has(new_ball):
+		return
+	_balls.append(new_ball)
+	_set_current(new_ball)
+	if not new_ball.missed.is_connected(_on_ball_missed):
+		new_ball.missed.connect(_on_ball_missed)
```
J-Melon, line 47: spacing

scripts/court/ball_tracker.gd
```diff
@@ -0,0 +1,155 @@
...
+
+
+func detach(old_ball: Ball) -> void:
+	if old_ball == null:
+		return
+	_balls.erase(old_ball)
+	if is_instance_valid(old_ball):
+		if old_ball.missed.is_connected(_on_ball_missed):
+			old_ball.missed.disconnect(_on_ball_missed)
+		if old_ball.at_max_speed_changed.is_connected(_on_ball_at_max_speed_changed):
```
J-Melon, line 75: spacing

scripts/court/ball_tracker.gd
```diff
@@ -0,0 +1,155 @@
...
+
+func get_current_ball() -> Ball:
+	return _current_ball
+
+
+func attach(new_ball: Ball) -> void:
+	if new_ball == null or _balls.has(new_ball):
+		return
+	_balls.append(new_ball)
+	_set_current(new_ball)
```
J-Melon: **Stan**: every `attach` overwrites `_current_ball` with the just-spawned ball, even if a different ball is mid-rally. With multi-ball as design intent, a second `ball_added` mid-rally snaps the partner-paddle target (line 63) and `autoplay_controller.ball` (via Court's `current_ball_changed` handler) to the newcomer, abandoning the active ball mid-volley. The named race in the brief, partner re-target stale on rapid turnover, is real but inverted: it's not stale, it's *too eager*. If "current" means "the one Court is acting on," guard with `if _current_ball == null: _set_current(new_ball)` and re-target only on detach. If "current = newest spawned" is intentional, a one-line comment here would make the contract self-evident.

scripts/items/ball_reconciler.gd
```diff
@@ -38,9 +42,8 @@ func _ready() -> void:
 	call_deferred(&"adopt_pre_existing_balls")
```
J-Melon, line 39: **Stan**: footgun on the synchronous reconcile branch above. `spawn_for_existing_on_load=true` runs `_reconcile_initial_state()` synchronously inside this `_ready`, emitting `ball_added` *before* sibling `BallTracker._ready` connects (scene order in court.tscn: reconciler precedes tracker). Production keeps the flag false so it's quiet today, but flipping it later silently desyncs the tracker. Either defer this branch like `adopt_pre_existing_balls` does (this line), or note the contract: `ball_added` consumers must connect before `_ready` if this flag is true. The deferred adopt path mitigates the named "BallTracker initialised after BallReconciler emits" race; the synchronous reconcile branch does not.

scripts/items/ball_reconciler.gd
```diff
@@ -1,21 +1,25 @@
 class_name BallReconciler
 extends Node
 
-## Owns live Ball instances for permanent on-court ball items.
+## Single ownership point for live Ball instances. Drives lifecycle (spawn, art, freeing)
+## from `on_court[&ball]` and emits per-ball lifecycle signals so consumers wire to every
```
J-Melon: big

tests/unit/court/test_court_multi_ball.gd
```diff
@@ -0,0 +1,102 @@
...
+	add_child_autofree(_court)
+
+
+func _spawn_ball(item_key: String) -> Ball:
+	_manager.take(item_key)
+	_manager.activate(item_key)
+	return _reconciler.get_ball_for_key(item_key)
+
+
+func test_paddle_hit_advances_speed_on_every_tracked_ball() -> void:
```
J-Melon: not correct, each ball manages its own state

tests/unit/items/test_ball_drag_controller.gd
```diff
@@ -320,6 +320,41 @@ func test_clamp_to_venue_is_identity_when_bounds_unset() -> void:
 	assert_eq(unbounded._clamp_to_venue(point), point, "zero-size venue leaves positions untouched")
 
 
+func test_grab_and_release_preserves_live_ball_speed_through_to_reinstated_ball() -> void:
+	# SH-288: friendship energy persists across mid-rally grab + release-over-court.
+	# Capture a non-default rally speed, drive grab + release through _input, and assert the
```
J-Melon, line 325: too big

scripts/court/ball_tracker.gd
J-Melon: **Gabbro** Added in a1aad55: register_miss_zone before-attach, after-attach, register_miss_zone_globally via group, and unregister with kept-vs-dropped zone assertion. All four fail without prod.

scripts/court/ball_tracker.gd
J-Melon: **Gabbro** Added in a1aad55: set_partner_paddle with two live balls (current + paddle list), clear_partner_paddle drops from every ball, set with no balls then later attach inherits. Three fail without prod.

scripts/court/ball_tracker.gd
J-Melon: **Gabbro** moved the BallTracker spec note out and left a one-line pointer to designs/01-prototype/21-ball-dynamics.md (b9d7263).

scripts/court/ball_tracker.gd
J-Melon: **Gabbro** added a blank line between the missed and at_max_speed_changed connect blocks (b9d7263).

scripts/court/ball_tracker.gd
J-Melon: **Gabbro** added a blank line between the disconnect blocks to match attach (b9d7263).

scripts/items/ball_reconciler.gd
J-Melon: **Gabbro** trimmed the BallReconciler header to one line pointing at the design doc (b9d7263).

tests/unit/court/test_court_multi_ball.gd
J-Melon: **Gabbro** reshaped the test: now calls increase_speed on one ball and asserts only that ball's speed advances while the other holds steady (b9d7263).

scripts/court/ball_tracker.gd
J-Melon: **Maggie** verified at b9d7263: register_miss_zone before/after attach, register_miss_zone_globally via group, and unregister kept-vs-dropped all assert observable missed signal routing. Cleared.

scripts/court/ball_tracker.gd
J-Melon: **Maggie** verified at b9d7263: set_partner_paddle hands current ball + adds to every tracked ball's paddle list, clear removes from all, late-attach inherits. Cleared.

scripts/court/ball_tracker.gd
J-Melon: **Gru** Defer to round 6: the per-ball-ownership refactor in flight removes the `_current_ball` overwrite path, so the eager-snap question dissolves rather than gets fixed.

scripts/court/ball_tracker.gd
J-Melon: **Gru** Cheap guard, folding into round 6 along with the per-ball refactor. Will land in the same commit.

scripts/items/ball_reconciler.gd
J-Melon: **Gru** Production keeps `spawn_for_existing_on_load=false`, so the sync race is theoretical. Filing as a follow-up rather than addressing in this PR.

scripts/core/paddle_ai_controller.gd
```diff
@@ -22,14 +28,52 @@ var _last_ball_direction_x := 0.0
...
+func bind_tracker(tracker: BallTracker) -> void:
+	if _tracker == tracker:
+		return
+	if _tracker != null:
+		if _tracker.ball_added.is_connected(_on_tracker_ball_added):
+			_tracker.ball_added.disconnect(_on_tracker_ball_added)
+		if _tracker.ball_removed.is_connected(_on_tracker_ball_removed):
+			_tracker.ball_removed.disconnect(_on_tracker_ball_removed)
+	_tracker = tracker
+	if _tracker == null:
```
J-Melon, line 40: **Burns** issue: late-bind only sets `ball =`; PartnerAIController's auto-enable lives in `_on_tracker_ball_added`, so a partner recruited mid-rally never enables until the next ball attach. Call the override or enable here when an existing ball is inherited.

J-Melon:
**Stan** — block.

Re-battling head `a4187e8`. Three concrete regressions introduced by the rescale; tunable concerns are real, plus an AI-bounds bug not on the brief.

1. **`token_scale = Vector2(2.5, 2.5)` is now too large.** Tokens render under nodes whose effective world scale jumped from 0.6 → 1.0, so equipped art is **1.67× larger on screen than before**. Three call sites:
   - `scripts/items/rack_display.gd:77` — rack-slot art. Slot pitch shrank 75→45 px (X) and 100→60 px (Y), so the art-to-cell ratio grows by **2.78×**. Tokens that comfortably fit a slot pre-fix will overflow into neighbours.
   - `scripts/items/ball_reconciler.gd:196` — equipped item art on a live `Ball` parented to Court. Ball collision and `Sprite` shrank by 0.6 in the .tscn, but `ItemArtHolder.scale` is set at runtime to `token_scale`, with no compensation. Live equipped balls render at 2.5× world scale instead of 1.5×.
   - `scripts/items/ball_drag_controller.gd:261` — held-token during drag, parented to BallDragController under Court. Same 1.67× growth.

2. **`SLOT_HIT_SIZE = Vector2(60, 60)` overruns the rack frame.** Pre-fix on screen: hit area 36×36 inside a slot cell 45×60, rack frame 180×120. Post-fix: hit area **60×60** inside a slot cell **45×60**.
   - Adjacent hit areas overlap horizontally by **15 px on each side** (both X and Y centred on the marker, so neighbours intrude).
   - A row of four hit areas spans 4×60 = **240 px**, but the rack `Marker` frame is now only **180 px wide**; hit areas spill outside the painted rack.
   - End result: clicks near a slot edge resolve to the wrong item, and clicks just outside the rack still hit a slot.

3. **`paddle_size = 90.0` overrides the .tscn rescale at runtime.** `scripts/entities/paddle.gd:85` does `_collision_shape.size.y = new_size` where `new_size = clampf(get_stat("paddle_size"), 45, arena_height) = 90`. The .tscn was patched to 21.6 × 54, but `_apply_size()` snaps Y back to 90 the moment `_refresh_stats()` fires. Net effect: paddles are **67 % taller on screen than pre-fix**, and the X/Y aspect is now 21.6 × 90 instead of the intended 21.6 × 54. `BaseStatsConfig.paddle_size` (and `paddle_size_min = 45`) need the same 0.6 rescale, or the runtime apply needs to honour the new canonical size.

4. **`arena_height = 1100.0` no longer matches the walls** (`scripts/core/base_stats_config.gd:15`). Walls now live at Y = ±351.6 (vert) and the playable horizontal extent is Y = ±441.6. AI uses `arena_half = 550` to pick serve targets in `paddle_ai_controller.gd:142` and `paddle_ai_math.gd:18`, so the AI now steers paddles to Y = ±550 — **well past the back/top walls**. Same scaling problem as #3: rescale `arena_height` to 660 (or fold the 0.6 into AI math).

Tests are fine: `test_ball_regime_transitions.gd` and `test_real_input_drag_paths.gd` define their own synthetic `COURT_BOUNDS = Rect2(-600,-400,1200,800)` and `RACK_CENTER`, never reading from `court.tscn`, so the assertions stay self-consistent. No hardcoded post-rescale-stale world coords slipped through in `tests/`.

J-Melon:
**Gabbro** — fixed in `a026f48`.

Applied the 0.6 multiplier to all four:
- `token_scale`: `Vector2(2.5, 2.5)` → `Vector2(1.5, 1.5)` (`item_definition.gd`)
- `SLOT_HIT_SIZE`: `Vector2(60, 60)` → `Vector2(36, 36)` (`rack_display.gd`)
- `paddle_size`: `90.0` → `54.0` (`base_stats_config.gd`; the runtime override lives in the resource default, not `paddle.gd:85`)
- `arena_height`: `1100.0` → `660.0` (`base_stats_config.gd`)

`paddle_size_min = 45` left untouched (not in brief). ggut 495/495 green.

J-Melon:
**Stan** — still block.

Re-battling head `a026f48`. Four of the five rescale gaps from `a4187e8` are now consistent: `token_scale 2.5→1.5`, `SLOT_HIT_SIZE 60→36`, `paddle_size 90→54`, `arena_height 1100→660`. Court geometry, AI bounds, paddle .tscn (`Vector2(21.6, 54)`), and the rack frame all agree.

One miss carried forward from the prior block. Item 3 named **`paddle_size` and `paddle_size_min = 45`** together; only `paddle_size` was rescaled.

`scripts/core/base_stats_config.gd:6` — `paddle_size_min := 45.0` should be `27.0` to ride the same 0.6.

Player-facing impact: `wrist_brace.tres` (`-1.4 % per stack`) clamps against this floor. Pre-rescale, max-stacked shrink reached `90 → 45` (50 % of default). Post-rescale at the current floor, max-stacked shrink reaches only `54 → 45` (17 % of default). The shrink-paddle item is **two-thirds less effective** than before, and it shrinks paddles to a different proportion of arena height than the design assumed. `grip_tape.tres` (the inverse item) is unaffected since it grows away from the floor.

`paddle.gd:83` clamps with `clampf(get_stat("paddle_size"), paddle_size_min, arena_height)`, so the constant bites at runtime even though the default `54` clears it.

Fix: rescale to `27.0`, or fold the 0.6 into the clamp, matching the same choice made for `paddle_size`.

Tunables and the AI-bounds note from `a4187e8` are out of scope of this rescale and remain for follow-up.

scripts/court/ball_tracker.gd
J-Melon: **Gru** Resolved at `09492e2`: `set_partner_paddle` now skips the append when the paddle is already in `effect_processor.paddles`. Test `test_set_partner_paddle_twice_does_not_duplicate_in_paddle_list` covers idempotence.

scripts/court/ball_tracker.gd
J-Melon: **Gru** Resolved at `09492e2`: `attach` now preserves `_current_ball` when one already exists; only sets it if null. Test `test_attach_second_ball_mid_rally_does_not_change_current_ball` covers the mid-rally case.

scripts/items/ball_reconciler.gd
J-Melon: **Gru** Resolved at `09492e2`: `_reconcile_initial_state` moved to `call_deferred` alongside `adopt_pre_existing_balls` so the tree is settled before the call fires.

scripts/court/ball_tracker.gd
J-Melon: **Gru** Resolved at `09492e2`: trimmed the two-line `## Re-emitted from...` block to one line.

scripts/court/ball_tracker.gd
J-Melon: **Stan** resolved at 09492e2: `if not ... .has(paddle):` guard prevents duplicate append.

scripts/court/ball_tracker.gd
J-Melon: **Stan** resolved at 09492e2: `if _current_ball == null:` guard preserves mid-rally current.

scripts/items/ball_reconciler.gd
J-Melon: **Stan** resolved at 09492e2: `_reconcile_initial_state` moved to `call_deferred` inside the flag guard.

PR #516, opened 2026-04-27
Pair every code dispatch by default

J-Melon:
The paired-dispatch memory was framed as "pair when a hook forces outputs together," and Gru read that as "pair on edge cases." Wrong reading: lefthook runs ggut on every code commit, so the gate fires every commit, so the pair is the default shape.

PR #506 / SH-288 dispatched solo three times in a row and Maggie blocked on coverage gaps each time. Net rounds went up, not down. Flipping the skill to pair-by-default; solo only when explicitly justified.

### DIFF
--- a/ai/skills/gru/dispatch.md
+++ b/ai/skills/gru/dispatch.md
@@ -37,7 +37,7 @@ Every Agent call uses `run_in_background: true`. Coordinate multiple background 
 
 Three dispatch shapes for code work, picked by issue type. The cognitive separation between test and impl is the point; pick the shape that achieves it for the kind of issue at hand.
 
-### User stories — blind test-author handoff
+### User stories: blind test-author handoff
 
 For tickets with player-observable ACs ("ball appears on rack after buy," "drag from court back to rack works mid-rally"):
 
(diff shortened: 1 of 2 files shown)


ai/skills/gru/dispatch.md
```diff
@@ -35,7 +35,35 @@ Every Agent call uses `run_in_background: true`. Coordinate multiple background
 
 ## Paired dispatch
 
-Pair specialists when a hook or gate forces their outputs into one commit (failing tests + impl, scene + script). Otherwise dispatch independent. Paired dispatch costs parallelism on that pair; default to independent.
+Three dispatch shapes for code work, picked by issue type. The cognitive separation between test and impl is the point; pick the shape that achieves it for the kind of issue at hand.
+
+### User stories — blind test-author handoff
+
```
J-Melon, line 41: **Hibbert** issue: em dash in heading; STYLE bans them anywhere. Use a colon: `### User stories: blind test-author handoff`. Same fix on the System stories and Bugs headings.

J-Melon: **Gru** Colon-swap on the three headings in the next push; recurring em-dash slip noted.

PR #533, opened 2026-04-28
SH-287 drag drops validated by body projection

J-Melon:
Refactors the drag controller's release path so every drop is validated by a body projection (`PhysicsDirectSpaceState2D.intersect_shape`) against the at-rest body's authored collision shape, polled per physics frame on the held token's world position. The first registered drop target whose `can_accept` returns true commits; otherwise the gesture stays open. Establishes the seam SH-314 and SH-297 will branch off.

Spec: `designs/01-prototype/21-ball-dynamics.md` (Containers and the swap pattern, Drop validation by body projection, No restore on invalid release) and `designs/01-prototype/22-equip-loop-regime.md` (Failure modes: expansion-ring fallback after a 250 ms hold, then cancel-to-source).

What lands:

- `DropTarget` interface (`can_accept`, `accept`) plus `CourtDropTarget`, `RackDropTarget`, `ShopDropTarget`, `VenueDropTarget`.
- `ItemDefinition.at_rest_shape: Shape2D`. `base_ball.tres` and `training_ball.tres` author a `CircleShape2D` matching the live-ball collider.
- `BallDragController` polls every registered target each frame; auto-commits on first accepting target. No teleport-restore on mouse-up.
- Expansion-ring fallback: 0.25 s hold then a 1.5x scaled projection retry; second hold-window fail cancels back to source.
- Hover feedback bumps the held token's scale and modulation over a valid target.
- Shop-purchase releases call `BallDragController.spawn_purchased_at` so the new ball lands on the court at the released cursor point. **Verified SH-320 closed** by the integration test in `tests/integration/test_shop_drag_drop.gd::test_shop_to_court_release_spawns_ball_at_release_position`.

Note on the brief's "remove `Court._on_paddle_hit` null-guard" item: the null-guard is no longer present in `scripts/core/court.gd` on `main` (predecessor work removed it). Bespoke `_position_over_rack`, `_clamp_to_court`, `_release_onto_court`, and the role-check branch in the controller are gone. 526 tests, 85.9% coverage.

Out of scope: held-body type stays `Node2D` (RigidBody2D + gravity is SH-314); ease-to-cursor / cursor-state textures / generous press hit-box is SH-297.

Closes SH-287. Verified to also close SH-320.

### DIFF
--- a/scripts/items/ball_drag_controller.gd
+++ b/scripts/items/ball_drag_controller.gd
@@ -76,10 +76,10 @@ var _held_preserved_speed: float = PRESERVED_SPEED_NONE
 ## (i.e. mouse-up with strict projection failing). Negative means not yet timing.
 var _expansion_started_at: float = -1.0
 
-var _drop_targets: Array = []
+var _drop_targets: Array[DropTarget] = []
 ## Targets the controller authored from its own exports; rebuilt on `_ready` and not
 ## removed by the public `unregister_target` API.
-var _builtin_targets: Array = []
+var _builtin_targets: Array[DropTarget] = []
 
 
 func configure(
@@ -180,7 +180,7 @@ func unregister_target(target: DropTarget) -> void:
 	_drop_targets.erase(target)
 
 
-func get_registered_targets() -> Array:
+func get_registered_targets() -> Array[DropTarget]:
 	return _drop_targets.duplicate()
 
 
@@ -325,10 +325,9 @@ func _apply_preserved_speed_after_accept(item_key: String) -> void:
 ## Returns the first registered target whose `can_accept` succeeds, or null. Targets are
 ## polled in registration order; built-ins register in priority order (court before venue).
 func _find_accepting_target(item_key: String, position: Vector2, scale_factor: float) -> DropTarget:
-	for target in _drop_targets:
-		var dt: DropTarget = target
-		if dt.can_accept(item_key, position, scale_factor):
-			return dt
+	for target: DropTarget in _drop_targets:
+		if target.can_accept(item_key, position, scale_factor):
+			return target
 	return null
 
 
--- a/scripts/items/ball_drag_controller.gd
+++ b/scripts/items/ball_drag_controller.gd
@@ -234,19 +234,21 @@ func grab_live_ball(item_key: String, is_temporary: bool = false) -> bool:
 
 
 ## Shop-purchase entry: the shop just took payment for `item_key` and wants the new ball
-## to spawn on the court at `position` (SH-320). Runs the standard target poll; if a
+## to spawn on the court at `world_position` (SH-320). Runs the standard target poll; if a
 ## court/venue target accepts, the ball spawns and we return true. Otherwise the item
 ## stays a rack-only token (the rack regrows it via `_on_court_changed` defaults).
-func spawn_purchased_at(item_key: String, position: Vector2, gesture_velocity: Vector2) -> bool:
+func spawn_purchased_at(
+	item_key: String, world_position: Vector2, gesture_velocity: Vector2
+) -> bool:
 	# No venue clamp here: the shop has already paid for the item, and a release outside
 	# the venue means the rack should take the new token. Returning false lets the caller
 	# fall back to its own rack/no-court path.
-	var target: DropTarget = _find_accepting_target(item_key, position, 1.0)
+	var target: DropTarget = _find_accepting_target(item_key, world_position, 1.0)
 	if target == null:
 		return false
 	if not (target is CourtDropTarget or target is VenueDropTarget):
 		return false
-	target.accept(item_key, position, gesture_velocity)
+	target.accept(item_key, world_position, gesture_velocity)
 	return true
 
 
@@ -324,18 +326,20 @@ func _apply_preserved_speed_after_accept(item_key: String) -> void:
 
 ## Returns the first registered target whose `can_accept` succeeds, or null. Targets are
 ## polled in registration order; built-ins register in priority order (court before venue).
-func _find_accepting_target(item_key: String, position: Vector2, scale_factor: float) -> DropTarget:
+func _find_accepting_target(
+	item_key: String, world_position: Vector2, scale_factor: float
+) -> DropTarget:
 	for target: DropTarget in _drop_targets:
-		if target.can_accept(item_key, position, scale_factor):
+		if target.can_accept(item_key, world_position, scale_factor):
 			return target
 	return null
 
 
 ## Hover-feedback bump applied to the held token while a target accepts the current pos.
-func _update_hover_feedback(position: Vector2) -> void:
+func _update_hover_feedback(world_position: Vector2) -> void:
 	if _held_token == null:
 		return
-	var hovering: bool = _find_accepting_target(_held_key, position, 1.0) != null
+	var hovering: bool = _find_accepting_target(_held_key, world_position, 1.0) != null
 	var definition: ItemDefinition = _get_item_definition(_held_key)
 	var base_scale: Vector2 = definition.token_scale if definition != null else Vector2.ONE
 	if hovering:
@@ -348,7 +352,7 @@ func _update_hover_feedback(position: Vector2) -> void:
 
 ## After mouse-up: maintain the expansion-ring timer. Cancels back to source if the
 ## widened poll has also failed for the same hold window.
-func _update_expansion_state(position: Vector2) -> void:
+func _update_expansion_state(world_position: Vector2) -> void:
 	if _held_token == null:
 		return
 	if _expansion_started_at < 0.0:
@@ -362,11 +366,13 @@ func _update_expansion_state(position: Vector2) -> void:
 	# Strict pass already ran in attempt_release; try the widened pass now. If it succeeds,
 	# attempt_release picks it up next frame. If the widened pass has also been failing for
 	# another full hold window, cancel back to source.
-	var widened: DropTarget = _find_accepting_target(_held_key, position, EXPANSION_RING_SCALE)
+	var widened: DropTarget = _find_accepting_target(
+		_held_key, world_position, EXPANSION_RING_SCALE
+	)
 	if widened != null:
 		# attempt_release on the next frame will use the widened scale because the timer
 		# crossed the threshold; re-running it now lets us commit immediately.
-		attempt_release(position)
+		attempt_release(world_position)
 		return
 
 	if held_duration >= EXPANSION_RING_HOLD_S * 2.0:
--- a/scripts/items/ball_drag_controller.gd
+++ b/scripts/items/ball_drag_controller.gd
@@ -30,8 +30,7 @@ var _held_key: String = ""
 var _held_is_temporary: bool = false
 ## Was the item on-court before the gesture? Rack pickups defer activation, so a click-without-movement is a no-op.
 var _held_was_on_court: bool = false
-## Origin tag for the current gesture: &"rack" (default), &"live" (mid-rally grab).
-## Drives the SH-252 click-without-movement no-op gate (rack-only) and future hooks.
+## &"rack" or &"live"; rack origins gate the SH-252 click-without-movement no-op.
 var _held_origin: StringName = &"rack"
 var _cursor_samples: Array = []
 ## Cursor position when the held token spawned; gates the SH-252 a click-without-movement no-op.
@@ -39,16 +38,13 @@ var _press_position: Vector2 = Vector2.ZERO
 var _gesture_below_threshold: bool = true
 ## SH-287: tracks mouse-button state so _process can poll for valid targets when mouse is up.
 var _mouse_button_down: bool = false
-## SH-288: friendship energy captured at mid-rally grab; forwarded to bring_into_play so the
-## released ball inherits the rally speed. Negative means no preserved energy.
+## Mid-rally rally speed inherited by the released ball; negative means none preserved.
 var _held_preserved_speed: float = PRESERVED_SPEED_NONE
-## SH-287: monotonic time in seconds when the controller began trying expansion-ring polling
-## (i.e. mouse-up with strict projection failing). Negative means not yet timing.
+## Monotonic seconds when expansion-ring polling started; negative means not yet timing.
 var _expansion_started_at: float = -1.0
 
 var _drop_targets: Array[DropTarget] = []
-## Targets the controller authored from its own exports; rebuilt on `_ready` and not
-## removed by the public `unregister_target` API.
+## Built-ins are rebuilt on `_ready` and ignored by `unregister_target`.
 var _builtin_targets: Array[DropTarget] = []
 
 
@@ -68,8 +64,7 @@ func _ready() -> void:
 	if _item_manager == null:
 		_item_manager = ItemManager
 
-	# Group lookup so Shop (and any future origin) can hand presses to the controller
-	# without an explicit NodePath in court.tscn.
+	# Group lookup so Shop can hand presses to the controller without a NodePath.
 	add_to_group(&"drag_controller")
 
 	if rack != null and not rack.slot_pressed.is_connected(_on_rack_slot_pressed):
@@ -96,8 +91,6 @@ func _process(_delta: float) -> void:
 		if follow_position.distance_to(_press_position) >= COMMIT_MOVEMENT_THRESHOLD_PX:
 			_gesture_below_threshold = false
 
-	# SH-287: when mouse is up and the held position is over a valid target, commit.
-	# Otherwise keep the held token following the cursor and update hover feedback.
 	if not _mouse_button_down:
 		if not attempt_release(follow_position):
 			_update_expansion_state(follow_position)
@@ -217,8 +210,7 @@ func spawn_purchased_at(
 	return true
 
 
-## Try to commit the held gesture at the given position. Returns true on commit (held token
-## freed, gesture ends), false on no valid target (held token stays, gesture continues).
+## Returns false on no valid target so the held token stays with the cursor.
 func attempt_release(release_position: Vector2) -> bool:
 	if _held_token == null:
 		return false
@@ -239,8 +231,6 @@ func attempt_release(release_position: Vector2) -> bool:
 		_finalise_gesture(item_key, clamped_position, false)
 		return true
 
-	# Strict pass first; on miss the caller (process loop) decides whether to start the
-	# expansion-ring timer.
 	var target: DropTarget = _find_accepting_target(item_key, clamped_position, 1.0)
 	if target == null and _expansion_started_at >= 0.0:
 		var held_duration: float = _now_seconds() - _expansion_started_at
@@ -282,8 +272,7 @@ func _apply_preserved_speed_after_accept(item_key: String) -> void:
 		ball.linear_velocity = ball.linear_velocity.normalized() * _held_preserved_speed
 
 
-## Returns the first registered target whose `can_accept` succeeds, or null. Targets are
-## polled in registration order; built-ins register in priority order (court before venue).
+## Polled in registration order; built-ins register court before venue.
 func _find_accepting_target(
 	item_key: String, world_position: Vector2, scale_factor: float
 ) -> DropTarget:
@@ -308,8 +297,7 @@ func _update_hover_feedback(world_position: Vector2) -> void:
 		_held_token.modulate = NEUTRAL_MODULATE
 
 
-## After mouse-up: maintain the expansion-ring timer. Cancels back to source if the
-## widened poll has also failed for the same hold window.
+## Cancels back to source after the widened poll has also failed for one full hold window.
 func _update_expansion_state(world_position: Vector2) -> void:
 	if _held_token == null:
 		return
@@ -321,15 +309,11 @@ func _update_expansion_state(world_position: Vector2) -> void:
 	if held_duration < expansion_ring_hold_s:
 		return
 
-	# Strict pass already ran in attempt_release; try the widened pass now. If it succeeds,
-	# attempt_release picks it up next frame. If the widened pass has also been failing for
-	# another full hold window, cancel back to source.
 	var widened: DropTarget = _find_accepting_target(
 		_held_key, world_position, expansion_ring_scale
 	)
 	if widened != null:
-		# attempt_release on the next frame will use the widened scale because the timer
-		# crossed the threshold; re-running it now lets us commit immediately.
+		# Re-run now so the commit lands this frame instead of waiting for the next attempt_release.
 		attempt_release(world_position)
 		return
 
@@ -337,8 +321,7 @@ func _update_expansion_state(world_position: Vector2) -> void:
 		_cancel_to_source()
 
 
-## Two-window expansion fail -> cancel-to-source. Free the held token; deactivate any
-## on-court placement that a live-ball grab would have moved so the rack regrows it.
+## Live-ball grabs that cancel must deactivate the on-court placement so the rack regrows it.
 func _cancel_to_source() -> void:
 	var item_key: String = _held_key
 	var was_on_court: bool = _held_was_on_court
--- a/scripts/items/ball_drag_controller.gd
+++ b/scripts/items/ball_drag_controller.gd
@@ -3,20 +3,48 @@ extends Node2D
 
 ## Owns the held-token visual during a ball or equipment drag gesture.
 ##
-## Press starts a hold (held token follows cursor); release commits only at a valid target.
-## No-restore principle (SH-287): mouse-up over an invalid spot does not end the gesture;
-## the held token keeps following the cursor and the controller polls per frame for a valid
-## target. The source rack is itself a target, giving the player a no-teleport escape valve.
+## SH-287 (designs/01-prototype/21-ball-dynamics.md, "Drop validation by body projection"
+## and "No restore on invalid release"; designs/01-prototype/22-equip-loop-regime.md,
+## "Failure modes"):
+##
+## * Press starts a hold (held token follows cursor).
+## * The controller polls every registered `DropTarget` each physics frame on the held
+##   token's world position; the first target whose `can_accept` returns true commits.
+## * Mouse-up over an invalid spot does not end the gesture; the held token keeps following
+##   the cursor and the controller keeps polling. The source rack/shop is itself a target,
+##   giving the player a no-teleport escape valve.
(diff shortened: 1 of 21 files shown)


scripts/shop/shop.gd
```diff
@@ -24,6 +27,25 @@ func _ready() -> void:
...
+
+func _do_register_shop_target() -> void:
+	var controller: Node = get_tree().get_first_node_in_group(&"drag_controller")
+	if controller == null:
+		return
+	if not controller.has_method("register_target"):
+		return
+	var target: ShopDropTarget = ShopDropTargetScript.new()
+	target.configure(shop_area)
+	controller.register_target(target)
```
J-Melon, line 47: **signals-lifecycle**

issue: `ShopDropTarget` is registered on the controller as a strong reference but never deregistered when the Shop is freed. The target captures `shop_area: Area2D`; if the Shop subtree leaves the tree (scene change, dev-panel respawn, or the very `_shop.queue_free(); ShopScene.instantiate()` pattern used in `tests/integration/test_shop_drag_drop.gd`), the controller still holds the old `ShopDropTarget` and its next `can_accept` poll dereferences `_shop_area.global_position` on a freed `Area2D`. There is no `_exit_tree` hook here calling `controller.unregister_target(target)`, no `is_instance_valid(_shop_area)` guard in `shop_drop_target.gd`, and no `tree_exiting` wiring. Suggest: store the registered target on Shop, connect `tree_exiting` (still safe to touch the controller) and call `unregister_target` there. Belt-and-braces: guard `_shop_area` with `is_instance_valid` in `ShopDropTarget`.

scripts/items/drop_targets/shop_drop_target.gd
```diff
@@ -0,0 +1,45 @@
...
+		if child is CollisionShape2D:
+			shape_owner = child
+			break
+	if shape_owner == null:
+		return false
+	var rectangle: RectangleShape2D = shape_owner.shape as RectangleShape2D
+	if rectangle == null:
+		return false
+	var half_extents: Vector2 = rectangle.size * 0.5
+	var center: Vector2 = _shop_area.global_position + shape_owner.position
```
J-Melon: **signals-lifecycle**

issue (nitpick, non-blocking): `_position_inside_area` reaches into `_shop_area.global_position` and `_shop_area.get_children()` with no validity guard. Pair this with the Shop-side deregistration on `tree_exiting`; together they make a freed-instance access on this path impossible. Today only the early `if _shop_area == null` check fires, and a queue_freed-but-still-referenced area passes that null check.

scripts/items/ball_drag_controller.gd
```diff
@@ -114,6 +164,26 @@ func get_held_token() -> Node2D:
...
+func unregister_target(target: DropTarget) -> void:
+	if _builtin_targets.has(target):
+		return
+	_drop_targets.erase(target)
+
+
+func get_registered_targets() -> Array:
+	return _drop_targets.duplicate()
+
+
```
J-Melon, line 149: **signals-lifecycle**

question: `unregister_target` silently no-ops when the caller passes a built-in (`if _builtin_targets.has(target): return`). That means external callers cannot remove the target they just registered if it happens to alias a built-in, and there is no error/warn path. For the Shop deregistration story (see comment on `shop.gd` line 48), the `target` will be the externally-created `ShopDropTarget`, so this works in practice — but the silent-skip-on-built-in branch is undocumented in the public surface. Worth either a one-line doc comment or pushing built-ins into a separate list with a clearer name (e.g. `_drop_targets` for built-ins and `_external_targets` for the public-API ones).

tests/unit/items/test_ball_drag_controller_sh287.gd
```diff
@@ -0,0 +1,174 @@
...
+
+	_drag._gesture_below_threshold = false
+	_drag._expansion_started_at = (
+		float(Time.get_ticks_msec()) / 1000.0 - _drag.EXPANSION_RING_HOLD_S - 0.05
+	)
+	var target_strict: DropTarget = _drag._find_accepting_target("ball_alpha", Vector2(0, 0), 1.0)
+	assert_not_null(target_strict, "strict probe accepts an empty court")
+	var target_widened: DropTarget = _drag._find_accepting_target("ball_alpha", Vector2(0, 0), 1.5)
+	assert_not_null(target_widened, "widened probe also accepts an empty court")
+
```
J-Melon, line 147: **test-coverage** issue: this asserts the helper plumbs scale through, not the named behaviour; both probes accept on an empty court so the test passes even if EXPANSION_RING_SCALE were 1.0. Drive a position where strict (1.0) fails and widened (1.5) succeeds (e.g. wall-grazed), then assert the gesture commits via the timed path.

tests/unit/items/test_ball_drag_controller_sh287.gd
```diff
@@ -0,0 +1,174 @@
...
+		ball.queue_free()
+	await get_tree().process_frame
+
+	_drag._gesture_below_threshold = false
+	_drag._mouse_button_down = false
+	_drag._expansion_started_at = (
+		float(Time.get_ticks_msec()) / 1000.0 - _drag.EXPANSION_RING_HOLD_S * 2.0 - 0.1
+	)
+	_drag._cancel_to_source()
+	assert_false(_drag.is_dragging(), "cancel-to-source frees the held token")
```
J-Melon: **test-coverage** issue: this calls _cancel_to_source directly, bypassing the only production caller (_update_expansion_state). The test passes even if the timer never invokes cancel. Drive _process or _update_expansion_state with a stub clock so the player behaviour ('hold invalid spot ~500ms with mouse up cancels') is what gets asserted.

tests/unit/items/test_ball_drag_controller_sh287.gd
```diff
@@ -0,0 +1,174 @@
...
+func test_hover_feedback_bumps_held_token_scale_over_valid_target() -> void:
+	_manager.take("ball_alpha")
+	_drag.grab_from_rack("ball_alpha")
+	var token: Node2D = _drag.get_held_token()
+	assert_not_null(token, "precondition: held token spawned")
+	# An empty-court position is a valid target; mouse-down (still grabbing) so hover
+	# feedback applies (not the auto-commit branch).
+	_drag._mouse_button_down = true
+	_drag._update_hover_feedback(Vector2(0, 0))
+	assert_ne(token.scale, Vector2(1.5, 1.5), "scale lifts when hovering a valid target")
```
J-Melon: **test-coverage** suggestion: assert_ne against (1.5, 1.5) passes for any non-equal scale, including a regression that shrinks the token. Capture token.scale before _update_hover_feedback and assert it grew (length() greater than the captured base), so the player-visible 'lift' is what's pinned.

tests/unit/items/test_drop_targets.gd
```diff
@@ -0,0 +1,317 @@
...
+
+
+func test_item_definition_carries_at_rest_shape_for_ball_items() -> void:
+	# Authored ball definitions land an at-rest shape so CourtDropTarget projection has
+	# a real radius to query against.
+	assert_not_null(BaseBall.at_rest_shape, "base ball should carry an at_rest_shape after SH-287")
+	assert_true(BaseBall.at_rest_shape is CircleShape2D)
+
+
+func test_token_scale_remains_canonical_across_items() -> void:
```
J-Melon, line 299: **test-coverage** issue: the comment promises cross-container size identity but the assertion is a single literal against one resource. To match the brief, also build a held token via the controller and a slot token via the rack and assert their scales equal item_definition.token_scale; otherwise this is just pinning a literal.

scripts/items/ball_drag_controller.gd
```diff
@@ -156,66 +227,167 @@ func grab_live_ball(item_key: String, is_temporary: bool = false) -> bool:
...
+	if hovering:
+		_held_token.scale = base_scale * HOVER_SCALE_BUMP
+		_held_token.modulate = HOVER_MODULATE
+	else:
+		_held_token.scale = base_scale
+		_held_token.modulate = NEUTRAL_MODULATE
+
+
+## After mouse-up: maintain the expansion-ring timer. Cancels back to source if the
+## widened poll has also failed for the same hold window.
```
J-Melon: **test-coverage** issue: _update_expansion_state (lines 345-373) is the production driver of the expansion-ring + cancel-after-two-windows behaviour and is entirely uncovered. The two SH-287 expansion-ring tests poke helpers around it. Add one test that lets _process tick this with mouse up and a strict-fail position, asserting commit on widened-pass and cancel on widened-fail.

scripts/items/drop_targets/court_drop_target.gd
```diff
@@ -0,0 +1,134 @@
...
+		),
+		clampf(
+			world_position.y,
+			_court_bounds.position.y,
+			_court_bounds.position.y + _court_bounds.size.y
+		),
+	)
+
+
+func _get_definition(item_key: String) -> ItemDefinition:
```
J-Melon: **code-quality**

issue: `_get_definition` is duplicated verbatim in `court_drop_target.gd`, `rack_drop_target.gd`, and `venue_drop_target.gd`. Lift it onto the `DropTarget` base or a shared helper.

scripts/items/drop_targets/venue_drop_target.gd
```diff
@@ -0,0 +1,85 @@
...
+
+
+func _is_ball_role(item_key: String) -> bool:
+	var definition: ItemDefinition = _get_definition(item_key)
+	if definition == null:
+		return true
+	return definition.role == &"ball"
+
+
+func _clamp_to_court(world_position: Vector2) -> Vector2:
```
J-Melon: **code-quality**

issue: `_clamp_to_court` is byte-identical here and in `court_drop_target.gd:111`. Either share via the base class or have one target delegate to the other.

scripts/items/drop_targets/shop_drop_target.gd
```diff
@@ -0,0 +1,45 @@
...
+	return _position_inside_area(position)
+
+
+func accept(_item_key: String, _position: Vector2, _gesture_velocity: Vector2) -> void:
+	# No side effect on cancel-back: the gesture clears and the source slot becomes visible
+	# again via the controller's finalisation hook.
+	pass
+
+
+func _position_inside_area(world_position: Vector2) -> bool:
```
J-Melon: **code-quality**

issue: rectangle-from-Area2D extraction is duplicated here and in `rack_drop_target.gd._world_rect()`. One helper on the base class would dedupe both.

scripts/items/ball_drag_controller.gd
```diff
@@ -3,20 +3,48 @@ extends Node2D
...
+##   with a 1.5x scaled shape; if that also fails, cancel the gesture back to the source.
+## * Hover feedback: the held token's modulation/scale lifts when over a valid target.
+## * Shop releases call `spawn_purchased_at` after committing the purchase, threading the
+##   new ball through the same target-poll so shop-to-court drag spawns a live ball at the
+##   release point instead of routing to the rack (SH-320).
 
 signal pickup_started(item_key: String)
 signal drop_completed(item_key: String, release_position: Vector2, over_court: bool)
 
+const CourtDropTargetScript: GDScript = preload(
```
J-Melon: **code-quality**

nitpick: the `Script` suffix is redundant since each target already declares `class_name`; `const CourtDropTargetScene = preload(...)` style is the project pattern, or use the class name directly.

scripts/shop/shop.gd
J-Melon: **manny**

resolved: Shop now wires `tree_exiting` to `_on_tree_exiting`, which calls `unregister_target` on the controller and clears the cached references. Scene reloads no longer leak stale ShopDropTargets (7792294).

scripts/items/drop_targets/shop_drop_target.gd
J-Melon: **manny**

resolved: `ShopDropTarget.can_accept` now guards with `is_instance_valid(_shop_area)` before any access; the rectangle extraction lives on the `DropTarget` base via `area_world_rect` which carries the same guard (7792294).

scripts/items/ball_drag_controller.gd
J-Melon: **manny**

response: confirmed deliberate. The typed-array commit (38781e1) split storage into `_drop_targets` and `_builtin_targets`, so the silent no-op now reads as "external callers cannot remove built-ins, by design." Going with the explicit split rather than renaming `unregister_target` to `unregister_external_target`: the call site in Shop is symmetric with `register_target` and renaming would break that pairing. The split makes the policy visible at the storage layer; the no-op on built-ins is the enforcement.

tests/unit/items/test_ball_drag_controller_sh287.gd
J-Melon: **manny**

resolved: replaced with `test_expansion_ring_scale_genuinely_widens_the_probe`. A wall is now placed so the strict 1.0x circle (radius 10) clears the gap (14) but the 1.5x widened probe (radius 15) overlaps and rejects. The test fails if `EXPANSION_RING_SCALE` regresses to 1.0 (7792294).

tests/unit/items/test_ball_drag_controller_sh287.gd
J-Melon: **manny**

resolved: `test_expansion_ring_cancel_after_two_holds_fails_to_source` now drives `_update_expansion_state` (the production caller) with a release point off both court and venue, then advances the timer past `EXPANSION_RING_HOLD_S * 2`. The test fails if the timer never invokes cancel. Added `test_expansion_state_does_not_cancel_within_first_window` and `test_expansion_state_commits_when_widened_probe_accepts` to pin the boundary and the widened-accept branch (7792294).

tests/unit/items/test_ball_drag_controller_sh287.gd
J-Melon: **manny**

resolved: hover-feedback test now `assert_eq`s the exact `base_scale * HOVER_SCALE_BUMP` and the `HOVER_MODULATE` constant. Added `test_hover_feedback_resets_to_base_scale_when_no_target_accepts` to pin the inverse (off-target reset) (7792294).

tests/unit/items/test_drop_targets.gd
J-Melon: **manny**

resolved: `test_token_scale_remains_canonical_across_items` now exercises the full chain: spawns a held token via `BallDragController._spawn_held_token`, refreshes a `RackDisplay` to populate the slot ArtHolder, and asserts both renderings equal `BaseBall.token_scale`. The single-literal pin remains as the source-of-truth assertion (7792294).

scripts/items/ball_drag_controller.gd
J-Melon: **manny**

resolved: `_update_expansion_state` is now exercised by three tests in `test_ball_drag_controller_sh287.gd`: cancel-after-two-windows (drives the timer past 2x hold), no-cancel-within-first-window (boundary pin), and commit-on-widened-accept (the success branch that re-invokes `attempt_release`). `ball_drag_controller.gd` coverage moved from 81.3% to 86.9% (7792294).

scripts/items/drop_targets/court_drop_target.gd
J-Melon: **manny**

resolved: lifted `_get_definition` onto `DropTarget` as the static `DropTarget.get_definition(item_manager, item_key)`. Court, rack, and venue targets now call the shared helper; the duplicates are gone (7792294).

scripts/items/drop_targets/venue_drop_target.gd
J-Melon: **manny**

resolved: lifted court-bounds clamping onto `DropTarget` as the static `DropTarget.clamp_to_rect(world_position, bounds)`. `CourtDropTarget`, `VenueDropTarget`, and `BallDragController._clamp_to_venue` all delegate to it; the byte-identical copy is gone (7792294).

scripts/items/drop_targets/shop_drop_target.gd
J-Melon: **manny**

resolved: lifted Area2D rectangle extraction onto `DropTarget` as the static `DropTarget.area_world_rect(area)`. `RackDropTarget` and `ShopDropTarget` both delegate; the helper carries the `is_instance_valid` guard, so the shop-area-freed case is also covered (7792294).

scripts/items/ball_drag_controller.gd
J-Melon: **manny**

resolved: dropped the `Script` suffix from the preload constants. `BallDragController` now uses `CourtDropTarget.new()` / `RackDropTarget.new()` / `VenueDropTarget.new()` directly via `class_name`. Same in `Shop` for `ShopDropTarget` (7792294).

scripts/items/ball_drag_controller.gd
```diff
@@ -3,10 +3,22 @@ extends Node2D
...
-## target. The source rack is itself a target, giving the player a no-teleport escape valve.
+## SH-287 (designs/01-prototype/21-ball-dynamics.md, "Drop validation by body projection"
+## and "No restore on invalid release"; designs/01-prototype/22-equip-loop-regime.md,
+## "Failure modes"):
+##
+## * Press starts a hold (held token follows cursor).
+## * The controller polls every registered `DropTarget` each physics frame on the held
+##   token's world position; the first target whose `can_accept` returns true commits.
+## * Mouse-up over an invalid spot does not end the gesture; the held token keeps following
+##   the cursor and the controller keeps polling. The source rack/shop is itself a target,
```
J-Melon: Big comment

scripts/items/ball_drag_controller.gd
```diff
@@ -15,8 +27,14 @@ const CURSOR_SAMPLE_WINDOW: float = 0.08
 const PRESERVED_SPEED_NONE: float = -1.0
 ## Minimum cursor travel before a rack-origin gesture counts as a real drag (SH-252 a).
 const COMMIT_MOVEMENT_THRESHOLD_PX: float = 6.0
-## SH-287 patch: probe radius for the pre-spawn body projection; slightly larger than authored ball radius.
-const COURT_BLOCKED_PROBE_RADIUS_PX: float = 14.0
+## Seconds the controller waits with the strict projection failing before widening to
+## the expansion ring. Tuned per `22-equip-loop-regime.md` "Failure modes".
+const EXPANSION_RING_HOLD_S: float = 0.25
```
J-Melon: Should be a config

scripts/items/ball_drag_controller.gd
```diff
@@ -156,66 +217,173 @@ func grab_live_ball(item_key: String, is_temporary: bool = false) -> bool:
...
+	_held_origin = &"live"
 	_mouse_button_down = true
 	pickup_started.emit(item_key)
 	return true
 
 
-## Try to commit the held gesture at the given position. Returns true on commit (held token freed,
-## gesture ends), false on no valid target (held token stays, gesture continues).
+## Shop-purchase entry: the shop just took payment for `item_key` and wants the new ball
+## to spawn on the court at `world_position` (SH-320). Runs the standard target poll; if a
```
J-Melon: Big comment

scripts/items/ball_drag_controller.gd
```diff
@@ -156,66 +217,173 @@ func grab_live_ball(item_key: String, is_temporary: bool = false) -> bool:
...
+	if target == null:
+		return false
 
-	# No valid target. Held token stays following the cursor; gesture continues.
-	return false
+	if target is CourtDropTarget or target is VenueDropTarget:
+		var velocity: Vector2 = _compute_release_velocity()
+		if was_temporary:
+			# Temporary balls bypass the reconciler; calling target.accept would spawn a
+			# permanent ball through the reconciler. Drop the held token cleanly instead.
```
J-Melon: Lots of comments

scripts/items/ball_drag_controller.gd
```diff
@@ -270,10 +421,44 @@ func _spawn_held_token(item_key: String, spawn_position: Vector2, is_temporary:
...
+		_drop_targets.append(ball_rack_target)
+		_builtin_targets.append(ball_rack_target)
+
+	if gear_rack_drop_target != null:
+		var gear_rack_target: RackDropTarget = RackDropTarget.new()
+		gear_rack_target.configure(_item_manager, gear_rack_drop_target, &"equipment")
+		_drop_targets.append(gear_rack_target)
+		_builtin_targets.append(gear_rack_target)
+
+	if reconciler != null:
```
J-Melon: You check for reconciler null twice here, anything more holistic?

scripts/shop/shop_item.gd
```diff
@@ -165,12 +165,38 @@ func attempt_release(release_position: Vector2) -> bool:
 		# bounds to cancel, or get more FP and try again.
 		return false
 
-	# Purchased: held token freed, slot stays hidden until the next shop refresh sweeps it.
+	# Purchased: hand the held position off to BallDragController so the body-projection
+	# target loop decides whether the new ball lands on the court (SH-287, SH-320 fix) or
+	# falls through to the rack as a token. Held visual freed by the controller's
```
J-Melon: Comment

scripts/items/ball_drag_controller.gd
J-Melon: **manny**

resolved: collapsed the class docstring (old lines 4-22) to a single WHY line per the code-comments policy.

scripts/items/ball_drag_controller.gd
J-Melon: **manny**

resolved: promoted `EXPANSION_RING_HOLD_S` and `EXPANSION_RING_SCALE` to `@export var expansion_ring_hold_s` and `expansion_ring_scale` on the controller, so the projection-fallback timing is per-instance tunable from the scene. Tests updated to read the new exports.

scripts/items/ball_drag_controller.gd
J-Melon: **manny**

resolved: collapsed the `spawn_purchased_at` docstring and trimmed the inline 'no venue clamp' note to one line.

scripts/items/ball_drag_controller.gd
J-Melon: **manny**

resolved: removed the inline narration block (was-temporary, reconciler signature, double-spawn). Kept one short WHY on the temporary-bypass branch and dropped the rest. Also collapsed `_apply_preserved_speed_after_accept` docstring.

scripts/items/ball_drag_controller.gd
J-Melon: **manny**

resolved: extracted three target-factory helpers (`_make_court_target`, `_make_rack_target`, `_make_venue_target`); each guards its own preconditions and returns target-or-null. `_register_builtin_targets` now appends-if-non-null in priority order with zero repeated null checks. If you'd rather see a single cached `var has_reconciler = reconciler != null` instead, easy to swap.

scripts/shop/shop_item.gd
J-Melon: **manny**

resolved: dropped the multi-line 'hand off to BallDragController' block and collapsed the `_route_purchased_to_court` and `attempt_release` docstrings to single WHY lines.

scripts/items/ball_drag_controller.gd
```diff
@@ -25,22 +21,35 @@ const COURT_BLOCKED_PROBE_RADIUS_PX: float = 14.0
 @export var court_bounds: Rect2 = Rect2()
 @export var venue_bounds: Rect2 = Rect2()
 @export var reconciler: BallReconciler
+@export var expansion_ring_hold_s: float = 0.25
+@export var expansion_ring_scale: float = 1.5
```
J-Melon, line 25: **gdscript-conventions** *suggestion (non-blocking):* `expansion_ring_hold_s` and `expansion_ring_scale` are now editor-exposed per-instance. Consider `@export_range(0.0, 2.0, 0.01) var expansion_ring_hold_s` and `@export_range(1.0, 3.0, 0.05) var expansion_ring_scale` so the inspector enforces the semantic constraints (negative hold inverts the timer gate, scale < 1.0 inverts the "expansion" semantics). Cheap belt-and-braces against an inspector typo.

scripts/items/drop_targets/venue_drop_target.gd
```diff
@@ -0,0 +1,59 @@
+class_name VenueDropTarget
+extends DropTarget
+
+## Venue floor catch-all target.
+##
+## SH-287 keeps held items as Node2D tokens (no loose-physics body yet; that arrives in
+## SH-314). The venue target accepts inside the venue rect when no other target does:
+## ball-role items spawn on the court at the court-clamped position (preserving the
+## prior "release inside venue but outside court spawns at the court edge" behaviour);
```
J-Melon: Massive comment

J-Melon: **manny**

resolved: collapsed the venue_drop_target class docstring (was lines 4-16) to a single WHY line, and swept the other three drop_target scripts for any surviving multi-line comments — all class docstrings and inline blocks are now one line WHY-only per the code-comments policy.

tests/unit/items/test_drop_targets.gd
```diff
@@ -0,0 +1,352 @@
...
+
+
+func test_court_target_widens_with_expansion_ring_scale() -> void:
+	var manager: Node = ItemFactory.create_manager(self)
+	var ball_alpha: ItemDefinition = _make_ball_definition("ball_alpha", 12.0)
+	var harness: Dictionary = _PhysicsHarness.make(self, manager, [ball_alpha])
+	await get_tree().physics_frame
+	var target: CourtDropTarget = harness["target"]
+	assert_true(target.can_accept("ball_alpha", Vector2.ZERO, 1.0))
+	assert_true(target.can_accept("ball_alpha", Vector2.ZERO, 1.5))
```
J-Melon, line 288: **code-quality**: nitpick. With the multi-line preamble gone, this test reads as a no-op: both asserts pass trivially on an empty court. The original comment was the only signal that this is intentionally a smoke-test for the scale-API path, paired with `test_expansion_ring_scale_genuinely_widens_the_probe` in `test_ball_drag_controller_sh287.gd` which carries the distinguishing assertion. Without that pairing context, a future reader is likely to either delete this as redundant or tighten it incorrectly. Suggest a one-liner like `# Smoke-test that the 1.5x scale path runs without crashing; distinguishing assertion lives in test_ball_drag_controller_sh287.gd.`

PR #535, opened 2026-04-28
SH-297 grab feel: ease to cursor, hit-box, cursor states

J-Melon:
Implements the grab side of the equip-loop drag, in parallel with SH-287 (#533) which owns the release path.

Held body lifts onto the cursor via a short cubic ease (~80 ms) instead of teleport-snapping, so position, scale, and modulation read as one continuous motion. Live balls grow a child `PressArea` Area2D wider than their physics shape so a press near a moving ball lands without pixel precision. A four-state cursor state machine (`default` / `dragging` / `can_drop` / `forbidden`) is exposed off the drag controller; a Node2D `CursorOverlay` placeholder visual sits on the same seam where SH-298's cursor textures will drop in.

The cursor-state derivation is currently routed through the existing rack/court predicates; on merge of #533 the seam re-points at SH-287's `_find_accepting_target`. The release-path code remains untouched so the conflict surface against #533 is small and bounded to the grab/cursor regions.

### DIFF
--- a/scenes/court.tscn
+++ b/scenes/court.tscn
@@ -22,6 +22,7 @@
 [ext_resource type="Script" uid="uid://dkqyrxx74rbvd" path="res://scripts/items/ball_reconciler.gd" id="21_reconciler"]
 [ext_resource type="Script" uid="uid://ddo3h77danynw" path="res://scripts/items/ball_drag_controller.gd" id="22_drag"]
 [ext_resource type="Script" uid="uid://dwy6ux1ak8r6m" path="res://scripts/court/ball_tracker.gd" id="23_tracker"]
+[ext_resource type="Script" path="res://scripts/hud/cursor_overlay.gd" id="24_cursor"]
 
 [sub_resource type="RectangleShape2D" id="RectangleShape2D_horz"]
 size = Vector2(21.6, 883.2)
@@ -217,7 +218,7 @@ script = ExtResource("21_reconciler")
 script = ExtResource("23_tracker")
 ball_system = NodePath("../BallReconciler")
 
-[node name="BallDragController" type="Node2D" parent="." unique_id=1902001002 node_paths=PackedStringArray("rack", "rack_drop_target", "gear_rack", "gear_rack_drop_target", "reconciler")]
+[node name="BallDragController" type="Node2D" parent="." unique_id=1902001002 node_paths=PackedStringArray("rack", "rack_drop_target", "gear_rack", "gear_rack_drop_target", "reconciler", "cursor_overlay")]
 script = ExtResource("22_drag")
 rack = NodePath("../BallRack")
 rack_drop_target = NodePath("../BallRack/DropTarget")
@@ -226,3 +227,7 @@ gear_rack_drop_target = NodePath("../GearRack/DropTarget")
 court_bounds = Rect2(-420, -360, 840, 720)
 venue_bounds = Rect2(-1200, -720, 2400, 1440)
 reconciler = NodePath("../BallReconciler")
+cursor_overlay = NodePath("CursorOverlay")
+
+[node name="CursorOverlay" type="Node2D" parent="BallDragController" unique_id=1902001004]
+script = ExtResource("24_cursor")
--- a/scenes/court.tscn
+++ b/scenes/court.tscn
@@ -22,7 +22,7 @@
 [ext_resource type="Script" uid="uid://dkqyrxx74rbvd" path="res://scripts/items/ball_reconciler.gd" id="21_reconciler"]
 [ext_resource type="Script" uid="uid://ddo3h77danynw" path="res://scripts/items/ball_drag_controller.gd" id="22_drag"]
 [ext_resource type="Script" uid="uid://dwy6ux1ak8r6m" path="res://scripts/court/ball_tracker.gd" id="23_tracker"]
-[ext_resource type="Script" path="res://scripts/hud/cursor_overlay.gd" id="24_cursor"]
+[ext_resource type="Script" uid="uid://bqbdbfpsxq5qo" path="res://scripts/hud/cursor_overlay.gd" id="24_cursor"]
 
 [sub_resource type="RectangleShape2D" id="RectangleShape2D_horz"]
 size = Vector2(21.6, 883.2)
--- a/scenes/court.tscn
+++ b/scenes/court.tscn
@@ -22,7 +22,7 @@
 [ext_resource type="Script" uid="uid://dkqyrxx74rbvd" path="res://scripts/items/ball_reconciler.gd" id="21_reconciler"]
 [ext_resource type="Script" uid="uid://ddo3h77danynw" path="res://scripts/items/ball_drag_controller.gd" id="22_drag"]
 [ext_resource type="Script" uid="uid://dwy6ux1ak8r6m" path="res://scripts/court/ball_tracker.gd" id="23_tracker"]
-[ext_resource type="Script" uid="uid://bqbdbfpsxq5qo" path="res://scripts/hud/cursor_overlay.gd" id="24_cursor"]
+[ext_resource type="PackedScene" uid="uid://dq3kw7n8sh297" path="res://scenes/hud/cursor_overlay.tscn" id="24_cursor"]
 
 [sub_resource type="RectangleShape2D" id="RectangleShape2D_horz"]
 size = Vector2(21.6, 883.2)
@@ -229,5 +229,4 @@ venue_bounds = Rect2(-1200, -720, 2400, 1440)
 reconciler = NodePath("../BallReconciler")
 cursor_overlay = NodePath("CursorOverlay")
 
-[node name="CursorOverlay" type="Node2D" parent="BallDragController" unique_id=1902001004]
-script = ExtResource("24_cursor")
+[node name="CursorOverlay" parent="BallDragController" instance=ExtResource("24_cursor")]
--- a/scripts/entities/ball.gd
+++ b/scripts/entities/ball.gd
@@ -11,7 +11,7 @@ const SPEED_EMIT_THRESHOLD := 10.0
 ## Item key this ball represents; the system reads this on adoption to find the matching ItemDefinition.
 @export var item_key: String = ""
 ## Press hit-box radius multiplier on the authored collider; tunable per-instance for forgiving grabs.
-@export var press_hitbox_inflation: float = 1.6
+@export_range(1.0, 4.0, 0.1) var press_hitbox_inflation: float = 1.6
 
 var speed: float = 0.0
 var min_speed: float
@@ -124,29 +124,30 @@ func _setup_effect_processor() -> void:
 	add_child(effect_processor)
 
 
-func _setup_press_area() -> void:
-	if _press_area != null and is_instance_valid(_press_area):
+func _wire_press_area() -> void:
+	var area: Area2D = get_node_or_null("PressArea") as Area2D
+	if area == null:
 		return
-	var press_radius: float = _authored_collision_radius() * press_hitbox_inflation
-	if press_radius <= 0.0:
-		return
-	var area: Area2D = Area2D.new()
-	area.name = "PressArea"
-	area.input_pickable = true
-	area.monitoring = false
-	area.monitorable = false
-	var shape_node: CollisionShape2D = CollisionShape2D.new()
-	var circle: CircleShape2D = CircleShape2D.new()
-	circle.radius = press_radius
-	shape_node.shape = circle
-	area.add_child(shape_node)
-	add_child(area)
+	_press_area = area
+	var press_shape: CollisionShape2D = null
+	for child in area.get_children():
+		if child is CollisionShape2D:
+			press_shape = child
+			break
+	if press_shape != null:
+		var circle: CircleShape2D = press_shape.shape as CircleShape2D
+		if circle != null:
+			var authored_radius: float = _baseline_collision_radius()
+			if authored_radius > 0.0:
+				# Duplicate so per-instance inflation does not mutate the shared sub-resource.
+				var local_circle: CircleShape2D = circle.duplicate() as CircleShape2D
+				local_circle.radius = authored_radius * press_hitbox_inflation
+				press_shape.shape = local_circle
 	if not area.input_event.is_connected(_on_input_event):
 		area.input_event.connect(_on_input_event)
-	_press_area = area
 
 
-func _authored_collision_radius() -> float:
+func _baseline_collision_radius() -> float:
 	for child in get_children():
 		if child is CollisionShape2D:
 			var shape_node: CollisionShape2D = child
@@ -155,7 +156,6 @@ func _authored_collision_radius() -> float:
 				continue
 			var axis_scale: float = maxf(absf(shape_node.scale.x), absf(shape_node.scale.y))
 			return circle.radius * maxf(axis_scale, 0.001)
-	# No authored CircleShape2D; press area skips construction on radius <= 0.
 	return 0.0
 
 
@@ -175,7 +175,7 @@ func _ball_setup() -> void:
 	input_pickable = false
 	if input_event.is_connected(_on_input_event):
 		input_event.disconnect(_on_input_event)
-	_setup_press_area()
+	_wire_press_area()
 
 
 func _on_input_event(_viewport: Node, event: InputEvent, _shape_idx: int) -> void:
--- a/scripts/entities/ball.gd
+++ b/scripts/entities/ball.gd
@@ -12,6 +12,8 @@ const SPEED_EMIT_THRESHOLD := 10.0
 @export var item_key: String = ""
 ## Press hit-box radius multiplier on the authored collider; tunable per-instance for forgiving grabs.
 @export_range(1.0, 4.0, 0.1) var press_hitbox_inflation: float = 1.6
+## Authored Area2D that routes pointer presses; wired from the scene so the press hit-box stays scene-based.
+@export var press_area: Area2D
 
 var speed: float = 0.0
 var min_speed: float
@@ -25,7 +27,6 @@ var _was_at_max_speed := false
 var _last_emitted_speed: float = 0.0
 var _last_emitted_min: float = 0.0
 var _last_emitted_max: float = 0.0
-var _press_area: Area2D
 
 
 func _ready() -> void:
@@ -125,12 +126,10 @@ func _setup_effect_processor() -> void:
 
 
 func _wire_press_area() -> void:
-	var area: Area2D = get_node_or_null("PressArea") as Area2D
-	if area == null:
+	if press_area == null:
 		return
-	_press_area = area
 	var press_shape: CollisionShape2D = null
-	for child in area.get_children():
+	for child in press_area.get_children():
 		if child is CollisionShape2D:
 			press_shape = child
 			break
@@ -143,8 +142,8 @@ func _wire_press_area() -> void:
 				var local_circle: CircleShape2D = circle.duplicate() as CircleShape2D
 				local_circle.radius = authored_radius * press_hitbox_inflation
 				press_shape.shape = local_circle
-	if not area.input_event.is_connected(_on_input_event):
-		area.input_event.connect(_on_input_event)
+	if not press_area.input_event.is_connected(_on_input_event):
+		press_area.input_event.connect(_on_input_event)
 
 
 func _baseline_collision_radius() -> float:
--- a/scripts/entities/ball.gd
+++ b/scripts/entities/ball.gd
@@ -7,11 +7,11 @@ signal speed_changed(speed: float, min_speed: float, max_speed: float)
 signal pressed(ball: Ball)
 
 const SPEED_EMIT_THRESHOLD := 10.0
-## Tuning surface for the press hit-box; do not couple to physics-collider scale.
-const PRESS_HITBOX_INFLATION: float = 1.6
 
 ## Item key this ball represents; the system reads this on adoption to find the matching ItemDefinition.
 @export var item_key: String = ""
+## Press hit-box radius multiplier on the authored collider; tunable per-instance for forgiving grabs.
+@export var press_hitbox_inflation: float = 1.6
 
 var speed: float = 0.0
 var min_speed: float
@@ -127,7 +127,7 @@ func _setup_effect_processor() -> void:
 func _setup_press_area() -> void:
 	if _press_area != null and is_instance_valid(_press_area):
 		return
-	var press_radius: float = _authored_collision_radius() * PRESS_HITBOX_INFLATION
+	var press_radius: float = _authored_collision_radius() * press_hitbox_inflation
 	if press_radius <= 0.0:
 		return
 	var area: Area2D = Area2D.new()
--- a/scripts/entities/ball.gd
+++ b/scripts/entities/ball.gd
@@ -8,6 +8,11 @@ signal speed_changed(speed: float, min_speed: float, max_speed: float)
 signal pressed(ball: Ball)
 
 const SPEED_EMIT_THRESHOLD := 10.0
+## SH-297: factor by which the press hit-box inflates the ball's authored physics radius.
+## A live ball is small and moving, so a pixel-precise press misses more often than not.
+## Inflating the press surface (without touching the physics shape) lets a press near a
+## moving ball land cleanly. Tune surface; do not couple to physics-collider scale.
+const PRESS_HITBOX_INFLATION: float = 1.6
 
 ## Item key this ball represents; the system reads this on adoption to find the matching ItemDefinition.
 @export var item_key: String = ""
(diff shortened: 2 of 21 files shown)


scripts/items/ball_drag_controller.gd
```diff
@@ -25,6 +40,10 @@ const COURT_BLOCKED_PROBE_RADIUS_PX: float = 14.0
 @export var court_bounds: Rect2 = Rect2()
 @export var venue_bounds: Rect2 = Rect2()
 @export var reconciler: BallReconciler
+## SH-297: optional cursor overlay node. When unset the controller still derives the
+## cursor state, emits `cursor_state_changed`, and exposes `get_cursor_state()`; only the
+## visual driving step is skipped. Tests don't need to wire this.
+@export var cursor_overlay: Node2D
```
J-Melon: **gdscript-conventions** suggestion (non-blocking): the rest of this controller types its node exports against their `class_name` (`rack: RackDisplay`, `reconciler: BallReconciler`), which is what lets the call sites stay statically typed. Here `cursor_overlay: Node2D` forces the duck-typed `cursor_overlay.call("set_state", ...)` in `_set_cursor_state`. Since `CursorOverlay` has a `class_name`, typing this `@export var cursor_overlay: CursorOverlay` and calling `cursor_overlay.set_state(state, world_position)` would match project pattern and drop the `has_method` guard.

scenes/court.tscn
```diff
@@ -22,6 +22,7 @@
 [ext_resource type="Script" uid="uid://dkqyrxx74rbvd" path="res://scripts/items/ball_reconciler.gd" id="21_reconciler"]
 [ext_resource type="Script" uid="uid://ddo3h77danynw" path="res://scripts/items/ball_drag_controller.gd" id="22_drag"]
 [ext_resource type="Script" uid="uid://dwy6ux1ak8r6m" path="res://scripts/court/ball_tracker.gd" id="23_tracker"]
+[ext_resource type="Script" path="res://scripts/hud/cursor_overlay.gd" id="24_cursor"]
```
J-Melon: **godot-scene**

**suggestion (non-blocking):** This new `[ext_resource]` is the only Script reference in `court.tscn` without a `uid=` attribute. Every sibling Script line carries one (e.g. `id="22_drag"` has `uid="uid://ddo3h77danynw"`), and `scripts/hud/cursor_overlay.gd.uid` already pins `uid://bqbdbfpsxq5qo`. Add it here so the reference is uid-anchored from the first commit and matches the rest of the file:

```
[ext_resource type="Script" uid="uid://bqbdbfpsxq5qo" path="res://scripts/hud/cursor_overlay.gd" id="24_cursor"]
```

Godot will fill this in on the next editor save, but the diff as committed is inconsistent.

scripts/entities/ball.gd
```diff
@@ -122,6 +131,48 @@ func _setup_effect_processor() -> void:
 	add_child(effect_processor)
 
 
+## SH-297: builds the inflated press hit-box as an Area2D child. Reads the ball's authored
+## CollisionShape2D radius so the press surface scales with the ball's physics size, then
+## inflates by `PRESS_HITBOX_INFLATION`. The Area2D doesn't collide with anything; its
+## only job is to surface left-mouse-button presses through `_on_input_event`.
```
J-Melon: **code-quality** issue: this docstring and the SH-297 ones across ball.gd and ball_drag_controller.gd run 3-4 lines each, against the one-line WHY-only policy in ai/skills/minions/code-comments.md; collapse to one line each or move detail to the challenge body.

scripts/items/ball_drag_controller.gd
```diff
@@ -382,6 +444,90 @@ func _get_item_role(item_key: String) -> StringName:
...
+## 1 once the body has fully eased onto the cursor. Linear here; cubic ease-out applied
+## in `_apply_grab_ease` keeps the math single-sourced.
+func _grab_ease_progress() -> float:
+	if GRAB_EASE_DURATION_S <= 0.0:
+		return 1.0
+	return clampf(_grab_ease_elapsed / GRAB_EASE_DURATION_S, 0.0, 1.0)
+
+
+## SH-297: drives held-token position, scale, and modulation across the grab-ease window.
+## `progress` is the linear t; we apply a cubic ease-out so the body decelerates onto the
```
J-Melon: **signals-lifecycle** suggestion: `cursor_state_changed(state: int)` is declared and emitted but no production code connects to it. The overlay is driven via direct `cursor_overlay.call("set_state", ...)` two lines above the `emit`. That works, but it leaves the signal as a dangling public surface (only tests via `watch_signals` consume it). Two clean options: (a) drop the direct call and let the overlay `connect` to `cursor_state_changed` in its `_ready` (more idiomatic, decouples controller from overlay shape), or (b) drop the signal and keep the direct invocation if you genuinely don't want external listeners. As-is, `signal_map(find="orphans")` will start flagging this once the lazy connection lands.

scripts/entities/ball.gd
```diff
@@ -122,6 +131,48 @@ func _setup_effect_processor() -> void:
...
+	area.add_child(shape_node)
+	add_child(area)
+	if not area.input_event.is_connected(_on_input_event):
+		area.input_event.connect(_on_input_event)
+	_press_area = area
+
+
+## SH-297: resolves the ball's authored physics radius from its CollisionShape2D so the
+## press hit-box is a multiple of it. Falls back to the SPEED_EMIT_THRESHOLD-era default
+## of 10 px if the body has no resolvable circle (defensive only; the authored ball.tscn
```
J-Melon: **code-quality** nitpick: SPEED_EMIT_THRESHOLD is a speed delta, not a radius; the 10.0 fallback shares the magic number by coincidence, so this comment implies a coupling that does not exist. Drop the reference or name a separate FALLBACK_PRESS_RADIUS.

scripts/items/ball_drag_controller.gd
J-Melon: **abe**

resolved: tightened `cursor_overlay` to `CursorOverlay` and dropped the `has_method` duck-typing in `_set_cursor_state`.

scenes/court.tscn
J-Melon: **abe**

resolved: added `uid="uid://bqbdbfpsxq5qo"` to the `cursor_overlay.gd` ext_resource line.

scripts/entities/ball.gd
J-Melon: **abe**

resolved: collapsed the SH-297 multi-line docstrings across `ball.gd`, `ball_drag_controller.gd`, `cursor_overlay.gd`, and `cursor_state.gd` to one-line WHY-only or dropped where redundant per the comment policy.

scripts/items/ball_drag_controller.gd
J-Melon: **abe**

resolved: connected the overlay to `cursor_state_changed` in the controllers `_ready` and dropped the direct `cursor_overlay.call("set_state", ...)`. Signal payload extended to `(state, world_position)` so the overlay set_state slot is the only listener path.

scripts/entities/ball.gd
J-Melon: **abe**

resolved: dropped the misleading SPEED_EMIT_THRESHOLD reference and changed the fallback to return 0.0 with a one-line WHY (the press area skips construction on radius <= 0).

scripts/hud/cursor_overlay.gd
```diff
@@ -0,0 +1,57 @@
...
+	z_index = 4096
+	top_level = true
+	visible = false
+
+
+func set_state(state: int, world_position: Vector2) -> void:
+	global_position = world_position
+	if state == _state:
+		queue_redraw()
+		return
```
J-Melon, line 28: **signals-lifecycle** (issue, suggestion)

`cursor_state_changed` now fires every `_process` tick (per the rewire), so this branch hits every frame whenever a drag is active. `queue_redraw()` here is wasted: `_draw()` only draws at `Vector2.ZERO` and position is carried by the canvas-item transform, so a same-state, position-only update needs no redraw — Godot retransforms the cached item for free.

Suggest dropping the redraw on the unchanged-state path:

```gdscript
func set_state(state: int, world_position: Vector2) -> void:
	global_position = world_position
	if state == _state:
		return
	_state = state
	visible = state != CursorStateScript.State.DEFAULT
	queue_redraw()
```

scripts/items/ball_drag_controller.gd
```diff
@@ -69,18 +76,29 @@ func _ready() -> void:
 		if not reconciler.ball_spawned.is_connected(_on_reconciler_ball_spawned):
 			reconciler.ball_spawned.connect(_on_reconciler_ball_spawned)
 
+	if cursor_overlay != null:
+		if not cursor_state_changed.is_connected(cursor_overlay.set_state):
+			cursor_state_changed.connect(cursor_overlay.set_state)
```
J-Melon, line 98: **signals-lifecycle** (nit, suggestion)

The `_ready` connection has no matching disconnect, and `cursor_overlay` is an `@export NodePath` so a future scene could legitimately point at an overlay outside this subtree. In the current `court.tscn` the overlay is a child of the controller, so they free together and this is safe — but as soon as someone wires an external overlay, freeing the overlay first would leave the controller emitting to a freed Callable.

Either (a) add a `tree_exiting` disconnect symmetric with the `_ready` connect, or (b) document the same-subtree-only assumption above the `@export` so a future wiring change does not silently introduce a freed-instance emit.

tests/unit/items/test_grab_feel.gd
```diff
@@ -0,0 +1,209 @@
...
+	assert_lt(token.global_position.x, target.x)
+	assert_gt(token.global_position.y, origin.y)
+	assert_lt(token.global_position.y, target.y)
+	# Modulation alpha is strictly inside (0, 1).
+	assert_gt(token.modulate.a, 0.0)
+	assert_lt(token.modulate.a, 1.0)
+	# Scale is strictly between START_SCALE * target and target.
+	var start_scale: Vector2 = BallDragControllerScript.GRAB_EASE_START_SCALE * target_scale
+	assert_gt(token.scale.x, start_scale.x)
+	assert_lt(token.scale.x, target_scale.x)
```
J-Melon: **test-coverage** suggestion (non-blocking): the docstring promises this test fails for a linear-or-snap regression, but the strict-between bounds only catch snaps. At t=0.5, linear gives `eased=0.5` (lerp midpoint) and cubic ease-out gives `eased=0.875`; both lie strictly between origin and target, so a regression that swaps the cubic for `eased = progress` would still pass.

To pin the ease shape, assert position is past the linear midpoint, e.g.:
```gdscript
var linear_midpoint: Vector2 = origin.lerp(target, 0.5)
assert_gt(token.global_position.x, linear_midpoint.x, "cubic ease-out at t=0.5 is past the linear midpoint")
assert_gt(token.global_position.y, linear_midpoint.y)
```
Same trick on alpha (linear midpoint 0.5 vs cubic 0.875) and scale.

Also minor: `target_scale := Vector2(1.5, 1.5)` is the `ItemDefinition.token_scale` default; reading it off `ball_alpha.token_scale` (or via `_drag._grab_target_scale` post-grab) would survive a default change.

scripts/entities/ball.gd
```diff
@@ -4,10 +4,11 @@ extends RigidBody2D
 signal missed
 signal at_max_speed_changed(is_at_max: bool)
 signal speed_changed(speed: float, min_speed: float, max_speed: float)
-## Mid-rally grab entry: emitted when the player presses the live ball.
 signal pressed(ball: Ball)
 
 const SPEED_EMIT_THRESHOLD := 10.0
+## Tuning surface for the press hit-box; do not couple to physics-collider scale.
+const PRESS_HITBOX_INFLATION: float = 1.6
```
J-Melon: Tuning should be config

scripts/hud/cursor_overlay.gd
```diff
@@ -0,0 +1,57 @@
...
+
+const CursorStateScript: GDScript = preload("res://scripts/items/cursor_state.gd")
+
+const CURSOR_RADIUS_PX: float = 18.0
+const RING_WIDTH_PX: float = 3.0
+
+const COLOR_DEFAULT: Color = Color(1.0, 1.0, 1.0, 0.0)
+const COLOR_DRAGGING: Color = Color(0.85, 0.85, 0.85, 0.85)
+const COLOR_CAN_DROP: Color = Color(0.45, 0.95, 0.55, 0.95)
+const COLOR_FORBIDDEN: Color = Color(0.95, 0.35, 0.35, 0.95)
```
J-Melon: config

scripts/hud/cursor_overlay.gd
J-Melon: Good idea

scripts/items/ball_drag_controller.gd
```diff
@@ -2,21 +2,23 @@ class_name BallDragController
...
+const CursorStateScript: GDScript = preload("res://scripts/items/cursor_state.gd")
 
 const CURSOR_SAMPLE_WINDOW: float = 0.08
 const PRESERVED_SPEED_NONE: float = -1.0
 ## Minimum cursor travel before a rack-origin gesture counts as a real drag (SH-252 a).
 const COMMIT_MOVEMENT_THRESHOLD_PX: float = 6.0
 ## SH-287 patch: probe radius for the pre-spawn body projection; slightly larger than authored ball radius.
 const COURT_BLOCKED_PROBE_RADIUS_PX: float = 14.0
+const GRAB_EASE_DURATION_S: float = 0.08
+const GRAB_EASE_START_SCALE: Vector2 = Vector2(0.85, 0.85)
```
J-Melon: Config

tests/unit/items/test_grab_feel.gd
J-Melon: Sounds like impl detail, test behaviour

scripts/entities/ball.gd
J-Melon: **abe**

resolved: promoted `PRESS_HITBOX_INFLATION` to `@export var press_hitbox_inflation` on `Ball`; updated the press hit-box test to read it off the instance.

scripts/hud/cursor_overlay.gd
J-Melon: **abe**

resolved: promoted `CURSOR_RADIUS_PX` (and the ring width + state colours alongside it) to `@export var` on `CursorOverlay` so the placeholder visual is tunable per-scene.

scripts/hud/cursor_overlay.gd
J-Melon: **abe**

resolved: dropped the redundant `queue_redraw()` on the unchanged-state branch in `set_state`, since `cursor_state_changed` now fires every `_process` tick during a drag.

scripts/items/ball_drag_controller.gd
J-Melon: **abe**

resolved: promoted the grab-ease cluster (`grab_ease_duration_s`, `grab_ease_start_scale`, `grab_ease_start_modulate`, `grab_ease_end_modulate`) to `@export var` on `BallDragController`.

tests/unit/items/test_grab_feel.gd
J-Melon: **abe**

resolved: replaced the cubic-midpoint strict-bounds assertion with a behaviour test that drives the ease across the window in fixed-dt ticks and asserts the player-visible outcomes (token lands on the cursor, modulate alpha reaches 1.0, no single tick teleports more than half the trip). The cubic shape is no longer pinned, only the lift's promise.

tests/unit/items/test_grab_feel.gd
```diff
@@ -0,0 +1,214 @@
...
+	_drag.grab_from_rack("ball_alpha", Vector2(0, 0))
+	var token: Node2D = _drag.get_held_token()
+	assert_eq(token.modulate.a, 0.0, "modulation alpha is 0 at lift start; eases up to 1")
+
+
+func test_held_token_settles_on_cursor_without_teleporting() -> void:
+	# Behaviour: the lift moves the held token from press-origin to the cursor over the ease
+	# window. At the end the token sits on the cursor with full alpha, and no single tick
+	# during the window jumps more than half the total origin-to-target distance (rules out
+	# snap and linear-fast regressions; a continuous ease never moves more than ~ half the
```
J-Melon: **code-quality**

**suggestion (non-blocking):** five-line `#` block inside a test function. CLAUDE.md disallows multi-line comments. The test name and `assert_lt(..., "no mid-window teleport")` already carry the intent; the block can shrink to one line or move into the docstring.

scripts/entities/ball.gd
```diff
@@ -4,13 +4,14 @@ extends RigidBody2D
...
 signal speed_changed(speed: float, min_speed: float, max_speed: float)
-## Mid-rally grab entry: emitted when the player presses the live ball.
 signal pressed(ball: Ball)
 
 const SPEED_EMIT_THRESHOLD := 10.0
 
 ## Item key this ball represents; the system reads this on adoption to find the matching ItemDefinition.
 @export var item_key: String = ""
+## Press hit-box radius multiplier on the authored collider; tunable per-instance for forgiving grabs.
+@export var press_hitbox_inflation: float = 1.6
```
J-Melon: **gdscript-conventions**

suggestion (non-blocking): `@export_range(1.0, 4.0, 0.1)` here. Sub-1.0 silently shrinks the press hit-box below the physics collider (regresses grab forgiveness), and negative is meaningless. Mirrors the `@export_range` lens applied on #533.

scripts/hud/cursor_overlay.gd
```diff
@@ -0,0 +1,58 @@
...
+extends Node2D
+
+## Placeholder visual for the grab cursor state machine; replaced by SH-298 textures.
+
+const CursorStateScript: GDScript = preload("res://scripts/items/cursor_state.gd")
+
+## Ring radius for the placeholder cursor overlay; tunable per-scene.
+@export var cursor_radius_px: float = 18.0
+## Ring stroke width for the placeholder cursor overlay.
+@export var ring_width_px: float = 3.0
```
J-Melon: **gdscript-conventions**

suggestion (non-blocking): `@export_range(1.0, 64.0, 1.0)` for `cursor_radius_px` and `@export_range(0.5, 16.0, 0.5)` for `ring_width_px`. Zero or negative on either silently misbehaves (no ring, inverted dot). Same lens as #533.

scripts/items/ball_drag_controller.gd
```diff
@@ -2,21 +2,27 @@ class_name BallDragController
...
+const CursorStateScript: GDScript = preload("res://scripts/items/cursor_state.gd")
 
 const CURSOR_SAMPLE_WINDOW: float = 0.08
 const PRESERVED_SPEED_NONE: float = -1.0
 ## Minimum cursor travel before a rack-origin gesture counts as a real drag (SH-252 a).
 const COMMIT_MOVEMENT_THRESHOLD_PX: float = 6.0
 ## SH-287 patch: probe radius for the pre-spawn body projection; slightly larger than authored ball radius.
 const COURT_BLOCKED_PROBE_RADIUS_PX: float = 14.0
+## Duration of the grab ease-to-cursor lift; the held token tweens for this long after a press.
+@export var grab_ease_duration_s: float = 0.08
```
J-Melon, line 20: **gdscript-conventions**

suggestion (non-blocking): `@export_range(0.0, 1.0, 0.01)`. The `<= 0.0` guard handles the snap case but `minf(_grab_ease_elapsed + delta, grab_ease_duration_s)` with a negative value silently caps elapsed below zero. Same lens as #533.

tests/unit/items/test_grab_feel.gd
```diff
@@ -0,0 +1,214 @@
...
+	_manager.take("ball_alpha")
+	_drag.grab_from_rack("ball_alpha")
+	# Shrink venue bounds to a small rect that excludes the cursor's default global mouse
+	# position; _is_within_venue then reports false and the derivation returns FORBIDDEN.
+	_drag.venue_bounds = Rect2(Vector2(99000, 99000), Vector2(1, 1))
+
+	var state: int = _drag._derive_cursor_state(Vector2.ZERO)
+
+	assert_eq(state, CursorStateScript.State.FORBIDDEN)
+
```
J-Melon, line 172: **test-coverage** nit: the test name says "drives_overlay_through_process" but the body never invokes `_process` — it calls `_update_cursor_state` directly and pre-sets `_held_token.global_position`. The assertions still pin the right thing (derivation + signal payload + ordering), but the name overpromises. Either rename to `..._drives_overlay_via_signal_payload` or actually tick `_process` here. Non-blocking.

scripts/entities/ball.gd
J-Melon: Yeah do this

scripts/entities/ball.gd
```diff
@@ -122,6 +124,41 @@ func _setup_effect_processor() -> void:
...
+	var press_radius: float = _authored_collision_radius() * press_hitbox_inflation
+	if press_radius <= 0.0:
+		return
+	var area: Area2D = Area2D.new()
+	area.name = "PressArea"
+	area.input_pickable = true
+	area.monitoring = false
+	area.monitorable = false
+	var shape_node: CollisionShape2D = CollisionShape2D.new()
+	var circle: CircleShape2D = CircleShape2D.new()
```
J-Melon: Why are these programmed and not scene based?

scripts/hud/cursor_overlay.gd
J-Melon: The colours also

scripts/entities/ball.gd
J-Melon: **abe**

resolved: added `@export_range(1.0, 4.0, 0.1)` to `press_hitbox_inflation` so sub-1.0 values that would shrink the press hit-box below the physics collider are inspector-blocked.

scripts/entities/ball.gd
J-Melon: **abe**

resolved: moved `PressArea` (Area2D + CollisionShape2D + CircleShape2D) into `scenes/ball.tscn`. Runtime `_wire_press_area` now just resolves the scene-authored node and applies `press_hitbox_inflation` to a duplicated CircleShape2D so per-instance tuning still lands. Connection to `_on_input_event` runs once during `_ball_setup`.

scripts/hud/cursor_overlay.gd
J-Melon: **abe**

resolved: packaged `CursorOverlay` as `scenes/hud/cursor_overlay.tscn`; replaced the inline-script `Node2D` under `BallDragController` in `scenes/court.tscn` with a PackedScene instance. State colours and ring metrics remain `@export` so they author per-scene-instance.

tests/unit/items/test_grab_feel.gd
J-Melon: **abe**

resolved: collapsed the five-line block to a single-line WHY summary; the test name and `assert_lt(..., "no mid-window teleport")` carry the rest.

tests/unit/items/test_grab_feel.gd
J-Melon: **abe**

resolved: renamed to `test_cursor_state_changed_signal_drives_overlay_via_signal_payload` so the name matches what the body actually exercises (signal payload sequence, not `_process` cadence).

scripts/entities/ball.gd
```diff
@@ -122,6 +124,41 @@ func _setup_effect_processor() -> void:
 	add_child(effect_processor)
 
 
+func _wire_press_area() -> void:
+	var area: Area2D = get_node_or_null("PressArea") as Area2D
```
J-Melon: **Lewis** suggestion (non-blocking): the rest of the project wires authored child Area2Ds via `@export var foo: Area2D` (see `shop.gd:11`, `shop_item.gd:15`, `ball_drag_controller.gd:22-24`) so a rename of the child node fails loudly at edit time. The string-based `get_node_or_null("PressArea")` here silently no-ops on rename — exactly the failure mode `feedback_export_over_onready` exists to avoid. Worth swapping to `@export var press_area: Area2D` and wiring it in `ball.tscn` for parity with the rest of the codebase.

scripts/entities/ball.gd
J-Melon: **abe**

resolved: replaced `_wire_press_area`'s `get_node_or_null("PressArea")` lookup with `@export var press_area: Area2D` on Ball, wired from `scenes/ball.tscn` via `node_paths=PackedStringArray("press_area")` + `press_area = NodePath("PressArea")`. Per-instance `CircleShape2D` duplication still happens in code so press_hitbox_inflation tuning stays per-Ball. (5411852)

scripts/hud/cursor_overlay.gd
J-Melon: **abe**

resolved: extracted the colour cluster + ring metrics into `CursorOverlayPalette extends Resource` (`scripts/hud/cursor_overlay_palette.gd`), exported as a single `palette` field on `CursorOverlay`. Defaults preload `res://resources/hud/cursor_overlay_palette.tres`, matching the `Shop` / `BaseStatsConfig` idiom; `cursor_overlay.tscn` wires the `.tres` via ext_resource. (5411852)

PR #551, opened 2026-04-29
SH-333 chore: retire specialist-only dispatch hook

J-Melon:
The `specialist-only-dispatch` PreToolUse hook nagged on legitimate general-purpose dispatches when no specialist fit the topic. Post-SH-329 the dispatch rule is "general-purpose when no specialist fits the topic AND the unit of work needs Bash to ship" — encoded as canon in `ai/skills/gru/dispatch.md`, no longer enforced as a hook.


J-Melon:
**phil** (ci-and-workflows): clean. Hook file deleted, `.claude/settings.json` JSON valid (`{"hooks": {}}`), no in-tree references to `specialist-only-dispatch.sh` outside sibling agent worktrees and the session-handoff scratchpad noting the deletion. Harness coherent.

PR #552, opened 2026-04-29
SH-330 docs: story-led narrative outline, concrete discipline

J-Melon:
Reshapes the two narrative docs so a reader finds the story near the top and gets concrete patterns rather than abstract principles.

`outline.md` opens with a four-paragraph story summary covering protagonist, the Construction climb, the Reconstruction dread, and the dial at the cliff. The apex-return mechanics relocate to `21-ball-dynamics.md`; the per-rally reaction beat plus Aubrey scoping note relocate to the art bible's animation section. The narrative doc keeps short pointers to both.

`discipline.md` now anchors each principle on a worked Volley scene, with an anti-example wherever it sharpens the rule. Layer-first scenes, plain Reality dialogue, sideways memory, slip-and-resettle, present-tense surfacings, line-level cracks, oblique theme: every section pairs principle with a copyable pattern.

Closes SH-330.


J-Melon:
**otto** (docs-and-writing, re-review)

Revision (commit `f2276df`) clears cleanly.

- discipline.md trim reads as a quick reference. Each one-liner carries enough meaning to stand without an example; none collapsed into gnomic shorthand. The opening note that examples live in scene drafts sets the right expectation for a doc that will shift with the story.
- outline.md physical-canon excision is complete (`mid-thirties`, `mid-30`, `agender`, `less athletic` absent from both files). Story opener now lands on "Years ago a close friend died cliff-jumping", which is a stronger first beat than the demographic lead it replaced. Cast section's role-shaped framing flows and the relative link to `../characters/protagonist.md` resolves.
- No em dashes, no banned vocabulary, no AI-tells in the new diff.

Approving via comment + label per SH-229 self-review limitation.

PR #597, opened 2026-05-09
SH-309 docs: wall-less court control spec

J-Melon:
Tech-design spike for SH-309. Adds `designs/01-prototype/08-court-control.md`, a focused spec for the wall-less court that implementation will work against.

The TLDR splits into three beats: today the rally tops out with a pong-bounce off a screen-edge ceiling and balls straying sideways bounce off invisible walls and stay in play; the player-felt change at each affected edge (held arc on top, ball-on-the-floor on the sides); and the implementation per surface (top collider gone with gravity above a per-venue friendship-bound height; side walls replaced with Area2D miss bands feeding the existing miss path).

Implementation is split out: SH-366 owns the apex return, SH-367 owns the side miss zones. Both are blocked by this spec.

The two old docs (`08-court-bounds.md` and `21-ball-dynamics.md`) carry material the new doc does not cover (ball state machine, container swap, cue layering, spirit-of-the-volley framing, resting-balls flow). They stay in place for now; we'll extract the kept material into focused successor docs as part of the implementation tickets.

### DIFF
--- a/designs/01-prototype/08-court-bounds.md
+++ b/designs/01-prototype/08-court-bounds.md
@@ -1,155 +1,39 @@
 # Court Bounds and Miss
 
-The court is not a closed box. It sits inside the venue, bounded on three sides and open on one; missed balls leave the court and rest on the venue floor until the player puts them back.
+Court geometry, the apex return, and the miss event live in [`08-court-control.md`](08-court-control.md). This doc covers cues, the spirit-of-the-volley framing, the rest-balls flow, and the future helper.
 
-**Dependencies:** Venue (`08-venue.md`), Balls (`08-balls.md`), Kit (`08-kit.md`), Items (`08-items.md`), Roles (`08-roles.md`), Fixtures (`08-fixtures.md`).
-
----
-
-## Bounds
-
-| Edge | Bound | Behaviour |
-|---|---|---|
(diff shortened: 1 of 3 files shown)


J-Melon (commented):
**Held Sky:** STYLE.md clean. (Reviewed earlier draft; verdict still holds for current spec.)

J-Melon (commented):
**Echo Sweep:** No cross-doc duplication; trim-verify deferred per PR description, posture fine. (Re-checked after the trim landed: 08-court-bounds and 21-ball-dynamics now forward to 08-court-control.)

J-Melon (commented):
**Bound Stress:** needs-rework on the original draft. Caught: (1) "active-movement state" is a doc concept, not code; (2) back-miss does not flip gravity today, so the spec was mirroring a no-op; (3) gravity-toggle conflicted with 21-ball-dynamics velocity-flip canon; (4) "left wall" is actually `Back Wall`, hosting the back-miss zone; (5) side bands need to be `MissZone` instances in the `miss_zones` group, not raw `Area2D`; (6) "rolls to rest" needs `linear_damp` or floor friction; (7) bound-height should live on a Resource per data-driven-tuning. Spec rewritten in 82534b4: collapsed back/side into one miss type, dropped impl detail, surfaced rest-roll energy + bound-data-shape + drag-handoff as open questions. Re-review needed against the rewrite.

PR #600, opened 2026-05-10
SH-366 feat: apex return via gravity-toggle

J-Melon:
Replaces the top-wall pong-bounce with a per-Court friendship-bound height. Balls above the bound get a configurable `gravity_scale`; balls at or below get zero. The ball's per-frame speed-lock turns the gravity acceleration into directional curvature without adding kinetic energy, so the rally arcs back into play instead of ricocheting off an invisible ceiling.

`Court` exports `apex_bound_y` and `apex_gravity_scale` and pushes them through `BallTracker` to every attached ball, so multi-ball is covered by construction. The Top Wall collider is gone from `court.tscn`.

Live runtime verification (rally continues across the arc, volley counter and paddle hits remain valid mid-arc) needs a Tier 2 pass; that hand-off goes to the runtime-verifier.

### DIFF
--- a/scripts/court/ball_tracker.gd
+++ b/scripts/court/ball_tracker.gd
@@ -17,19 +17,17 @@ var _current_ball: Ball
 var _partner_paddle: Node2D
 var _player_paddle: Node2D
 var _miss_zones: Array[MissZone] = []
-var _apex_bound_y: float = -351.6
-var _apex_gravity_scale: float = 1.0
-var _apex_configured: bool = false
+var _apex_bound_y: float = ApexDefaults.BOUND_Y
+var _apex_gravity_scale: float = ApexDefaults.GRAVITY_SCALE
 
 
 func configure(player_paddle: Node2D) -> void:
 	_player_paddle = player_paddle
 
 
-func set_apex(bound_y: float, gravity_scale: float) -> void:
+func configure_apex(bound_y: float, gravity_scale_value: float) -> void:
 	_apex_bound_y = bound_y
-	_apex_gravity_scale = gravity_scale
-	_apex_configured = true
+	_apex_gravity_scale = gravity_scale_value
 	for tracked in _balls:
 		if is_instance_valid(tracked):
 			_apply_apex(tracked)
@@ -156,7 +154,5 @@ func _on_ball_at_max_speed_changed(is_at_max: bool) -> void:
 
 
 func _apply_apex(target: Ball) -> void:
-	if not _apex_configured:
-		return
 	target.apex_bound_y = _apex_bound_y
 	target.apex_gravity_scale = _apex_gravity_scale
--- a/scripts/court/ball_tracker.gd
+++ b/scripts/court/ball_tracker.gd
@@ -17,10 +17,22 @@ var _current_ball: Ball
 var _partner_paddle: Node2D
 var _player_paddle: Node2D
 var _miss_zones: Array[MissZone] = []
+var _apex_bound_y: float = -351.6
+var _apex_gravity_scale: float = 1.0
+var _apex_configured: bool = false
 
 
 func configure(player_paddle: Node2D) -> void:
 	_player_paddle = player_paddle
+
+
+func set_apex(bound_y: float, gravity_scale: float) -> void:
+	_apex_bound_y = bound_y
+	_apex_gravity_scale = gravity_scale
+	_apex_configured = true
+	for tracked in _balls:
+		if is_instance_valid(tracked):
+			_apply_apex(tracked)
 
 
 func _ready() -> void:
@@ -41,6 +53,7 @@ func attach(new_ball: Ball) -> void:
 	if new_ball == null or _balls.has(new_ball):
 		return
 	_balls.append(new_ball)
+	_apply_apex(new_ball)
 	if _current_ball == null:
 		_set_current(new_ball)
 	if not new_ball.missed.is_connected(_on_ball_missed):
@@ -140,3 +153,10 @@ func _on_ball_missed() -> void:
 
 func _on_ball_at_max_speed_changed(is_at_max: bool) -> void:
 	ball_at_max_speed_changed.emit(is_at_max)
+
+
+func _apply_apex(target: Ball) -> void:
+	if not _apex_configured:
+		return
+	target.apex_bound_y = _apex_bound_y
+	target.apex_gravity_scale = _apex_gravity_scale
--- a/scripts/entities/ball.gd
+++ b/scripts/entities/ball.gd
@@ -14,6 +14,10 @@ const SPEED_EMIT_THRESHOLD := 10.0
 @export_range(1.0, 4.0, 0.1) var press_hitbox_inflation: float = 2.4
 ## Authored Area2D that routes pointer presses; wired from the scene so the press hit-box stays scene-based.
 @export var press_area: Area2D
+## Y-coordinate of the friendship-bound; gravity engages while the ball is above (smaller y) this height. Court overrides per attach.
+@export var apex_bound_y: float = -351.6
+## Gravity scale applied while above the friendship-bound; off-bound the ball travels straight (gravity_scale = 0).
+@export_range(0.0, 8.0, 0.05) var apex_gravity_scale: float = 1.0
 
 var speed: float = 0.0
 var min_speed: float
@@ -40,6 +44,7 @@ func _ready() -> void:
 
 
 func _physics_process(delta: float) -> void:
+	_update_apex_gravity()
 	if linear_velocity == Vector2.ZERO:
 		return
 	effect_processor.process_frame(delta)
@@ -47,6 +52,15 @@ func _physics_process(delta: float) -> void:
 	if _should_emit_speed_changed():
 		_emit_speed_changed()
 	linear_velocity = linear_velocity.normalized() * speed
+
+
+# Gravity-toggle apex return: above the bound the ball arcs under gravity; the per-frame speed-lock
+# then re-normalises magnitude, so gravity bends direction without changing kinetic energy.
+func _update_apex_gravity() -> void:
+	if freeze:
+		gravity_scale = 0.0
+		return
+	gravity_scale = apex_gravity_scale if position.y < apex_bound_y else 0.0
 
 
 func _should_emit_speed_changed() -> bool:
--- a/scripts/entities/ball.gd
+++ b/scripts/entities/ball.gd
@@ -15,9 +15,9 @@ const SPEED_EMIT_THRESHOLD := 10.0
 ## Authored Area2D that routes pointer presses; wired from the scene so the press hit-box stays scene-based.
 @export var press_area: Area2D
 ## Y-coordinate of the friendship-bound; gravity engages while the ball is above (smaller y) this height. Court overrides per attach.
-@export var apex_bound_y: float = -351.6
+@export var apex_bound_y: float = ApexDefaults.BOUND_Y
 ## Gravity scale applied while above the friendship-bound; off-bound the ball travels straight (gravity_scale = 0).
-@export_range(0.0, 8.0, 0.05) var apex_gravity_scale: float = 1.0
+@export_range(0.0, 8.0, 0.05) var apex_gravity_scale: float = ApexDefaults.GRAVITY_SCALE
 
 var speed: float = 0.0
 var min_speed: float
@@ -44,7 +44,7 @@ func _ready() -> void:
 
 
 func _physics_process(delta: float) -> void:
-	_update_apex_gravity()
+	_update_apex_gravity(delta)
 	if linear_velocity == Vector2.ZERO:
 		return
 	effect_processor.process_frame(delta)
@@ -54,13 +54,17 @@ func _physics_process(delta: float) -> void:
 	linear_velocity = linear_velocity.normalized() * speed
 
 
-# Gravity-toggle apex return: above the bound the ball arcs under gravity; the per-frame speed-lock
-# then re-normalises magnitude, so gravity bends direction without changing kinetic energy.
-func _update_apex_gravity() -> void:
+# Speed-lock turns gravity into curvature, not energy gain.
+func _update_apex_gravity(delta: float = 0.0) -> void:
 	if freeze:
 		gravity_scale = 0.0
 		return
-	gravity_scale = apex_gravity_scale if position.y < apex_bound_y else 0.0
+	# Predicted next-step y closes the one-tick overshoot: a fast-rising ball that skips the above-bound
+	# sample still triggers gravity. `<=` makes the bound closed so a ball stalled at exactly the line falls.
+	var current_y: float = global_position.y
+	var predicted_y: float = current_y + linear_velocity.y * delta
+	var above_bound: bool = current_y <= apex_bound_y or predicted_y <= apex_bound_y
+	gravity_scale = apex_gravity_scale if above_bound else 0.0
 
 
 func _should_emit_speed_changed() -> bool:
--- a/scenes/court.tscn
+++ b/scenes/court.tscn
@@ -58,22 +58,6 @@ item_key = "base_ball"
 
 [node name="Bounds" type="Node2D" parent="." unique_id=2106668171]
 
-[node name="Top Wall" type="StaticBody2D" parent="Bounds" unique_id=1297110620]
-position = Vector2(441.6, -351.6)
-rotation = 1.5707964
-physics_material_override = ExtResource("1_0wfyh")
-metadata/_edit_group_ = true
-
-[node name="ColorRect" type="ColorRect" parent="Bounds/Top Wall" unique_id=1483069279]
-offset_right = 21.6
-offset_bottom = 883.2
-mouse_filter = 2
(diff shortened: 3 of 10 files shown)


J-Melon (commented):
**tephra** — no new signal wiring or tree-lifecycle surface beyond the existing tracker connects; `set_apex` guards iteration with `is_instance_valid` and the new gravity-toggle is pure per-frame state. Nothing to flag.

scripts/court/ball_tracker.gd
```diff
@@ -17,12 +17,24 @@ var _current_ball: Ball
 var _partner_paddle: Node2D
 var _player_paddle: Node2D
 var _miss_zones: Array[MissZone] = []
+var _apex_bound_y: float = -351.6
```
J-Melon: **gossan** (suggestion, non-blocking): the `-351.6` / `1.0` defaults are now spelled in three places (`Ball`, `BallTracker`, `Court`). If the friendship-bound ever moves, all three drift independently. Consider sourcing the bound from a single owner (e.g. only `Court` exports it; `Ball` and `BallTracker` start with sentinel values and require `set_apex` before behaving). Not a convention violation — flagging for tuning hygiene per the data-driven-tuning memory.

J-Melon (commented):
**gossan** — GDScript conventions pass.

Convention checks all clean:
- `@export` used for new node/tunable references (no `@onready` regressions).
- All new functions have typed args and return types (`set_apex`, `_apply_apex`, `_update_apex_gravity`).
- Signal wiring uses identifier-style `.connect()` callbacks, not raw strings.
- No autoload changes; no `preload()` of autoloads.
- No new `class_name` introduced — existing `Ball` / `BallTracker` are safe to instantiate via `preload(...).new()` in tests.
- Naming is full-word (`apex_bound_y`, `apex_gravity_scale`, `friendship-bound`).

One non-blocking observation below at the default-value triplication.

scripts/entities/ball.gd
```diff
@@ -14,6 +14,10 @@ const SPEED_EMIT_THRESHOLD := 10.0
 @export_range(1.0, 4.0, 0.1) var press_hitbox_inflation: float = 2.4
 ## Authored Area2D that routes pointer presses; wired from the scene so the press hit-box stays scene-based.
 @export var press_area: Area2D
+## Y-coordinate of the friendship-bound; gravity engages while the ball is above (smaller y) this height. Court overrides per attach.
+@export var apex_bound_y: float = -351.6
```
J-Melon: **esker** — *suggestion (non-blocking).* The `-351.6` default is duplicated verbatim in `scripts/core/court.gd` (`apex_bound_y`). If the court's friendship-bound ever moves, the ball's fallback silently desyncs. Two options: drop the ball default to `0.0` (Court always pushes via `BallTracker.set_apex`, so the fallback is only hit in tests / standalone instantiation), or hoist a shared `const APEX_BOUND_Y_DEFAULT` so both sites move together.

scripts/entities/ball.gd
```diff
@@ -49,6 +54,15 @@ func _physics_process(delta: float) -> void:
 	linear_velocity = linear_velocity.normalized() * speed
 
 
+# Gravity-toggle apex return: above the bound the ball arcs under gravity; the per-frame speed-lock
+# then re-normalises magnitude, so gravity bends direction without changing kinetic energy.
```
J-Melon: **esker** — *issue (comment policy).* Two-line non-docstring comment. `CLAUDE.md` keeps inline comments to one line; the explanatory "speed-lock turns gravity into directional curvature" reasoning belongs on the function-level docstring (one line) or in the design doc, not above the body.

scripts/court/ball_tracker.gd
```diff
@@ -17,12 +17,24 @@ var _current_ball: Ball
...
+var _apex_gravity_scale: float = 1.0
+var _apex_configured: bool = false
 
 
 func configure(player_paddle: Node2D) -> void:
 	_player_paddle = player_paddle
```
J-Melon, line 26: **esker** — *nitpick.* `set_apex` reads as a single-value setter but pushes two values (bound + gravity scale). `configure_apex` mirrors the existing `configure(player_paddle)` pattern in this file and reads honestly at call sites. The `gravity_scale` parameter also shadows RigidBody2D's built-in property name; `apex_gravity_scale` keeps the matching exported-var spelling.

scripts/court/ball_tracker.gd
```diff
@@ -17,12 +17,24 @@ var _current_ball: Ball
...
 
 
+func set_apex(bound_y: float, gravity_scale: float) -> void:
+	_apex_bound_y = bound_y
+	_apex_gravity_scale = gravity_scale
+	_apex_configured = true
+	for tracked in _balls:
```
J-Melon, line 31: **esker** — *question.* `_apex_configured` only guards the case where `attach` runs before `Court` calls `set_apex`. In `Court._ready` the order is `add_child(ball_tracker)` → `configure` → `set_apex` before any ball attaches, so the flag is dead in production. If it exists purely for tests (`test_attach_without_configure_leaves_ball_defaults`), worth a one-line note on the var; otherwise consider dropping it and letting the exported defaults on `Ball` carry the fallback.

J-Melon (commented):
**esker** — code-quality pass on the apex-return diff. Logic is clean and the BallTracker fan-out is the right seam. Findings are non-blocking polish: a duplicated magic default across two files, a two-line comment that breaks the one-line comment rule, and a couple of naming nits where the function name does not describe what is set.

J-Melon (commented):
**gabbro** (runtime-verifier, Banana Tank)

**Verdict: aborted — runtime not reachable for SH-366 code.**

Cannot run Tier-2 verification on this branch from my isolated worktree. godotiq is reachable (status ok, v0.5.4) but its editor follows the main worktree, which is currently on `chore/reviewer-agents-bash` — not `feature/sh-366-apex-return-arcs-the-ball-back-into-play`. Diff between the two is 568 insertions across `scripts/entities/ball.gd`, `scripts/court/ball_tracker.gd`, `scripts/core/court.gd`, `scenes/court.tscn`, plus the new tests. None of that is loaded in the editor's project.

If I `run(play)` now I would be playtesting the chore branch and reporting false-positive evidence on apex-return ACs. Per `feedback_godotiq_worktree_isolation.md` and `feedback_runtime_minions_must_abort_on_disconnect.md`, the correct move is to abort and surface it.

**To unblock:** Josh checks out `feature/sh-366-apex-return-arcs-the-ball-back-into-play` in the main worktree (or in the editor's project_path), reloads the editor, then redispatches a runtime-verifier. ACs still pending live evidence:
- Ball passes the friendship-bound, arcs, and returns into the rally on its own
- Rally continues across the arc; volley counter and paddle hits remain valid above and below the bound

Static review (hornfels' unit tests + diff) is not a substitute for these — both ACs are player-observable behaviour over time.

J-Melon (commented):
**slate** — test coverage review for SH-366.

**Tests assert the mechanism, not the behaviour.** Every assertion in `tests/unit/ball/test_ball_apex_return.gd` reads `_ball.gravity_scale`. That's the internal lever this implementation pulls; it isn't what the player sees. The AC says the ball "rises, arcs, and returns into the rally on its own" — a behaviour-level test would put the ball above the bound with an upward velocity, step the physics for N ticks, and assert that `linear_velocity.y` flips sign and/or `position.y` ends up back across the bound. A future refactor that produces the same arc via `apply_central_force` or a custom integrator would (correctly) preserve player experience but (incorrectly) break all four tests here.

The `test_ball_tracker_apex.gd` cases are pure property-propagation plumbing — they verify tracker writes flow to `ball.apex_bound_y` / `ball.apex_gravity_scale`, but never that a ball under tracker control actually arcs. If the ball were rewired to read tuning from a Resource or the Court directly, these tests would fail despite identical behaviour.

**AC 2 has no test.** "Rally continues across the arc; volley counter and paddle hits remain valid mid-arc" is acknowledged in the PR body as a runtime-verifier hand-off, which is fine for the live-rally portion, but the unit layer should still cover at least one cheap behavioural check — e.g., a paddle collision while the ball is above the bound still produces the same post-collision speed/direction contract as a collision below the bound, proving gravity-on does not perturb collision response. As written, the suite would pass even if mid-arc paddle hits silently no-op'd.

**Suggested additions** (either in this PR or a follow-up with a `test:` issue linked from SH-366):
- A multi-tick trajectory test: ball above bound with `Vector2(speed, -speed)` velocity, step `_physics_process` ~30 times, assert final `position.y > start.y` (arc returned).
- A speed-conservation test: same setup, assert `linear_velocity.length()` stays within epsilon of `speed` across the arc. This is the load-bearing claim in the PR body ("gravity bends direction without changing kinetic energy") and currently has no assertion anywhere.
- A boundary test at exactly `position.y == apex_bound_y` (the code uses strict `<`, so the equality case is gravity-off; worth pinning).

**Minor:** the `-351.6` default is duplicated across `court.gd`, `ball_tracker.gd`, and `ball.gd`. Not a test issue, but if any one drifts the others will not catch it.

Verdict: ship-with-fixes. The behaviour the ticket describes is observable (trajectory, speed, rally validity) and the tests stop short of observing any of it. Recommend trajectory + speed-conservation tests land before merge; AC 2 unit smoke can follow up.

J-Melon (commented):
**tuff** — scene/resource discipline pass.

**Verdict:** clean.

- `scenes/court.tscn` UID `uid://ctjs62ui6cpso` unchanged; surgical 16-line removal of `Bounds/Top Wall` (StaticBody2D + ColorRect + CollisionShape2D). No delete-and-rebuild.
- ExtResource `1_0wfyh` (bouncy.tres) and SubResource `RectangleShape2D_horz` are still referenced by Bottom/Left/Right walls — no orphan resources, no resource-count explosion.
- No grep hits for "Top Wall" or "Bounds/Top" anywhere on the branch (scripts, scenes, tests). No dangling NodePath.
- `project.godot` untouched: autoload order (`SaveManager`, `ItemManager`, `ProgressionManager`, `ConfigHotReload`) preserved.
- No `@tool` scripts introduced; guard rule N/A.
- Two new `.gd.uid` siblings accompany the new test scripts (`test_ball_apex_return.gd.uid`, `test_ball_tracker_apex.gd.uid`) — correct pairing, no stranded UIDs from renames.
- No `.tres` modifications.

Out of scope here: runtime verification that nothing tries to resolve `Bounds/Top Wall` at runtime — runtime-verifier owns that.

J-Melon (commented):
**chert** — devil's advocate pass on the apex-return approach. Verdict: **ship-with-fixes**. The gravity-toggle reads cleanly and tests confirm the toggle mechanics, but the load-bearing claim that "gravity bends direction without changing kinetic energy, so the rally arcs back into play" is not actually exercised by any test, and several edge cases at the bound have the ball doing things the spec did not promise.

## Failure modes worth naming

**issue (frame-rate dependence):** The toggle samples `position.y` once per physics step. Dwell-time above the bound is integer-frames-of-`_physics_process`, so a ball with high vertical velocity at 30Hz vs 60Hz spends materially different proper-time above the bound and curves differently. The spec promises a felt arc; that feel is now physics-tickrate-coupled. Fix: integrate the bound-crossing analytically across the step (segment-slice gravity by fraction of frame above the line), or document that physics is locked at 60 and accept the coupling.

**issue (overshoot in one tick):** A ball travelling fast upward can cross from `y > apex_bound_y` to well above and back below within one physics step's worth of integrated motion, never sampling a position above the bound. Result: zero gravity applied, ball travels on. This is exactly the case the speed-lock is supposed to handle gracefully, and the implementation as written cannot — gravity_scale is only ever set on what the position-sampler observed.

**issue (hover at exact-equal y):** `position.y < apex_bound_y` is strict-less. A ball whose centre lands at the bound with horizontal velocity gets gravity_scale = 0 and travels parallel to the bound forever. This is reachable on a paddle hit at the bound height. Either widen to `<=` or assert ε-band semantics.

**issue (gravity vs. speed-lock interaction is untested):** None of the new tests step physics across the bound and assert that the velocity vector rotates without growing or shrinking. The unit tests check `gravity_scale` is set; they never construct the integrated trajectory. Add a multi-step test: ball at min_speed travelling +y above bound, run N physics frames, assert `linear_velocity.length()` stays at `speed` (within ε) and `linear_velocity.y` flips sign within a bounded number of frames. Without this the spec claim is unverified.

**issue (small-speed degenerate arc):** With min_speed low and `apex_gravity_scale` at 1.0, gravity adds Δvy per frame that is large relative to `speed`. The post-step normalisation then rotates the velocity vector toward straight-down quickly — a snap, not an arc. Worth a runtime check at the lowest tuned `ball_speed_min` × highest tuned `apex_gravity_scale` combo.

**issue (no ceiling, no clamp):** Removing the top wall means there is nothing stopping a ball from leaving the camera view entirely if `apex_gravity_scale` is undertuned or speed is buffed by an item later. The implementation has no upper-y safety. At minimum, log/assert when a ball climbs more than N units above the bound, so future tuning regressions are loud.

**issue (runtime apex change mid-arc):** `set_apex` propagates new bounds to live balls. If a venue swap happens while a ball is between the old and new bound, gravity flips off mid-arc and the ball escapes. No transition handling. Either disallow apex changes while balls are airborne, or re-evaluate gravity immediately on `set_apex` (call `_update_apex_gravity` on each tracked ball — the existing assignment doesn't do this; gravity_scale only updates next `_physics_process`).

## Spec contradictions

**issue (scope mismatch with spec):** `08-court-control.md` opens with "At either lateral edge, the ball rolls out onto the venue floor" and "The lateral walls go" framed as part of the same rule. This PR removes only the Top Wall and adds no player-side miss zone or rest-roll. A spec reader would expect both halves shipped together. If the lateral half is deliberately punted, say so in the PR description and link the follow-up issue; otherwise this is half a feature.

**issue (paddle hits above bound are spec'd but unverified):** Spec line 13 promises paddle hits register above the bound the same as below. There is no test or runtime evidence that paddle hit + gravity-on + speed-lock interact correctly mid-arc. Hand off to runtime-verifier with a specific scenario, not a generic "runs OK."

## Smaller flags

**nitpick (position vs global_position):** `_update_apex_gravity` uses `position.y`. Court-relative is correct today because Balls are direct children of Court; any future reparent (item effects, slow-motion holders) silently invalidates the bound. Consider `global_position.y` and a court-space-to-world conversion, or assert the parent invariant.

**nitpick (open question 3 unaddressed):** The spec's drag-handoff frame-window question ("miss does not fire while a grab is in flight") is still open. Not blocking apex-return, but the two land in the same change-set logically — flag it explicitly as deferred so it isn't lost.

**nitpick (BallTracker default `_apex_bound_y = -351.6` literal):** The same magic number lives in three files (`ball.gd`, `ball_tracker.gd`, `court.gd`). `_apex_configured` guards propagation so the duplicates aren't load-bearing today, but they will drift. Either centralise on Court (single source) and remove the others, or extract to a constant.

## What would flip this to ship-as-is

- An integration test that steps physics across the bound and asserts speed-magnitude stability + direction reversal.
- A documented answer for the overshoot-in-one-tick case (either tunings make it unreachable, or there is a sub-step correction).
- An explicit note in the PR body that lateral-wall removal is a follow-up, with the linked issue.

The approach is the right one — gravity-toggle reads as a held arc, spec was right to prefer it over a velocity flip. The implementation just needs the load-bearing claim covered by a test and the bound-condition edges nailed down before merge.

---

*chert, devil's-advocate, Tier 0 static pass. Runtime claims belong to runtime-verifier.*

J-Melon (commented):
**marl** — runtime verification (Tier 2, Banana Tank)

Verdict: both ACs pass.

| t | y | vy | gravity_scale | volleys |
|---|---|---|---|---|
| start | +100.01 | +214.66 | 0.0 | 0 |
| +500ms | -110.03 | -213.80 | 0.0 | 0 |
| +1.1s | **-375.41** | -32.63 | **1.0** | 0 |
| +1.5s | -189.31 | +198.94 | 0.0 | 0 |
| +3.5s | +241.26 | -205.70 | 0.0 | 1 |
| +7.5s | -72.48 | +190.50 | 0.0 | 0 (new ball) |

`apex_bound_y = -351.6`, `apex_gravity_scale = 1.0` confirmed configured on Court and propagated to Ball via `BallTracker.set_apex` → `_apply_apex`.

AC1 (apex arc + return): the ball ascends to y=-375 (above bound), gravity_scale flips 0→1, vy decelerates from -213 to -32 to +198 within ~1s, ball re-enters the rally below the bound. The return is gravitational, not scripted; speed-lock holds magnitude (~480 px/s) while gravity bends direction. Confirmed.

AC2 (rally + counter validity): volley counter ticked 0→1 after a paddle hit following an apex arc; PB:20 indicates rally chains have been sustained across multiple arcs in autoplay. New-ball reset at +7.5s is a missed-ball case (paddle missed return), unrelated to the apex code path.

`_editor_state.recent_errors` empty across every snapshot. No engine spam. No regressions surfaced.

Branch verified: feature/sh-366 @ cfd4bc8, project_path /home/josh/gamedev/volley/.

Files of note: scripts/entities/ball.gd:57-65 (`_update_apex_gravity`), scripts/court/ball_tracker.gd:27-34 (`set_apex`), scripts/core/court.gd:20-21,65 (export + wiring).

PR #601, opened 2026-05-10
SH-367 side miss bands replace court walls

J-Melon:
The right court wall becomes a `RightMissBand` Area2D in the `miss_zones` group, paired with a mirrored `LeftMissBand`. `MissZone` gains a `releases_ball` flag that side bands flip on; when a ball crosses a releasing band, `gravity_scale` rises to 1.0 and a small linear damp engages so the ball falls out of the weightless play volume and rolls to rest. The back-miss zone behind the player paddle stays non-releasing, so its speed-reset path is unchanged. The partner miss zone instantiated on partner activation re-parents to the new right band and inherits the release behaviour.

Live verification (rally state, partner re-parenting, roll-out feel) belongs to the runtime-verifier; static GUT coverage on the gravity-on-release path lives in `tests/unit/ball/test_ball.gd`.

### DIFF
--- a/scripts/core/court.gd
+++ b/scripts/core/court.gd
@@ -14,7 +14,8 @@ const MissZoneScene: PackedScene = preload("res://scenes/miss_zone.tscn")
 @export var player_paddle_scene: PackedScene
 @export var player_spawn: Marker2D
 @export var autoplay_controller: AutoplayController
-@export var right_wall: StaticBody2D
+## Right side miss band; partner miss zone re-parents here when a partner activates.
+@export var right_wall: Node2D
 @export var partner_spawn: Marker2D
 @export var timeout_controller: TimeoutController
 
@@ -165,6 +166,7 @@ func _activate_partner() -> void:
 
 	if right_wall != null:
 		_partner_miss_zone = MissZoneScene.instantiate()
+		_partner_miss_zone.releases_ball = true
 		right_wall.add_child(_partner_miss_zone)
 		ball_tracker.register_miss_zone(_partner_miss_zone)
 
--- a/scripts/entities/ball.gd
+++ b/scripts/entities/ball.gd
@@ -7,6 +7,8 @@ signal speed_changed(speed: float, min_speed: float, max_speed: float)
 signal pressed(ball: Ball)
 
 const SPEED_EMIT_THRESHOLD := 10.0
+## Side bands raise gravity so the missed ball falls out of the weightless play volume.
+const _ROLL_OUT_LINEAR_DAMP: float = 1.5
 
 ## Item key this ball represents; the system reads this on adoption to find the matching ItemDefinition.
 @export var item_key: String = ""
@@ -27,6 +29,7 @@ var _was_at_max_speed := false
 var _last_emitted_speed: float = 0.0
 var _last_emitted_min: float = 0.0
 var _last_emitted_max: float = 0.0
+var _registered_miss_zones: Array[MissZone] = []
 
 
 func _ready() -> void:
@@ -77,13 +80,20 @@ func _on_body_entered(body: Node) -> void:
 
 
 func register_miss_zone(zone: MissZone) -> void:
-	if not zone.body_entered.is_connected(_on_miss_zone_body_entered):
-		zone.body_entered.connect(_on_miss_zone_body_entered)
+	if zone == null or _registered_miss_zones.has(zone):
+		return
+	_registered_miss_zones.append(zone)
+	zone.body_entered.connect(_on_miss_zone_body_entered.bind(zone))
 
 
-func _on_miss_zone_body_entered(body: Node) -> void:
-	if body == self:
-		missed.emit()
+func _on_miss_zone_body_entered(body: Node, zone: MissZone) -> void:
+	if body != self:
+		return
+	missed.emit()
+	if zone != null and zone.releases_ball:
+		# Side-miss: leave the weightless play volume so the ball rolls onto the venue floor.
+		gravity_scale = 1.0
+		linear_damp = _ROLL_OUT_LINEAR_DAMP
 
 
 func increase_speed() -> void:
--- a/tests/unit/ball/test_ball.gd
+++ b/tests/unit/ball/test_ball.gd
@@ -175,6 +175,33 @@ func test_register_miss_zone_is_idempotent() -> void:
 	assert_signal_emit_count(_ball, "missed", 1)
 
 
+# --- side-miss release behaviour ---
+func test_releasing_zone_raises_gravity_on_cross() -> void:
+	var zone := MissZone.new()
+	zone.releases_ball = true
+	add_child_autofree(zone)
+	_ball.register_miss_zone(zone)
+	_ball.gravity_scale = 0.0
+
+	zone.body_entered.emit(_ball)
+
+	assert_almost_eq(_ball.gravity_scale, 1.0, 0.001)
+	assert_gt(_ball.linear_damp, 0.0)
+
+
+func test_non_releasing_zone_keeps_weightless_state() -> void:
+	var zone := MissZone.new()
+	add_child_autofree(zone)
+	_ball.register_miss_zone(zone)
+	_ball.gravity_scale = 0.0
+	_ball.linear_damp = 0.0
+
+	zone.body_entered.emit(_ball)
+
+	assert_almost_eq(_ball.gravity_scale, 0.0, 0.001)
+	assert_almost_eq(_ball.linear_damp, 0.0, 0.001)
+
+
 # --- speed offset oscillation ---
 func test_oscillation_never_drops_below_min_speed() -> void:
 	var effect := _make_oscillation_effect(2.0)
--- a/scenes/court.tscn
+++ b/scenes/court.tscn
@@ -10,6 +10,7 @@
 [ext_resource type="PackedScene" uid="uid://mws700v36gko" path="res://scenes/gear_rack.tscn" id="9_c7axm"]
 [ext_resource type="PackedScene" uid="uid://3l88siilc0hu" path="res://scenes/shipment_mat.tscn" id="10_5mv5b"]
 [ext_resource type="PackedScene" path="res://scenes/miss_zone.tscn" id="10_miss_zone"]
+[ext_resource type="Script" uid="uid://dv6u410fejjs7" path="res://scripts/entities/miss_zone.gd" id="1_miss_zone_script"]
 [ext_resource type="Script" uid="uid://cp16q4n8se0mu" path="res://scripts/court/speed_bar.gd" id="11_2768l"]
 [ext_resource type="PackedScene" uid="uid://kyv43r8jy2b" path="res://scenes/recruit_panel.tscn" id="12_rnrgf"]
 [ext_resource type="Script" uid="uid://b8611ynp8yggm" path="res://scripts/court/volley_counter.gd" id="14_vc"]
@@ -37,7 +38,7 @@ ball_tracker = NodePath("BallTracker")
 player_paddle_scene = ExtResource("9_0tnpc")
 player_spawn = NodePath("PlayerSpawn")
 autoplay_controller = NodePath("AutoplayController")
-right_wall = NodePath("Bounds/Right Wall")
+right_wall = NodePath("Bounds/RightMissBand")
 partner_spawn = NodePath("PartnerSpawn")
 timeout_controller = NodePath("TimeoutController")
 
@@ -89,17 +90,23 @@ mouse_filter = 2
 position = Vector2(10.8, 441.6)
 shape = SubResource("RectangleShape2D_horz")
 
-[node name="Right Wall" type="StaticBody2D" parent="Bounds" unique_id=1091055427]
+[node name="RightMissBand" type="Area2D" parent="Bounds" groups=["miss_zones"] unique_id=1091055427]
 position = Vector2(420, -351.6)
-physics_material_override = ExtResource("1_0wfyh")
+script = ExtResource("1_miss_zone_script")
+releases_ball = true
 metadata/_edit_group_ = true
 
-[node name="ColorRect" type="ColorRect" parent="Bounds/Right Wall" unique_id=679820908]
-offset_right = 21.6
-offset_bottom = 703.2
-mouse_filter = 2
+[node name="CollisionShape2D" type="CollisionShape2D" parent="Bounds/RightMissBand" unique_id=2042401507]
+position = Vector2(10.8, 351.6)
+shape = SubResource("RectangleShape2D_vert")
 
-[node name="CollisionShape2D" type="CollisionShape2D" parent="Bounds/Right Wall" unique_id=2042401507]
+[node name="LeftMissBand" type="Area2D" parent="Bounds" groups=["miss_zones"] unique_id=1091055428]
+position = Vector2(-441.6, -351.6)
+script = ExtResource("1_miss_zone_script")
+releases_ball = true
+metadata/_edit_group_ = true
+
+[node name="CollisionShape2D" type="CollisionShape2D" parent="Bounds/LeftMissBand" unique_id=2042401508]
 position = Vector2(10.8, 351.6)
 shape = SubResource("RectangleShape2D_vert")
 
--- a/scripts/entities/miss_zone.gd
+++ b/scripts/entities/miss_zone.gd
@@ -2,3 +2,6 @@ class_name MissZone
 extends Area2D
 
 ## Area that fires body_entered on overlap; the ball hooks in via register_miss_zone().
+
+## Side bands set this true so the ball falls out of the weightless play volume on cross.
+@export var releases_ball: bool = false


scripts/core/court.gd
```diff
@@ -14,7 +14,8 @@ const MissZoneScene: PackedScene = preload("res://scenes/miss_zone.tscn")
 @export var player_paddle_scene: PackedScene
 @export var player_spawn: Marker2D
 @export var autoplay_controller: AutoplayController
-@export var right_wall: StaticBody2D
+## Right side miss band; partner miss zone re-parents here when a partner activates.
+@export var right_wall: Node2D
```
J-Melon, line 18: **mica** — suggestion (non-blocking): the field is now an Area2D miss-band, not a wall. The docstring papers over the rename. Renaming to `right_miss_band` (and the matching scene NodePath, the local `right_wall` references in `_activate_partner`, and the export type back to `MissZone` if appropriate) would remove the cognitive friction. Holding the old name keeps every future reader doing the same double-take you just documented.

scripts/entities/ball.gd
```diff
@@ -7,6 +7,8 @@ signal speed_changed(speed: float, min_speed: float, max_speed: float)
 signal pressed(ball: Ball)
 
 const SPEED_EMIT_THRESHOLD := 10.0
+## Side bands raise gravity so the missed ball falls out of the weightless play volume.
+const _ROLL_OUT_LINEAR_DAMP: float = 1.5
```
J-Melon, line 11: **mica** — nit: leading underscore on a file-scope `const` is inconsistent with `SPEED_EMIT_THRESHOLD` directly above. GDScript convention reserves the underscore prefix for instance-level privates. Suggest `ROLL_OUT_LINEAR_DAMP`.

scripts/entities/ball.gd
```diff
@@ -77,13 +80,20 @@ func _on_body_entered(body: Node) -> void:
...
 
-func _on_miss_zone_body_entered(body: Node) -> void:
-	if body == self:
-		missed.emit()
+func _on_miss_zone_body_entered(body: Node, zone: MissZone) -> void:
+	if body != self:
+		return
+	missed.emit()
+	if zone != null and zone.releases_ball:
+		# Side-miss: leave the weightless play volume so the ball rolls onto the venue floor.
```
J-Melon, line 94: **mica** — nit: the inline comment restates what the const docstring on `_ROLL_OUT_LINEAR_DAMP` already says. One of the two is redundant under the project's one-line comment rule — prefer keeping the const docstring and dropping this line.

J-Melon (commented):
**mica** — review for SH-367 side-miss zones.

Two naming issues and one constant-style nit. No blockers; logic and tests look sound.

scripts/core/court.gd
J-Melon, line 18: **porphy** — **suggestion (non-blocking).** The field is now typed `Node2D` and the scene path points at `Bounds/RightMissBand`, an Area2D miss band — not a wall. Per CLAUDE.md naming (full words, no misleading shorthand) consider renaming `right_wall` → `right_miss_band` (and the matching `.tscn` export) so the partner re-parent site reads true. Mechanical rename, safe to defer if you'd rather keep this PR scoped.

J-Melon (commented):
**porphy** — GDScript convention pass for SH-367.

Diff is clean on the core Volley conventions (`@export` over `@onready`, typed signatures, autoload usage, no new `class_name` so no `load()` concerns, signal wiring uses identifiers not strings). One naming nit inline.

J-Melon (commented):
**hal** (scene discipline)

Reviewed `scenes/court.tscn` for autoload order, `@tool` guards, UID stability, surgical-edit discipline, node-path discipline, and ext_resource sprawl.

- No autoload reorder, no new `@tool` scripts, no `.uid` movement. Clean on those axes.
- Edit is surgical: `Right Wall` StaticBody2D becomes `RightMissBand` Area2D in place, reusing `unique_id=1091055427`; `LeftMissBand` is a sibling addition with a fresh `1091055428`. Old `ColorRect` child is dropped, the existing `CollisionShape2D` (`unique_id=2042401507`) is preserved on the right band, and the left band gets a parallel `2042401508`. No delete-and-rebuild.
- `right_wall = NodePath("Bounds/RightMissBand")` stays scene-root relative. Good.
- Both bands carry `groups=["miss_zones"]`; partner re-parent target (`right_wall`) is a valid `Node2D` per the relaxed `@export` type in `court.gd`. `register_miss_zone` will fire on the partner Area2D independently of the parent band's own area — Godot allows nested Area2D and signals don't shadow.
- One new `[ext_resource]` for `miss_zone.gd`. Not an explosion, but note: the bands attach the script directly rather than instancing `scenes/miss_zone.tscn`. Two sources of truth for "what a MissZone is" now exist (the `.tscn` and the inline scripted Area2Ds in `court.tscn`). Non-blocking — flag for follow-up if `miss_zone.tscn` ever gains structure (extra children, default shape changes) the bands will silently drift.

No blocking findings on scene discipline.

J-Melon (commented):
**spinel** — signal/lifecycle review for the side-miss bands.

**issue (likely):** double-fire on the right side. `RightMissBand` is now an Area2D in group `miss_zones`, and `_activate_partner` parents a fresh `MissZone` *inside* it (also in `miss_zones`, shapes overlap). When a ball crosses the right band post-partner-activation, both Area2Ds fire `body_entered`, so `Ball._on_miss_zone_body_entered` runs twice → `missed` emitted twice → `Court._on_ball_missed` halves/zeros the streak twice and resets paddle-streak trackers twice in one frame. Pre-PR this was fine because the right wall was a StaticBody2D, not a miss zone. Either drop the partner miss-zone instantiation now that the wall itself misses, or move the partner zone elsewhere / out of the group.

**issue:** `Ball._registered_miss_zones` accumulates freed references. `Ball.register_miss_zone` appends every zone but nothing ever removes them — `BallTracker.unregister_miss_zone` only erases from the tracker list, not from balls. When the partner miss zone is `queue_free`d in `_deactivate_partner`, the `bind(zone)` Callable on every live ball is left dangling (auto-disconnect on free covers the signal, but the array still holds a freed `MissZone`). Idempotency check `_registered_miss_zones.has(zone)` then operates on a list with freed entries. Add a `Ball.unregister_miss_zone(zone)` and call it from `BallTracker.unregister_miss_zone` for every live ball, or filter freed entries on access.

**nit:** `_on_miss_zone_body_entered` reads `zone.releases_ball` without `is_instance_valid(zone)`. With `bind(zone)`, a queued-for-free zone still in the deferred-emit window can crash on property access. Cheap guard: `if zone != null and is_instance_valid(zone) and zone.releases_ball:`.

Files: `scripts/entities/ball.gd`, `scripts/core/court.gd`, `scripts/court/ball_tracker.gd`, `scenes/court.tscn`.

J-Melon (commented):
**arkose** (test-coverage-checker, mission Banana Tank) — review against AC-to-assertion mapping and the changed-line coverage from the latest Tests run on e8a8ced.

**Assertion quality on the new ball-side tests (lines 178-202): good.**
Both tests assert player-observable outcomes — gravity flips on, damping rises, weightless state holds when `releases_ball` is false. They drive the public signal entry (`zone.body_entered.emit(_ball)`) rather than poking the private handler, and the literal `1.0` for `gravity_scale` is the documented "rejoin gravity" sentinel rather than a tuning number, so the hardcoded value is load-bearing. The damping check uses `assert_gt(... , 0.0)` rather than the `_ROLL_OUT_LINEAR_DAMP` constant, which is the right call — it asserts "damping kicks in" without coupling the test to the specific 1.5 tuning.

**Gaps against the AC list.**

1. **Partner-miss re-parent AC has no test.** Coverage shows `scripts/core/court.gd::_activate_partner` (lines 141-194) at 0 hits across the run. The diff in that function — `_partner_miss_zone.releases_ball = true` and `right_wall.add_child(_partner_miss_zone)` against the new `Node2D` band parent — is the only place the band-parent-vs-StaticBody2D type change matters, and nothing exercises it. A test that constructs a stub `right_wall` `Node2D`, calls `_activate_partner`, and asserts the spawned `MissZone.releases_ball == true` and `get_parent() == right_wall` would close this. Without it, the export-type change from `StaticBody2D` to `Node2D` is unverified.

2. **Velocity-preserved-on-miss AC has no assertion.** `test_releasing_zone_raises_gravity_on_cross` sets gravity and damping but never reads `linear_velocity` before/after the cross. Given the spec-vs-existing-test conflict you flagged, the assertion you want is the one this PR can defend: `linear_velocity` is unchanged in magnitude and direction across the `body_entered` emission. A two-line `var v_before := _ball.linear_velocity` / `assert_eq(_ball.linear_velocity, v_before)` makes the contract explicit and would catch any future "kill velocity on miss" regression in either direction.

3. **Idempotent registration test silently changed semantics.** `test_register_miss_zone_is_idempotent` (line 166) now passes because the new `register_miss_zone` early-returns on duplicates rather than relying on `is_connected`. That's fine, but the test no longer covers the null-zone guard added on line 89 (`if zone == null or _registered_miss_zones.has(zone)`). A `_ball.register_miss_zone(null)` call asserting no crash and no signal connection would lock that branch in — it's currently uncovered.

4. **Mirror-band AC (left band exists, fires `ball_missed`) is structural and lives in `court.tscn`.** No unit test can assert this and that's acceptable, but flag it for the runtime-verifier pass — the test suite alone does not prove the left band wires to `BallTracker`.

**Coverage floor.** `ball.gd` changed lines all hit (`82=10, 83=1, 84=9, 89=9, 90=1, 91=8, 92=8, 94=1, 95=1`). `court.gd::_activate_partner` is below the 75% floor on the changed surface (0% on the modified function) — that's the headline gap.

**Verdict: ship-with-fixes.** The two new tests are well-shaped behaviour assertions. The partner-miss path and the velocity-preservation contract are the missing pieces; both are small additions, neither blocks the architectural direction.

scripts/entities/ball.gd
```diff
@@ -77,13 +80,20 @@ func _on_body_entered(body: Node) -> void:
...
+		gravity_scale = 1.0
+		linear_damp = _ROLL_OUT_LINEAR_DAMP
```
J-Melon, line 96: **galena** issue (blocker): the per-frame speed-lock at `_physics_process` line ~51 (`linear_velocity = linear_velocity.normalized() * speed`) runs unchanged on the released ball. Gravity adds a downward component, the lock renormalises and rescales to `speed`, so the ball cruises at constant magnitude on a slowly-skewing diagonal. `linear_damp = 1.5` is also fought by the same line (damp reduces magnitude, lock restores it). The spec's "retains velocity at crossing, decelerates against friction, comes to rest" is unreachable while `_physics_process` keeps locking speed. Either skip the lock once released (e.g. an `_is_released` flag the early-return tests) or convert the body so the lock no longer applies.

scripts/entities/ball.gd
```diff
@@ -77,13 +80,20 @@ func _on_body_entered(body: Node) -> void:
...
+	zone.body_entered.connect(_on_miss_zone_body_entered.bind(zone))
 
 
-func _on_miss_zone_body_entered(body: Node) -> void:
-	if body == self:
-		missed.emit()
+func _on_miss_zone_body_entered(body: Node, zone: MissZone) -> void:
+	if body != self:
+		return
+	missed.emit()
```
J-Melon, line 92: **galena** issue: `missed.connect(reset_speed)` in `_ball_setup` fires for every miss, releasing or not. Inside this callback `missed.emit()` runs `reset_speed` synchronously before the gravity flip below, so velocity-at-crossing is overwritten with `min_speed * direction`. The riebeck caveat is real and load-bearing for side-miss feel. Suggestion: gate the connect on the zone, e.g. only auto-wire `reset_speed` when the miss came from a non-releasing zone, or move the speed reset into the back-miss path explicitly.

J-Melon (commented):
**galena** — verdict: **block**.

Steel-manned the side-miss-bands plan against the spec and the live ball runtime. Two findings are load-bearing failures of player-facing behaviour, not polish; one is a quiet contradiction with the PR description. Tests pass while the feature does not.

**Load-bearing**

1. `_physics_process` re-locks `linear_velocity = linear_velocity.normalized() * speed` every frame. Once the band flips `gravity_scale = 1.0` and `linear_damp = 1.5`, gravity nudges velocity downward, the next frame normalises and rescales to `speed`, and damp reduces magnitude only to be restored on the same line. Net result: the released ball cruises at constant `min_speed` along a slowly-skewing diagonal, never decelerating, never resting. The spec's "retains whatever velocity it had at the cross, decelerates against friction, and comes to rest" cannot happen while that physics-lock is in place.

2. The released body is still a `Ball` (live `RigidBody2D`), not a `HeldBody`. `BallDragController` only adopts `HeldBody` instances via `_spawn_loose_body_at` / `_on_loose_body_pressed`. There is no path that converts a missed live ball into a draggable loose body, so the spec's drag-it-back-to-the-rack flow is unreachable on this branch. The ball that exits sideways becomes an unpickable nuisance.

**Spec drift the other way**

3. `court.gd::_activate_partner` sets `_partner_miss_zone.releases_ball = true`. The partner miss zone sits behind the partner paddle — that is a back-miss surface, and the PR description explicitly says back-miss stays non-releasing. The releasing flag should not propagate to the partner zone; otherwise both back-miss lines release while only the side bands should.

**Other observations** (inline)

- New unit test exercises `zone.body_entered.emit(_ball)` directly, which never lets `_physics_process` run and so green-lights finding #1.
- `riebeck`'s magnitude-reset caveat: `_ball_setup` line ~181 wires `missed → reset_speed` unconditionally, so even without finding #1 the velocity at crossing is overwritten with `min_speed * direction` synchronously inside the `body_entered` callback before the gravity flip. The order is fine; the unconditional connect is the issue. Fix: gate the reset on `releases_ball == false`, e.g. only the back-miss zone should reset speed.
- `@export var right_wall: Node2D` keeps the legacy name while the node is now a `RightMissBand` Area2D. Misleading on read; rename to `right_miss_band` while the rename is cheap.
- Apex interaction (PR #600, still open): if apex code iterates over tracked balls and forces `gravity_scale = apex_gravity_scale` per-frame, it will undo the release. Needs an explicit "released balls opt out" flag or merge-order discipline.

**To unblock**

- Decide what a released ball is: keep it as `Ball` and skip the per-frame speed-lock once released, or convert it to a `HeldBody` on cross and route through the existing loose-body host. Either path resolves findings #1 and #2; punting both to follow-up means the merged feature does not behave as the spec describes.
- Gate `missed → reset_speed` on `releases_ball == false`, or move the speed-reset into the back-miss path explicitly.
- Don't set `releases_ball = true` on the partner miss zone.
- Add a runtime test (or at least the runtime-verifier check the PR body promises) that confirms the released ball decelerates and reaches a rest position.

scripts/entities/ball.gd
```diff
@@ -77,13 +80,20 @@ func _on_body_entered(body: Node) -> void:
...
+	if zone != null and zone.releases_ball:
+		# Side-miss: leave the weightless play volume so the ball rolls onto the venue floor.
+		gravity_scale = 1.0
```
J-Melon, line 95: **galena** issue (blocker): a side-miss leaves this node as a live `Ball` (`RigidBody2D`). `BallDragController` only knows how to adopt `HeldBody` instances via `_on_loose_body_pressed` / `_adopt_loose_body_as_held`. Nothing converts a released `Ball` into a draggable loose body, so the spec's "drag it from the venue floor onto the BallRack" path does not exist after this PR. Either spawn a `HeldBody` on cross and free the live ball, or extend the drag controller to accept `Ball` as a loose-body source. Without one of those, the released ball sits on the floor unpickable.

scripts/core/court.gd
```diff
@@ -165,6 +166,7 @@ func _activate_partner() -> void:
 
 	if right_wall != null:
 		_partner_miss_zone = MissZoneScene.instantiate()
+		_partner_miss_zone.releases_ball = true
```
J-Melon, line 169: **galena** issue: `_partner_miss_zone.releases_ball = true` makes the partner's back-miss surface release the ball, which contradicts the PR description ("the back-miss zone behind the player paddle stays non-releasing") and the spec's intent (only the side bands release). The partner zone is a back-miss line behind the partner paddle, not a side band; it should keep the default `false` and let `reset_speed` handle it. Drop the `releases_ball = true` line.

scripts/core/court.gd
J-Melon, line 18: **galena** suggestion: the variable is now misleading — the bound node is `Bounds/RightMissBand`, an `Area2D`, not a wall. Rename to `right_miss_band: Area2D` (or `MissZone` if you want the type to enforce intent). Cheaper to rename now than at every future read site.

tests/unit/ball/test_ball.gd
```diff
@@ -175,6 +175,33 @@ func test_register_miss_zone_is_idempotent() -> void:
...
+func test_releasing_zone_raises_gravity_on_cross() -> void:
+	var zone := MissZone.new()
+	zone.releases_ball = true
+	add_child_autofree(zone)
+	_ball.register_miss_zone(zone)
+	_ball.gravity_scale = 0.0
+
+	zone.body_entered.emit(_ball)
+
+	assert_almost_eq(_ball.gravity_scale, 1.0, 0.001)
```
J-Melon, line 188: **galena** issue: this test calls `zone.body_entered.emit(_ball)` directly, which sidesteps `_physics_process` entirely. The speed-lock interaction (constant magnitude after release, damp neutralised) cannot show up in this shape, so the green light here is misleading. Add a test that runs the body for a few frames after the cross and asserts magnitude actually decreases — that's the test that would catch the blocker.

PR #603, opened 2026-05-10
docs(court): friendship uplift, apex return, centripetal above bound

J-Melon:
The court-bounds spec catches up with the SH-309 spike outcome and the Banana Tank architecture call. Top edge becomes the friendship-bound, not a hard ceiling. Friendship is reframed as uplift: it cancels gravity and locks speed while the ball is in play, and crossing out of the play volume (above the bound or past either side band) releases the uplift. Above the bound, an additional centripetal force scaled by speed bends velocity perpendicular toward the play volume. Past either side band there is no centripetal; the ball falls under real gravity and rolls onto the venue floor.

This sets the canon that the SH-366 apex impl, the SH-367 side-miss impl, and the SH-369 velocity-preservation follow-up all rebuild against.


J-Melon (commented):
**coleus** — devils-advocate pass on Banana Tank's friendship-uplift canon. Verdict: **ship-with-fixes**. The reframing is a real upgrade on the speed-lock-as-curvature model, and the prose lands. But three things need clarifying in this PR before SH-366/367/369 build against it, and one is a hard contradiction with 21-ball-dynamics that will bite the apex implementer on day one.

**Headline issues**

1. *Direct contradiction with `21-ball-dynamics.md` § Friendship-bound apex return.* That doc still describes the apex as "a vertical-velocity flip at the bound's height rather than a wall collision" and calls 08-court-bounds the "placeholder." This PR makes 08-court-bounds canon and the model is no longer a velocity flip — it's constant gravity plus a speed-scaled centripetal. SH-366's implementer will read both files and pick the wrong one. Either update 21-ball-dynamics in this PR or add an explicit "21-ball-dynamics § apex section is superseded" line in 08-court-bounds.

2. *Pure-vertical and minimum-speed degeneracy is unaddressed.* "Centripetal force scaled by speed pulls velocity perpendicular toward the play volume" — perpendicular to what, in which direction? At a perfectly vertical trajectory at the lateral centre of the court, "perpendicular to velocity, toward play volume" is symmetric: left and right are equally valid. The implementer needs a tie-breaker (sign of horizontal velocity, last-paddle-side, deterministic bias). At very low speed the centripetal also goes to zero while gravity is constant, which means a ball that just barely crosses the bound with near-zero vertical velocity falls straight down through the bound on the same line it left — the loop you describe never forms. Worth naming a minimum speed, or noting that gravity dominates at low speed and that's intentional.

3. *Transition continuity claim is too strong.* The doc says "No interpolation, no blend window." Fine for the *flag*. But the velocity profile at the moment of crossing has a discrete jerk: above-bound, gravity is on and a centripetal force is added; the instant after crossing back, both vanish and speed is locked to whatever magnitude it has at that frame. A ball decelerated by gravity above the bound re-enters with reduced speed and that reduction is now permanent for the rally. Is that the design intent? It's a meaningful change from the previous model where the rally's speed was preserved by the flip. Spell it out: "crossing back resumes uplift at the ball's current speed; gravity above the bound can bleed rally speed, and that's fine." Otherwise SH-369 will guess.

**Coupling and gaps**

4. The *resting-ball drag-back* path (line 87) names the gesture but not the mechanism. `08-balls.md` describes drag of *live* balls via `BallDragController`/`HeldBody`. A rested ball is a `RigidBody2D` on the venue floor. Whether the existing controller can pick it up, or whether a new pickup path is needed, is unstated. This is a canon gap, not a doc bug — file or note it.

5. *Side-miss visual coherence.* Past a side band: gravity engages, speed-lock releases, **no centripetal**. A ball moving nearly straight up that clips a side band laterally will, under the new rule, suddenly start arcing under real gravity with full residual upward speed — sailing far above the friendship-bound height before falling. Visually that reads as "the ball escaped the sky." Is that intended, or should the friendship-bound still cap such balls? Worth one sentence either way.

6. *Per-venue bound height* (line 28) is a single line with no constraints. What's the floor? A venue with a low bound and a fast ball would spend most of the rally above the bound under gravity, which inverts the "weightless rally" feel. Either name a minimum-bound-relative-to-paddle-arc rule, or call out that venue authors must tune bound height against ball-speed tiers.

7. *Multi-ball with mixed states* is undiscussed. One ball above bound (under gravity + centripetal), one below (uplift). Different `gravity_scale`, different speed-lock state, simultaneous. Confirm this is per-body and there's no global toggle — current prose reads per-body but the previous 'court Area2D applies the treatment' phrasing has been deleted without a replacement saying "each ball owns its own state."

**Smaller**

8. "Friendship's uplift" is new vocabulary. It's good, and it's consistent with the spirit-of-the-volley framing in the same file and with `bible.md` § rally-apex. No conflict found with the FP-balance use of "friendship" elsewhere — those are points-as-currency, this is the bond-as-medium reading from `21-ball-dynamics.md` line 351 and `narrative/00-two-styles.md`. Fine to land.

**Tests / follow-up canon to consider**

- Runtime test: serve a ball straight up at very low speed from court centre. Document expected behaviour.
- Runtime test: serve at the highest tier speed and verify the loop tightens enough to not exceed venue height visually.
- Canon clarification: speed-bleed across the bound — preserved, partially preserved, or as-is.
- Canon clarification: rest-ball pickup mechanism.
- Update or supersede `21-ball-dynamics.md` § Friendship-bound apex return in this PR.

The core idea survives the pass. Land the contradiction fix and the degeneracy/intent clarifications in this PR; the rest can follow.

J-Melon (commented):
**cassava** — ship-with-fixes.

Relock-ramp shape is a real improvement; multi-ball mixed-state language reads well; side-miss path is still right post-ramp. Inline comments pin the ambiguities the next implementer will hit. Strongest opposing read: ship as-is, the prototype has one venue and SH-366 playtesting will surface the answers anyway. Fair, but #1 and #6 below are cheap to pin and the cost of not pinning them is a wrong implementation that ships and sets the feel.

designs/01-prototype/08-court-bounds.md
```diff
@@ -50,11 +60,17 @@ All three are audio + world-space only. No screen-space banners (per venue diege
...
+State applies per ball, so multi-ball mixed-state is well-defined: one ball above the bound under real gravity while another below stays in uplift.
+
+While a ball is alive inside the play volume it is held in friendship's uplift: `gravity_scale` is `0` and speed is locked. Linear damping is off; every bounce is pong-crisp, and energy comes from the paddle, not from falling.
+
+Crossing out of the play volume past either side band or above the friendship-bound flips that ball out of the uplift. `gravity_scale` rises to `1`, the speed-lock releases, damping kicks in. The ball retains its velocity at the moment of the cross, so a fast ball sails further before it lands and a slow one drops almost immediately.
+
+Above the friendship-bound a centripetal force scales with speed and pulls velocity perpendicular toward the play volume. The arc is what the ball does on its way back. Re-crossing below the bound restores the uplift on that ball; speed ramps back up to its pre-bound entry value so rally energy is preserved across the apex visit.
```
J-Melon: **cassava — issue (clarification):** "pre-bound entry value" is load-bearing and undefined. Per ball, fine — but stored where, overwritten on every upward cross, or sticky for the rally? If a paddle (or future tall-reach item) hits the ball *above* the bound and adds energy, does the downward relock use the original entry speed (energy lost) or the post-paddle speed (energy preserved)? Both defensible; the spec needs one.

J-Melon: **cassava — issue (precision):** "perpendicular toward the play volume" parses two ways. If perpendicular-to-velocity, the force does no work (matches 21-ball-dynamics line 358). If perpendicular-to-the-bound (i.e. straight down toward play), it *does* work and breaks the no-work claim. Say "perpendicular to current velocity, biased toward the play-volume side" to remove the ambiguity.

designs/01-prototype/08-court-bounds.md
```diff
@@ -50,11 +60,17 @@ All three are audio + world-space only. No screen-space banners (per venue diege
...
 
-A miss trigger (miss line or side band) flips the ball out of this state. `gravity_scale` rises to `1`. Linear damping kicks in. The ball retains whatever velocity it had at the moment of the cross, so a fast ball sails further before it lands; a slow one drops almost immediately. From there it rolls across the venue floor, decelerates against friction, and comes to rest.
+Past either side band there is no centripetal and no ramp. The ball falls under real gravity, rolls across the venue floor, and comes to rest.
 
-The transition is a single signal on `Area2D.body_exited`: clear the in-court flag, unlock gravity, engage damping. No interpolation, no blend window. The ball was in play and now it is not.
+The transition is a single signal on the boundary trigger per ball: clear the in-court flag, unlock gravity, release the speed-lock, engage damping. No interpolation, no blend window.
```
J-Melon: **cassava — nitpick:** the relock path restores uplift and ramps speed but doesn't explicitly say damping turns off. Implied by "uplift" (line 66) but worth one sentence so the implementer doesn't ship a ball that decelerates after every apex visit.

designs/01-prototype/08-court-bounds.md
```diff
@@ -50,11 +60,17 @@ All three are audio + world-space only. No screen-space banners (per venue diege
 
 ### Physics at the court boundary
 
-While the rally is alive, balls live in a physics volume that treats them as weightless. Their `gravity_scale` is `0` and linear damping is off; every bounce is pong-crisp, and energy comes from the paddle, not from falling. The court's `Area2D` is what applies this treatment.
+State applies per ball, so multi-ball mixed-state is well-defined: one ball above the bound under real gravity while another below stays in uplift.
```
J-Melon: **cassava — question:** multi-ball cross-collision at the bound is underspecified. Ball A above (real physics, no lock) hits ball B below (uplift, locked). Solver resolves symmetrically, B re-normalises to its locked speed post-tick, A keeps whatever momentum it got. Asymmetric energy. Probably the right answer at prototype scope, but worth one line.

designs/01-prototype/21-ball-dynamics.md
```diff
@@ -350,11 +350,13 @@ A press on any container's at-rest representation lifts the held preview, but th
...
+Mechanically the ball changes direction at apex. Diegetically nothing physical bounces it. Canon shape lives in [`08-court-bounds.md`](08-court-bounds.md) under "Friendship-bound and apex return".
 
 Implementation notes:
 
-- The active-movement state already covers it. The body is a `RigidBody2D` with `gravity_scale = 0` and frictionless momentum (see "Containers and the swap pattern" above); the apex return is a vertical-velocity flip at the bound's height rather than a wall collision.
+- Per-ball state. While inside the play volume the ball is held in friendship's uplift: `gravity_scale = 0`, speed locked. State applies per-body so multi-ball mixed-state (one above the bound, one below) is well-defined.
+- Crossing above the bound releases the uplift on that ball: gravity engages at constant world value, speed-lock releases, an additional centripetal force scaled by speed pulls velocity perpendicular toward the play volume. Gravity decelerates the vertical component; the centripetal force redirects without doing work.
```
J-Melon: **cassava — issue (numerical):** centripetal scaled by speed at fixed step can gain or lose energy at high tangential speed. At the Tier 3 ceiling in `20-ball-speed-tiers.md` this matters. Either note as a tuning concern or specify substepping / explicit re-projection onto the entry-speed circle each tick. The "without doing work" claim is mathematically clean; the integrator is what'll betray it.

designs/01-prototype/21-ball-dynamics.md
```diff
@@ -350,11 +350,13 @@ A press on any container's at-rest representation lifts the held preview, but th
...
+- Re-crossing below the bound restores the uplift. Speed ramps back up to its pre-bound entry value as the uplift relocks; rally energy is preserved across the apex visit. Without the ramp, gravity-bleed above the bound would drain rally speed.
```
J-Melon: **cassava — issue (cross-doc):** this line is precise ("redirects without doing work"); 08-court-bounds line 69 is looser ("perpendicular toward the play volume"). Implementer reading only one builds a different force. Tighten 08 to match, or echo "perpendicular to current velocity" here too.

designs/01-prototype/08-court-bounds.md
J-Melon: Resolved on `08-court-control.md` after merging main: the ball tracks a pre-bound entry value that updates on every speed change above the bound, so paddle hits or partner-active arcs above the bound feed the relock ramp.

J-Melon: Resolved on `08-court-control.md`: centripetal force is now explicit as perpendicular to velocity, toward the play volume, rotating velocity without doing work. Magnitude follows gravity, direction bends.

designs/01-prototype/08-court-bounds.md
J-Melon: Acknowledged. Damping engagement on cross-out and disengagement on relock left as implementation detail; the new `08-court-control.md` is more abstract than the previous `08-court-bounds.md` location and treats the friendship-uplift state as a logical pair (gravity + speed-lock) rather than enumerating every physics knob. Implementer brief picks up damping.

designs/01-prototype/08-court-bounds.md
J-Melon: Resolved on `08-court-control.md`: explicit note that cross-bound collisions resolve under each body's current physics state and the asymmetric energy resolution is acceptable. The bound is a state line, not a collision filter.

designs/01-prototype/21-ball-dynamics.md
J-Melon: Resolved on `08-court-control.md`: implementation should re-project velocity onto its target direction or substep at high speeds so "no work" holds numerically as well as mathematically; per-frame rotation gets sharp at Tier-3 speed ceilings.

designs/01-prototype/21-ball-dynamics.md
J-Melon: Cross-doc drift is now impossible: SH-309's restructure merged on main while this PR was open and reduced `21-ball-dynamics.md §Friendship-bound apex return` to a single-line redirect at `08-court-control.md`. Only one source of truth.

designs/01-prototype/20-ball-speed-tiers.md
```diff
@@ -54,7 +54,7 @@ Reaching a tier ceiling fires `on_tier_completed(tier_index)` on the item effect
...
+`Ball` gains `current_tier: int` and `tier_floor` / `tier_ceiling` derived from `current_tier` against a `SpeedTierTable` resource. Each tier entry in the table carries `{ floor, ceiling, max_range, reward }`, where `max_range` is the per-tier promotion of the flat `ball_speed_max_range` stat from design/21-ball-dynamics.md. Tier 0's `max_range` holds the existing flat value from design/21-ball-dynamics.md's base-stats tuning surface, so that surface keeps its meaning and lives on the Tier 0 entry. `increase_speed` and `set_speed_for_streak` clamp against `tier_ceiling` instead of `max_speed`. Crossing `tier_ceiling` triggers `_advance_tier` which emits `tier_advanced(new_tier)` and `on_tier_completed` through `ItemManager.process_event`.
```
J-Melon, line 57: **yarrow (suggestion, non-blocking):** the rewrite repeats `design/21-ball-dynamics.md` twice in one paragraph (`...flat 'ball_speed_max_range' stat from design/21-ball-dynamics.md. Tier 0's 'max_range' holds the existing flat value from design/21-ball-dynamics.md's base-stats tuning surface...`). The second mention can be `that doc's base-stats tuning surface` or just `the base-stats tuning surface`. Style guide cuts filler (#122) and the doubled path reads as one. Outside the scope of this PR's intent (link rewrites), so happy to leave.

J-Melon (commented):
**yarrow** — STYLE.md compliance pass on PR #603 (court-bounds friendship uplift).

**Verdict:** clean.

New prose in `design/08-court-bounds.md` and `tech/08-court-control.md` carries the voice. No em dashes, no AI-tell vocabulary, no banned constructions, no stray exclamation marks. Openings are particulars, not theses ("The court is open. The ball is held in by friendship, not by walls."). Paragraphs land on loaded sentences ("the ball on the venue floor is just a ball" earns its bathos). The table at the top of `design/08-court-bounds.md` is a structural choice the style guide neither blesses nor bars; it works as a reader's map before the prose unpacks each edge.

The other six files are link-path updates only.

One minor editorial note inline.

designs/01-prototype/design/08-court-bounds.md
```diff
@@ -0,0 +1,77 @@
...
+| Top | The friendship-bound | The ball is held in by friendship's uplift. Past the bound, gravity takes the ball through an arc and friendship draws it home. |
+| Bottom | The ground | A pong bounce. The floor is physical, hitting it doesn't end the rally. |
+| Back | The paddle line | The miss line. The ball getting past the paddle ends the rally. |
+| Sides | Open | A ball crossing sideways rolls onto the venue floor and lies among the items the player has placed. |
+
+No side walls and no ceiling. The court visibly opens onto the rest of the venue.
+
+## Friendship as the upper bound
+
+The top of the play volume is the friendship-bound. Below it the ball is weightless under friendship's uplift. Above it gravity engages, and a centripetal force scaled by the ball's speed bends it back into play.
```
J-Melon: **phlox** — issue (clarity): friendship vs spirit are doing the same metaphorical work and the doc never reconciles them. This section credits friendship's uplift for holding the ball; the Spirit-of-the-volley section credits the spirit for the same property ("loses what kept it weightless" = the same uplift this section names friendship). A reader will reasonably ask: when a miss fires, is friendship withdrawing or the spirit leaving? If they're the same, name it once. If friendship is the mechanic and the spirit is the rally's emotional charge layered on top, say so explicitly — one sentence at the seam between the two sections.

designs/01-prototype/design/08-court-bounds.md
```diff
@@ -0,0 +1,77 @@
...
+The bound height is per-venue. Small venues have a tighter ceiling, large venues breathe.
+
+## Miss
+
+A miss ends the rally. Two ways:
+
+- The ball gets past the paddle's miss line (back-miss).
+- The ball strays past either side band (side-miss).
+
+Both reset the counter. Both fire the same gravity-and-roll behaviour: the ball retains its velocity at the moment of the cross, rolls onto the venue floor, and comes to rest. The player drags it back to the rack to re-serve.
```
J-Melon: **phlox** — issue (consistency): "retains its velocity at the moment of the cross" promises continuity, but `tech/08-court-control.md` line 27 says "damping engages" the same instant. A sharp lateral hit that side-misses will feel like it punched invisible molasses if damping is non-trivial. Either soften this to "carries its velocity outward as it leaves the court, then rolls to rest" (acknowledges the deceleration the tech doc mandates), or have the tech doc clarify damping is the rest-roll friction kicking in over time, not an instantaneous velocity scrub on cross. Right now the two docs read in tension.

designs/01-prototype/tech/08-court-control.md
```diff
@@ -0,0 +1,45 @@
...
+
+A ball whose centre crosses either lateral side band fires a miss: uplift releases on that ball, gravity engages, speed-lock releases, damping engages, the rally counter resets. The ball retains its velocity at the moment of the cross, comes to rest on the venue floor, and the player drags it back to the rack via the existing live-ball drag path. Past either side band there is no centripetal and no relock ramp.
+
+Player-side and partner-side are the same event. The partner already has its own miss zone in code; the player-side gets the same treatment when its wall is removed.
+
+A live ball pulled to the rack mid-rally is not a miss; the drag controller replaces the live ball with a held body before any band crossing fires.
+
+## Cross-bound collisions
+
+Collisions between an above-bound ball and a below-bound ball resolve under each body's current physics state. Energy resolution is asymmetric across the bound and that is acceptable. The bound is a state line, not a collision filter.
```
J-Melon: **phlox** — issue (correctness): "asymmetric is acceptable" hand-waves the case the rewrite hasn't pinned down. Above-bound fast ball clips below-bound held-arc ball: the held ball is speed-locked under uplift. Collision imparts velocity. Does speed-lock immediately renormalise the held ball back to its prior locked speed (eating collision energy)? Re-lock to post-collision speed (pre-bound entry tracking only covers above-bound balls)? Pass through with no resolution? The asymmetry the doc accepts is on energy magnitude, not on whether the speed-lock contract holds. Take a position: I'd suggest "below-bound ball's speed-lock re-locks to post-collision speed; that becomes its new locked value until the next paddle hit." One sentence resolves the gap.

designs/01-prototype/tech/08-court-control.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Numerical stability
+
+At Tier-3 speeds the per-frame rotation is sharp. Implementation should re-project velocity onto its target direction or substep so "no work" holds numerically as well as mathematically.
```
J-Melon: **phlox** — nitpick (decisiveness): "re-project or substep" — pick one as the spec default and name the other as the fallback if profiling demands it. Leaving both equally weighted hands the choice to whichever implementer touches it first, and a future reader of this doc won't know which one the spec endorsed. Suggest: re-project as default (cheaper, deterministic), substep as escalation if Tier-3 still drifts.

designs/01-prototype/tech/08-court-control.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Open questions
+
+- **Energy loss on the rest-roll.** Venue-floor friction, ball damping, or both? Tuning, not correctness.
+- **Bound-height data shape.** `VenueConfig` Resource from day one, or a loose `@export` on `Court` promoted later?
```
J-Melon: **phlox** — issue (decisiveness): the energy-loss bullet is genuinely tuning and reads fine open. The other two are correctness questions in tuning-question clothing.

- **Bound-height data shape:** memory-rule precedent ("Resource over loose @export when a thematic cluster exists") points at VenueConfig from day one. Bound height + future centripetal scale + future gravity scale is a thematic cluster of three before this ticket is even merged. Loose `@export` then promote later is the more expensive path. Spec should say VenueConfig.
- **Drag-handoff frame window:** this is a correctness question. If a deferred swap can fire a spurious miss, that's a flaky bug nobody owns until it bites. Mandate it in spec: "miss does not fire while a grab is in flight" — one line, costs nothing, prevents the bug class entirely.

J-Melon (commented):
**phlox** — verdict: **ship-with-fixes**.

Clean rewrite. The per-ball state, centripetal-bend framing, pre-bound entry tracking, relock ramp, and substep note all land cleanly and address the prior battle's structural concerns. The two-doc split mostly works. Five issues below worth resolving before merge — three on the tech doc taking positions where it currently defers, two on the design doc's metaphor coherence and a contradiction with the tech doc on side-miss feel.

None are blocking; all are spec-takes-a-position calls that will save a future implementer (or future Josh) from re-litigating.

designs/01-prototype/design/08-court-bounds.md
```diff
@@ -0,0 +1,77 @@
...
+---
+
+## The bounds
+
+| Edge | Bound | What the player feels |
+|---|---|---|
+| Top | The friendship-bound | The ball is held in by friendship's uplift. Past the bound, gravity takes the ball through an arc and friendship draws it home. |
+| Bottom | The ground | A pong bounce. The floor is physical, hitting it doesn't end the rally. |
+| Back | The paddle line | The miss line. The ball getting past the paddle ends the rally. |
+| Sides | Open | A ball crossing sideways rolls onto the venue floor and lies among the items the player has placed. |
```
J-Melon: **privet** issue (non-blocking): "a centripetal force scaled by the ball's speed" is implementation vocabulary in a player-facing doc. The surrounding sentences ("wide loop", "tightens the arc", "bends it back into play") already carry the feel. Consider dropping the centripetal name and leaving it to `tech/08-court-control.md` to name the mechanism.

designs/01-prototype/design/08-court-bounds.md
J-Melon: **privet** issue (non-blocking): "retains its velocity at the moment of the cross" appears verbatim in `tech/08-court-control.md` miss section. It's an impl-phrasing tell that reads as bleed from the tech rewrite. The player-feel version could be "keeps its momentum and rolls" or similar; leave the precise phrasing to tech.

J-Melon (commented):
**privet** blocked at `164cd5e`.

Trim-verify on the substance: per-ball state, apex return, miss event, friendship-uplift, side-miss, cues, spirit, rest-flow, drag-out, temporary balls, helper-future all land in one file or the other. The side-miss cue distinction (softer, lower-pitched variant) is new canon introduced by the rewrite rather than a trim — flagging only so it's intentional.

Blocking: four inbound refs in the tree still point at the old `21-ball-dynamics.md` path. The rename is not complete. These are outside this PR's diff so I can't anchor inline; locations:

- `scripts/court/ball_tracker.gd:4` — `## Multi-ball ownership; spec lives in designs/01-prototype/21-ball-dynamics.md.`
- `scripts/items/ball_reconciler.gd:4` — `## Live-ball lifecycle owner; spec lives in designs/01-prototype/21-ball-dynamics.md.`
- `tests/unit/items/test_dragged_gravity_states.gd:1` — `## Dragged-gravity state-machine transitions; canon at designs/01-prototype/21-ball-dynamics.md.`
- `tests/integration/test_ball_regime_transitions.gd:1` — `## End-to-end ball regime transitions; see designs/01-prototype/21-ball-dynamics.md#regime-unification.`

All four want `designs/01-prototype/design/21-ball-dynamics.md`.

Mild bleed between design and tech rewrites flagged inline (non-blocking).

designs/01-prototype/tech/08-court-control.md
```diff
@@ -0,0 +1,53 @@
...
+
+The bound is a state line, not a collision filter.
+
+## Numerical stability
+
+The centripetal force only acts above the friendship-bound. After each physics tick that applied it, re-project velocity to its tracked magnitude so "no work" holds numerically as well as mathematically. Below the bound the speed-lock does the same job; no separate stabilisation is needed there.
+
+## Bound-height data shape
+
+The friendship-bound height lives on a `VenueConfig` Resource from day one. Per-venue tunables cluster on this Resource alongside the bound height; `Court` reads from it. Loose `@export` on `Court` is not the path.
```
J-Melon: **conifer** — *thought (non-blocking).* VenueConfig from day one is the right shape for where this is heading, but at prototype scale the only named per-venue tunable is the bound height itself. The cluster is a cluster of one. Worth either (a) naming the second and third tunables you expect to live here (gravity scale? side-band width? floor friction?) so the Resource isn't speculative, or (b) acknowledging in the doc that the Resource is provisioned ahead of the cluster forming, deliberately, to dodge a later migration. As written it reads like the cluster already exists.

designs/01-prototype/tech/08-court-control.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Drag-handoff frame window
+
+Miss does not fire while a grab is in flight. The drag controller's deferred swap creates a window where the live ball can cross a side band before being replaced by the held body; during that window the side-miss check skips. The check resumes once the swap completes or the gesture cancels.
```
J-Melon: **conifer** — *thought (non-blocking).* "Miss does not fire while a grab is in flight" is the right call but creates a narrow exploit shape: a player who initiates a drag on a ball already heading offside suppresses the miss. The deferred-swap window is one frame-ish, so at human reaction time this is theoretical, not practical. Flag for awareness if drag input ever gets predictive smoothing or input buffering — that's the point where the window stretches enough to matter.

J-Melon (commented):
**conifer** — confidence pass, ship.

The seven calls land cleanly across both docs and the previously-flagged inconsistencies are gone. Speed/velocity distinction is explicit in the tech doc and the design doc avoids the trap by speaking in feel terms only. The friendship-as-spirit bridge reads as one concept, not two stuck together — the opening line earns it and the body sustains it. Drag-handoff window is bounded by swap-or-cancel, re-projection is correctly scoped to the centripetal regime, VenueConfig from day one matches the Resource-over-loose-@export rule.

Two non-blocking thoughts inline. Neither moves the verdict.

J-Melon (commented):
**pye** — confidence pass on the friendship/bounds rewrite. Verdict: **ship**.

Fresh-reader cold-read holds together. TLDR lands the goal-then-change shape; `friendship.md` reads as the canonical seed and the design/tech split sits cleanly around it. Stale vocabulary (back-miss, VenueConfig, play volume, in-friendship's-uplift, frenzy, cue layering, rack footprint) is fully gone from the branch — grep clean. INDEX, `narrative/outline.md`, and `art/bible.md` all repoint to the new `design/21-ball-dynamics.md` path; code/test refs match.

Two nits inline, neither blocking. One out-of-scope flag on the lefthook removal so it's logged.

designs/01-prototype/tech/08-court-control.md
```diff
@@ -0,0 +1,53 @@
...
+
+Implementation spec for the wall-less court: friendship-bound replaces the top collider, lateral side bands replace the side wall colliders, the ground is unchanged. Drives SH-309. Player-facing design lives in [`../design/08-court-bounds.md`](../design/08-court-bounds.md).
+
+## Per-ball state
+
+Each live ball is in one of two states. The rule applies per ball, so multi-ball mixed-state is well-defined.
+
+**At or below the friendship-bound (held state).** `gravity_scale = 0`, speed is locked, linear damping is off.
+
+**Above the friendship-bound (released state).** `gravity_scale = 1`, speed-lock releases, linear damping engages. A centripetal force scaled by speed acts perpendicular to velocity, toward the play area. The centripetal force rotates velocity without doing work; magnitude follows gravity, direction bends.
```
J-Melon: **pye** — *suggestion (non-blocking).* Seam between design and tech vocabulary: design doc frames the whole play area as "held by friendship" with the bound as a soft ceiling that bends balls back into play, while the tech doc names above-bound the **released state**. A reader bouncing between docs has to reconcile whether friendship has *let go* above the bound or is *still holding* via the bend. Both readings are defensible, but the words pick different ones. Either rename to something like *bend state* / *uplift-bend state* in the tech doc, or add one sentence to the design doc that owns the release framing ("above the bound, friendship releases its hold and gravity does the returning work"). Confidence pass — flagging, not blocking.

designs/01-prototype/design/08-court-bounds.md
```diff
@@ -0,0 +1,41 @@
...
+
+Implementation spec lives in [`../tech/08-court-control.md`](../tech/08-court-control.md). Narrative canon for friendship lives in [`../../narrative/friendship.md`](../../narrative/friendship.md).
+
+**Dependencies:** Venue (`../08-venue.md`), Balls (`../08-balls.md`), Items (`../08-items.md`), Roles (`../08-roles.md`).
+
+---
+
+## The friendship-bound
+
+The top of the play area is the friendship-bound. Below it the ball is held by friendship. Above it gravity engages and the ball bends back into play; faster balls turn tighter, slower balls trace wider loops. Either way the ball returns and the held state resumes.
```
J-Melon, line 13: **pye** — *question.* The doc names the top (friendship-bound) and the sides (via Miss). The court's bottom edge inside the play area — what the ball does at ground level mid-rally — isn't mentioned here. Tech doc says "the ground is unchanged" which carries the implementation, but a design-doc reader cold may wonder whether the ball bounces off a court floor or whether the venue floor is the only ground surface. One line ("the ball bounces off the court floor as before; misses roll past the side bands onto the venue floor") would close the loop without re-introducing a bounds table.

designs/01-prototype/design/08-court-bounds.md
```diff
@@ -0,0 +1,41 @@
...
+
+A miss is one event: the ball arcs past a paddle on either side. The counter resets, the ball keeps its velocity, falls under gravity, and rolls to rest on the venue floor.
+
+Player-side and partner-side share the same event. The ball got past a paddle, regardless of whose paddle.
+
+## Resting balls
+
+A ball rolled out of the court sits visibly on the venue floor wherever it stopped. Rested balls stay put across rallies, saves, and scene reloads.
+
+To put a rested ball back into play, the player drags it onto the court for a manual serve, or drags it onto the rack to re-enter the auto-serve loop. Either path is supported and already implemented.
```
J-Melon, line 27: **pye** — *question (deliberate cut?).* The pre-PR doc carried a paragraph about the empty-rack idle: if the player owns exactly one permanent ball and it's resting, the rack is empty and MC idles, teaching the rack gesture before the helper arrives. That player-experience moment is dropped here. If it's now owned by the rack/items doc, fine; if it just fell out, worth re-homing — it was the doc's load-bearing argument for why a missed ball matters beyond the counter reset.

lefthook.yml
```diff
@@ -5,11 +5,6 @@ output:
   - execution_out
```
J-Melon, line 5: **pye** — *flag (out of stated scope).* Removing the `prepare-commit-msg` ticket-prefix hook is a behaviour change for every future commit on this repo — Josh now hand-prefixes `SH-N` in subjects, or the merge-gate ruleset gets loosened elsewhere. Not a blocker for a docs PR, but logging it because it's not in the brief's change list and the script file is also being deleted. Intentional?

PR #608, opened 2026-05-10
SH-366 feat: apex return arcs ball back into play

J-Melon:
Lands the ball state machine (`PLAY-NORMAL`, `PLAY-ARC`, `OUT-REST`, `OUT-HELD`, `STORED`) with per-state physics. Below the friendship-bound the ball stays in NORMAL with gravity off and the existing speed-lock; above the bound the ball flips to ARC, where gravity engages, damping turns on, and a centripetal force scaled by speed acts perpendicular to velocity toward the play area. Magnitude is re-projected each tick to the tracked entry value so the centripetal does no work and the speed stays steady through the curl, per the spec.

The entry-value register lives on the body. The first NORMAL→ARC cross sets it; subsequent crosses do not reset it; speed-change events while in ARC update it. On the ARC→NORMAL cross, speed ramps back to the tracked entry value over a configurable window, preserving rally energy across the apex visit.

Friendship-bound height and apex tunables live on a new `CourtConfig` resource at `resources/court_config.tres`. `Court` exposes a `court_config` export, propagates the resource through `BallTracker`, and balls receive it on `attach`.

Side-band miss events and the OUT-HELD ↔ OUT-REST drag-controller seams are scoped to SH-367 and not touched here; miss handling stays at parity with current main.

Live runtime verification (centripetal feel, rally counter mid-arc, apex feel against the canon arc) is owed to ramie, the runtime-verifier.


scripts/entities/ball.gd
```diff
@@ -22,16 +35,26 @@ var speed_increment: float
...
+var _entry_speed_initialised: bool = false
+var _relock_remaining: float = 0.0
+var _relock_from_speed: float = 0.0
 
 
 func _ready() -> void:
 	if _item_manager == null:
 		_item_manager = ItemManager
+	if court_config == null:
+		court_config = CourtConfig.new()
```
J-Melon: **arthur** issue: `CourtConfig.new()` on a class added this session can miss the async class-name cache; use `load("res://scripts/core/court_config.gd").new()` for the fallback.

scripts/core/court.gd
```diff
@@ -57,6 +58,9 @@ func _ready() -> void:
 		ball_tracker = BallTracker.new()
 		ball_tracker.ball_system = ball_system
 		add_child(ball_tracker)
+	if court_config == null:
+		court_config = CourtConfig.new()
```
J-Melon: **arthur** issue: same session-new class hazard as ball.gd; swap to `load("res://scripts/core/court_config.gd").new()` for the null-fallback.

scripts/entities/ball.gd
```diff
@@ -22,16 +35,26 @@ var speed_increment: float
 var effect_processor: BallEffectProcessor
 var is_temporary: bool = false
 
+# State-machine register: persists the pre-bound entry speed. First NORMAL->ARC cross sets it;
+# subsequent crosses do not reset it; speed-change events while in ARC update it.
+var play_state: int = PlayState.PLAY_NORMAL
```
J-Melon: **arthur** suggestion: type as `PlayState` rather than `int` so enum mismatches surface at edit time; same for the `signal play_state_changed(state: int)` payload.

scripts/entities/ball.gd
```diff
@@ -43,10 +66,87 @@ func _physics_process(delta: float) -> void:
...
+func _enter_normal() -> void:
+	gravity_scale = 0.0
+	linear_damp = 0.0
+	play_state = PlayState.PLAY_NORMAL
+	if _entry_speed_initialised and court_config.relock_ramp_seconds > 0.0:
+		_relock_remaining = court_config.relock_ramp_seconds
+		_relock_from_speed = linear_velocity.length()
+	else:
+		# Snap when the ramp is disabled or no entry value was tracked.
+		speed = entry_speed if _entry_speed_initialised else speed
```
J-Melon: **annona** (issue, non-blocking): `speed = entry_speed if _entry_speed_initialised else speed` — the `else speed` branch self-assigns. Read literally this line is dead in the un-initialised case. Either `if _entry_speed_initialised: speed = entry_speed` (drop the else) or guard the whole snap block on `_entry_speed_initialised` since the next line renormalises `linear_velocity` from `speed` regardless.

scripts/entities/ball.gd
```diff
@@ -43,10 +66,87 @@ func _physics_process(delta: float) -> void:
...
+func _advance_relock_ramp(delta: float) -> void:
+	_relock_remaining = maxf(_relock_remaining - delta, 0.0)
+	var ramp_total: float = court_config.relock_ramp_seconds
+	var t: float = 1.0 - (_relock_remaining / ramp_total) if ramp_total > 0.0 else 1.0
+	var ramped_speed: float = lerpf(_relock_from_speed, entry_speed, clampf(t, 0.0, 1.0))
+	speed = ramped_speed
+	linear_velocity = linear_velocity.normalized() * ramped_speed
+	if _relock_remaining <= 0.0:
+		speed = entry_speed
+		linear_velocity = linear_velocity.normalized() * speed
```
J-Melon: **annona** (issue, non-blocking): when `_relock_remaining <= 0.0`, `t` is already 1.0 so `lerpf` on line 144 returned `entry_speed` and lines 145-146 already wrote it. The final `if` block is duplicated work — drop lines 147-149.

scripts/entities/ball.gd
```diff
@@ -43,10 +66,87 @@ func _physics_process(delta: float) -> void:
...
+	# Re-project to the entering magnitude so the centripetal does no work.
+	var target_speed: float = entry_speed if _entry_speed_initialised else current_speed
+	if bent.length() > 0.0:
+		linear_velocity = bent.normalized() * target_speed
+
+
+func _advance_relock_ramp(delta: float) -> void:
+	_relock_remaining = maxf(_relock_remaining - delta, 0.0)
+	var ramp_total: float = court_config.relock_ramp_seconds
+	var t: float = 1.0 - (_relock_remaining / ramp_total) if ramp_total > 0.0 else 1.0
```
J-Melon: **annona** (nitpick): `t` is single-letter; rename to `progress` or `ramp_t`. Same line, the ternary `1.0 - (_relock_remaining / ramp_total) if ramp_total > 0.0 else 1.0` is correct by GDScript precedence but hard to read; split into a guarded `if ramp_total > 0.0:` block for clarity.

scripts/entities/ball.gd
```diff
@@ -5,6 +5,17 @@ signal missed
 signal at_max_speed_changed(is_at_max: bool)
 signal speed_changed(speed: float, min_speed: float, max_speed: float)
 signal pressed(ball: Ball)
+signal play_state_changed(state: int)
+
+# Ball state machine. PLAY covers NORMAL/ARC; OUT covers REST/HELD; STORED is on the rack.
+# See designs/01-prototype/tech/08-court-control.md. SH-367 lands the OUT/HELD transitions.
```
J-Melon: **annona** (issue): two-line `#` comment block. Per CLAUDE.md the project keeps comments to one line. Same pattern at lines 38-39 and 118-119 — collapse each to a single line or move the second sentence into the spec doc that's already referenced.

scripts/core/court_config.gd
```diff
@@ -0,0 +1,14 @@
+class_name CourtConfig
+extends Resource
+
+## Per-court tunables: friendship-bound height, apex physics knobs.
+## See designs/01-prototype/tech/08-court-control.md.
```
J-Melon: **annona** (issue): two-line `##` docstring. One-line comment policy applies to doc-comments too — the per-property `##` lines below are fine, but this header pair should collapse to a single line.

tests/unit/ball/test_ball_apex_arc.gd
```diff
@@ -0,0 +1,177 @@
+## State-machine, entry-value register, and relock-ramp tests for SH-366.
+## Spec: designs/01-prototype/tech/08-court-control.md.
```
J-Melon: **annona** (issue): two-line `##` file docstring violates the one-line comment rule.

tests/unit/ball/test_ball_apex_arc.gd
```diff
@@ -0,0 +1,177 @@
...
+	_ball.global_position = Vector2(0.0, BOUND_Y - 5.0)
+	_ball._physics_process(0.016)
+	assert_almost_eq(_ball.entry_speed, 700.0, 0.5)
+	# Knock magnitude off the entry value before re-crossing (simulates damping/gravity drift).
+	_ball.linear_velocity = Vector2(0.0, 200.0)
+	_ball.global_position = Vector2(0.0, BOUND_Y + 5.0)
+	_ball._physics_process(0.016)
+	assert_eq(_ball.play_state, Ball.PlayState.PLAY_NORMAL)
+	# Advance the ramp past completion.
+	for tick in 12:
```
J-Melon: **annona** (nitpick): `for tick in 12:` — the loop variable is unused. Convention is `for _i in 12:` (or `for _ in 12:`). Same on line 170.

J-Melon (commented):
**annona** — Code-quality review of SH-366 apex return v2. Flagging a few semantic issues gdlint won't catch: a self-assigning dead branch, a duplicated post-lerp snap, a single-letter ternary, plus several multi-line comments / docstrings that violate the one-line comment rule in CLAUDE.md. Tests have unused loop vars. Architecture and centripetal physics look right; entry-value register and ramp behaviour are well-covered.

resources/court_config.tres
```diff
@@ -0,0 +1,10 @@
+[gd_resource type="Resource" script_class="CourtConfig" load_steps=2 format=3 uid="uid://b66666sh366cc"]
+
+[ext_resource type="Script" path="res://scripts/core/court_config.gd" id="1_courtcfg"]
```
J-Melon: **marvin** (nit): the script `[ext_resource]` here uses `path=` only. The `scripts/core/court_config.gd.uid` sidecar exists (`uid://c4ayxy1m7fhfn`); including `uid="uid://c4ayxy1m7fhfn"` on this line matches the convention used throughout `court.tscn` and survives a script rename via `file_ops`. Godot will resolve it either way today, so non-blocking.

J-Melon (commented):
**marvin** — Scene/resource discipline review.

Clean pass on the load-bearing checks:

- `project.godot` untouched — autoload order (`SaveManager`, `ItemManager`, `ProgressionManager`, `ConfigHotReload`) preserved.
- No `@tool` scripts added.
- New `.tres` declares `uid="uid://b66666sh366cc"` inline; the `.tscn` `ext_resource` references the same UID — link is stable.
- `scenes/court.tscn` edit is surgical (one `[ext_resource]` line, one property assignment on the `Court` node). No delete-and-rebuild, no path-relative violations, no orphan child references.
- Script `.gd.uid` sidecar present for the new `court_config.gd`.

One minor note inline.

J-Melon (commented):
**ford** — test-coverage review for SH-366.

Assertion quality is good: tests assert player-observable state (play_state, entry_speed, linear_velocity.length, gravity_scale, linear_damp, signal emission) rather than the internal private vars (_relock_remaining, _relock_from_speed). State-machine transitions cover NORMAL→ARC and ARC→NORMAL. Entry-register persistence across multiple crossings is tested. Relock ramp has a landing assertion and an intermediate-bracket assertion (mid bounded by start and target — not a hardcoded midpoint). Hardcoded speeds 500/600/700 are test inputs, not derived constants — fine.

Coverage artifact (run 25636250093) shows ball.gd new helper bodies (`_enter_arc`, `_enter_normal`, `_apply_arc_physics`, `_advance_relock_ramp`, `_track_arc_speed_change`) all hit. The handful of "MISS" lines are GUT/godot-coverage quirks on multi-line expression continuations and `signal.emit()` calls — function bodies are exercised.

ACs covered:
- AC1 (crosses bound, returns, rally continues) — covered by the up-cross, down-cross, relock-landing, and reprojection tests.
- AC2 (volley counter, paddle hits valid above and below) — `test_speed_change_in_arc_updates_entry_value` exercises `increase_speed` in ARC. `reset_speed` and `set_speed_for_streak` got the same `_track_arc_speed_change` hook with no test — see inline.
- AC3 (no regression below the bound) — partially covered. `test_arc_to_normal_on_downward_cross` confirms gravity/damp reset; the post-ramp landing test confirms speed re-locks. See inline on the snap-path gap (`relock_ramp_seconds == 0.0`).

Two small inline asks. Approve once they land or get a one-line rationale.

scripts/entities/ball.gd
```diff
@@ -43,10 +66,87 @@ func _physics_process(delta: float) -> void:
...
+	play_state_changed.emit(play_state)
+
+
+# Centripetal force scaled by speed, perpendicular to velocity, toward the play area (downward).
+# Re-project the magnitude after integration to cancel the tick-by-tick drift, per the spec.
+func _apply_arc_physics(delta: float) -> void:
+	var velocity: Vector2 = linear_velocity
+	var current_speed: float = velocity.length()
+	if current_speed <= 0.0:
+		return
```
J-Melon: **eddie** — issue (blocking).

"Toward the play area" is reduced to "the perpendicular with positive Y". For a ball rising up-and-right with velocity `(+x, -y)`, the two perpendiculars are `(+y, +x)` and `(-y, -x)`. The positive-y branch picks `(+y, +x)` — bend is down-and-further-right, away from court centre. Canon (08-court-control.md L36, 08-court-bounds.md L13) says the ball *bends back into play*, faster tighter, slower wider; that requires the bend to point toward the court centre on the X axis too, not purely downward. A ball glancing outward will get pushed down before it gets pulled back, which reads as a ceiling more than a held curve.

Suggestion: pick the perpendicular whose dot with `(court_centre - position)` is positive, or at minimum gate sign on `direction.x` so the curl pulls inward.

scripts/entities/ball.gd
```diff
@@ -43,10 +66,87 @@ func _physics_process(delta: float) -> void:
...
+		linear_velocity = linear_velocity.normalized() * speed
+		_relock_remaining = 0.0
+	play_state_changed.emit(play_state)
+
+
+# Centripetal force scaled by speed, perpendicular to velocity, toward the play area (downward).
+# Re-project the magnitude after integration to cancel the tick-by-tick drift, per the spec.
+func _apply_arc_physics(delta: float) -> void:
+	var velocity: Vector2 = linear_velocity
+	var current_speed: float = velocity.length()
```
J-Melon: **eddie** — issue (non-blocking).

Pure-vertical tie-break is silent. For exactly-up velocity `(0,-v)` perpendicular is `(+v,0)` with `y=0`, so the `< 0.0` test never fires and the ball always curls right. For exactly-down `(0,+v)` perpendicular is `(-v,0)`, also `y=0`, so it always curls left. Deterministic asymmetry that canon doesn't authorise. Consider tie-breaking on `sign(position.x - court_centre.x)` so the ball curls toward court centre regardless of which axis dominates.

scripts/entities/ball.gd
```diff
@@ -43,10 +66,87 @@ func _physics_process(delta: float) -> void:
...
+	var direction: Vector2 = velocity / current_speed
+	# Two perpendiculars; pick the one with downward (positive Y) component, "toward play area".
+	var perpendicular: Vector2 = Vector2(-direction.y, direction.x)
+	if perpendicular.y < 0.0:
+		perpendicular = -perpendicular
+	var bend: Vector2 = (
+		perpendicular * (current_speed * court_config.arc_centripetal_coefficient * delta)
+	)
+	var bent: Vector2 = velocity + bend
+	# Re-project to the entering magnitude so the centripetal does no work.
```
J-Melon: **eddie** — question (blocking).

`arc_linear_damp = 0.6` is exposed on CourtConfig, but the re-projection on L134 forces `linear_velocity` magnitude back to `entry_speed` every tick the ball is in ARC. Engine-side damping shaves the magnitude before this runs, then the re-projection puts it back exactly. Net effect: damping is a no-op while in ARC. Either the canonical "damping on" was meant to act *between* re-projections (it doesn't, given the order) or the re-projection should preserve a damped target rather than `entry_speed`. As shipped the knob is a lie.

Resolve one way: remove the export, comment it as authored-but-cancelled, or change the target to a damped value.

scripts/entities/ball/ball.gd
```diff
@@ -43,10 +66,87 @@ func _physics_process(delta: float) -> void:
...
 	_emit_max_speed_if_changed()
 	if _should_emit_speed_changed():
 		_emit_speed_changed()
-	linear_velocity = linear_velocity.normalized() * speed
+	if play_state == PlayState.PLAY_NORMAL:
+		linear_velocity = linear_velocity.normalized() * speed
+
+
+# State transitions and per-state physics. Crossing is read off the body's current Y vs the bound.
+func _update_play_state(delta: float) -> void:
```
J-Melon, line 78: **eddie** — issue (non-blocking).

State flip uses strict `<` against `bound_y` with no hysteresis. A ball jittering at the bound (numerical noise, paddle struck right at the line) toggles `gravity_scale` and `linear_damp` every physics tick, plus emits `play_state_changed` each flip. Subscribers will see a chatter signal. A small dead-band (e.g. ±2 px) or edge-trigger guard would smooth it.

scripts/entities/ball/ball.gd
```diff
@@ -162,7 +273,10 @@ func _ball_setup() -> void:
 	speed = min_speed
 	effect_processor.sync_base_speed()
 	lock_rotation = true
+	# PLAY-NORMAL physics floor: gravity off, damping off, speed locked by _physics_process.
+	gravity_scale = 0.0
 	linear_damp = 0.0
+	play_state = PlayState.PLAY_NORMAL
 	linear_velocity = Vector2(min_speed, min_speed * 0.5).normalized() * speed
```
J-Melon, line 306: **eddie** — issue (blocking, scope-adjacent).

`_ball_setup` resets `play_state` to `PLAY_NORMAL` but does not reset `_entry_speed_initialised` or `entry_speed`. If a Ball instance is reused across rallies (rack-served balls are pooled in current code paths), a fresh serve inherits a stale `entry_speed` from the prior rally; first apex visit would relock to that ghost. Reset the register here too, or document why pool reuse is impossible.

scripts/entities/ball/ball.gd
```diff
@@ -5,6 +5,17 @@ signal missed
 signal at_max_speed_changed(is_at_max: bool)
 signal speed_changed(speed: float, min_speed: float, max_speed: float)
 signal pressed(ball: Ball)
+signal play_state_changed(state: int)
+
+# Ball state machine. PLAY covers NORMAL/ARC; OUT covers REST/HELD; STORED is on the rack.
+# See designs/01-prototype/tech/08-court-control.md. SH-367 lands the OUT/HELD transitions.
+enum PlayState {
```
J-Melon, line 10: **eddie** — issue (non-blocking).

Enum lists `STORED` but no transition writes it. Canon (08-court-control.md L23) requires `STORED → PLAY: serve`. PR description scopes "OUT/HELD" to SH-367 — STORED is neither. Either wire it on the rack/serve path or add a comment beside the enum saying "STORED transitions land in [SH-N]" so the dangling state isn't ambiguous.

scripts/entities/ball/ball.gd
```diff
@@ -91,16 +191,27 @@ func increase_speed() -> void:
 		return
 	speed = min(speed + speed_increment, max_speed)
 	_apply_speed()
+	_track_arc_speed_change()
 
 
 func reset_speed() -> void:
 	speed = min_speed
 	_apply_speed()
+	_track_arc_speed_change()
```
J-Melon, line 242: **eddie** — issue (non-blocking, future-proofing for SH-367).

`reset_speed` is connected to the `missed` signal (L285-286) and now calls `_track_arc_speed_change`. Today the state machine doesn't model OUT, so a miss fired while above the bound would still see `play_state == PLAY_ARC` and update `entry_speed` to `min_speed` immediately before SH-367's miss transition lands. Footgun. Either fence `_track_arc_speed_change` against in-flight miss events or have SH-367 flip state before `reset_speed` runs. Worth a TODO at minimum.

scripts/entities/ball.gd
```diff
@@ -43,10 +66,87 @@ func _physics_process(delta: float) -> void:
...
+	var target_speed: float = entry_speed if _entry_speed_initialised else current_speed
+	if bent.length() > 0.0:
+		linear_velocity = bent.normalized() * target_speed
+
```
J-Melon: **eddie** — nit.

`linear_velocity.normalized() * ramped_speed` — if `linear_velocity` is exactly zero (degenerate, e.g. ball lands right at the bound with zero residual velocity), `normalized()` returns zero and the ramp produces zero velocity, leaving the ball stuck. Same shape on L94/L150. Guard or document the precondition.

tests/unit/ball/test_ball_apex_arc.gd
```diff
@@ -0,0 +1,177 @@
...
+	_ball._physics_process(0.016)
+	# Subsequent ARC ticks: speed stays close to entry_speed despite the centripetal bend.
+	for tick in 5:
+		_ball._physics_process(0.016)
+		assert_almost_eq(
+			_ball.linear_velocity.length(),
+			500.0,
+			5.0,
+			"in-ARC magnitude held to entry value tick over tick"
+		)
```
J-Melon: **eddie** — nit.

No multi-ball test, despite the canon explicitly calling out per-ball state and "multi-ball mixed-state is well-defined" (08-court-control.md L7). Two balls, one in NORMAL and one in ARC, asserted via `play_state` independently — cheap to add and locks in the per-instance contract.

Also: tests call `_physics_process` directly, bypassing `_ready`, so the default-CourtConfig fallback path on L51 is uncovered. One test that omits `_config` injection and relies on the default would catch a future regression there.

J-Melon (commented):
**eddie** — devil's advocate, **ship-with-fixes**.

The state machine, register, and ramp shape match the canon and the tests cover them. The arc physics carry two concerns I'd want resolved before merge: the perpendicular selector reduces "toward the play area" to "downward", which mis-bends balls heading laterally outward; and the in-arc damping knob is silently cancelled by the re-projection. Plus a handful of smaller seams worth tightening so SH-367 lands cleanly on top of this.

Verdict: ship after the perpendicular selector and the dead-damping knob are addressed (or one of them is consciously deferred with a comment). Other inlines are nits or future-proofing.

tests/unit/ball/test_ball_apex_arc.gd
```diff
@@ -0,0 +1,177 @@
...
+	_ball._physics_process(0.016)
+	assert_almost_eq(_ball.entry_speed, 500.0, 0.5)
+	# Paddle hit while in ARC: increase_speed updates the register.
+	_ball.increase_speed()
+	assert_almost_eq(
+		_ball.entry_speed,
+		_ball.speed,
+		0.5,
+		"in-ARC speed-change events update the tracked entry value"
+	)
```
J-Melon, line 110: **ford** — **suggestion (non-blocking)**: AC2 says the volley counter and paddle hits remain valid above the bound. `_track_arc_speed_change` was wired into `increase_speed`, `reset_speed`, and `set_speed_for_streak`, but only `increase_speed` is exercised here. A one-liner asserting `entry_speed` tracks `set_speed_for_streak(n)` while in ARC would close the gap and guard against a future change to either helper.

tests/unit/ball/test_ball_apex_arc.gd
```diff
@@ -0,0 +1,177 @@
...
+	# Advance the ramp past completion.
+	for tick in 12:
+		_ball._physics_process(0.016)
+	assert_almost_eq(
+		_ball.linear_velocity.length(),
+		700.0,
+		1.0,
+		"post-ramp magnitude matches the tracked entry value"
+	)
+	assert_almost_eq(_ball.speed, 700.0, 0.5)
```
J-Melon: **ford** — **issue (non-blocking)**: the snap path in `_enter_normal` (lines 110-114 of ball.gd, taken when `relock_ramp_seconds <= 0.0`) is uncovered. AC3 calls out no regression below the bound, and a court_config with the ramp disabled is a shippable configuration. Worth one test that sets `relock_ramp_seconds = 0.0`, crosses up then down, and asserts `_ball.speed == entry_speed` immediately (no ramp ticks). Falsifies a future bug where the snap branch silently leaves speed at the post-arc drift value.

J-Melon (commented):
**ford** — inline asks on AC2 and AC3 coverage gaps; see summary comment above for the full assessment.

scripts/entities/ball.gd
```diff
@@ -43,10 +64,90 @@ func _physics_process(delta: float) -> void:
...
+		linear_velocity = linear_velocity.normalized() * speed
+		_relock_remaining = 0.0
+	else:
+		_relock_remaining = 0.0
+	play_state_changed.emit(play_state)
+
+
+## Centripetal force scaled by speed, perpendicular to velocity, X-component points toward court centre.
+## Magnitude is held by re-projection to entry_speed; linear damping in PLAY-ARC is intentionally absent
+## because the re-projection would silently cancel any damped value tick-by-tick.
```
J-Melon: **slartibartfast** — **nitpick (non-blocking):** three-line `##` docstring. Project comment policy caps docstrings at one line. Compress to a single `##` (the second/third lines explain a deliberate omission and can move to a one-line `#` inside the function near the re-projection, or drop entirely since the code is self-evident).

scripts/entities/ball.gd
```diff
@@ -43,10 +64,90 @@ func _physics_process(delta: float) -> void:
...
+	var direction: Vector2 = velocity / current_speed
+	var perpendicular: Vector2 = Vector2(-direction.y, direction.x)
+	# Pick the perpendicular whose X points toward court centre (X=0). Degenerate at pos.x==0: fall back to +Y.
+	var pos_x: float = global_position.x
+	var wants_negative_x: bool = pos_x > 0.0
+	var wants_positive_x: bool = pos_x < 0.0
+	if (wants_negative_x and perpendicular.x > 0.0) or (wants_positive_x and perpendicular.x < 0.0):
+		perpendicular = -perpendicular
+	elif pos_x == 0.0 and perpendicular.y < 0.0:
+		perpendicular = -perpendicular
```
J-Melon: **slartibartfast** — **suggestion (non-blocking):** the centripetal direction block has two booleans, a four-clause `if`, and an exact `pos_x == 0.0` tiebreak. Reads as `desired_x_sign = -signf(pos_x); if signf(perpendicular.x) != desired_x_sign and desired_x_sign != 0.0: perpendicular = -perpendicular; if desired_x_sign == 0.0 and perpendicular.y < 0.0: perpendicular = -perpendicular`. Same behaviour, half the surface area, and the float-equality on `pos_x` is gone — `signf` collapses the degenerate case naturally.

scripts/entities/ball.gd
```diff
@@ -43,10 +64,90 @@ func _physics_process(delta: float) -> void:
...
+	play_state = PlayState.PLAY_NORMAL
+	if _entry_speed_initialised and court_config.relock_ramp_seconds > 0.0:
+		_relock_remaining = court_config.relock_ramp_seconds
+		_relock_from_speed = linear_velocity.length()
+	elif _entry_speed_initialised:
+		speed = entry_speed
+		linear_velocity = linear_velocity.normalized() * speed
+		_relock_remaining = 0.0
+	else:
+		_relock_remaining = 0.0
```
J-Melon: **slartibartfast** — **nitpick (non-blocking):** `_relock_remaining = 0.0` appears in two of three branches and is the default state anyway. Hoist it: set `_relock_remaining = 0.0` once at the top of the function, then only the ramp branch overwrites it. Drops the `else` and one assignment.

J-Melon (commented):
**slartibartfast** — Three nits on the new arc physics. Polish from the prior sweep landed cleanly (single-letter `t` → `ramp_progress`, dead self-assign gone, snap path deduped). The new centripetal direction block is the main code-quality concern.

J-Melon (commented):
**fenchurch** — Volley-convention re-review of cb9b74c. Arthur's prior blocks resolved; no new convention issues.

**Verified fixes:**
- `load("res://scripts/core/court_config.gd").new()` fallback in `scripts/entities/ball.gd:_ready` and `scripts/core/court.gd:_ready` — session-new `CourtConfig` instantiated via path-load, no class-name cache race.
- `play_state` typed `PlayState` on the variable declaration and on the `play_state_changed(state: PlayState)` signal payload.

**Convention scan (clean):**
- All new funcs fully typed (args + return). No untyped seams.
- `@export` discipline preserved; no `@onready` regressions for node refs (`ball_system`, `ball_tracker`, `court_config`, `press_area` all `@export`).
- Tests construct session-new `CourtConfig` and `Ball` via `load(path).new()` — same path-load convention applied for `class_name` consumers under test.
- Naming stays full-words (`entry_speed`, `friendship_bound_y`, `arc_centripetal_coefficient`, `relock_ramp_seconds`).
- Signal wiring in `scripts/core/court.gd` uses identifier-style `connect`, no raw string signal names introduced.

No blockers from a Volley-convention angle.

J-Melon (commented):
**wonko** — re-review of test coverage at cb9b74c.

Ford's prior asks are all answered with assertions against player-observable state, not implementation:

- `test_set_speed_for_streak_in_arc_updates_entry_value` asserts `entry_speed == speed` after the streak-set call. Matches the behaviour Ford flagged: in-ARC speed mutations must update the tracked entry value so the post-apex relock lands at the right magnitude.
- `test_relock_ramp_zero_snaps_to_entry_speed` asserts both `speed` and `linear_velocity.length()` land at the entry value in a single tick when `relock_ramp_seconds == 0.0`. Covers the branch in `_advance_relock_ramp` where `ramp_total > 0.0` is false.
- Centripetal-direction tests do exercise the X-axis tie-break, not just sign-of-bend. With velocity (0, -500) and direction (0, -1), the raw perpendicular comes out (1, 0). Under the old "positive Y always" rule, X would always end positive — which would fail the positive-X test (ball at x=200 expecting negative X bend). The flip path `(wants_negative_x and perpendicular.x > 0.0)` is the one under test, and the negative-X mirror covers the no-flip branch. Both assertions are on `linear_velocity.x` sign, which is what the player sees as "ball bends toward centre."

Coverage on the changed lines in `ball.gd` looks complete for the player-facing behaviour: state machine, entry register persistence, in-ARC speed-event tracking via `increase_speed` + `set_speed_for_streak`, ramp lerp + ramp==0 snap, centripetal direction on both sides. Magnitudes use `assert_almost_eq` with tolerances rather than hardcoded math, and direction tests use sign-of-component rather than computed magnitudes — robust to refactor.

No further test-coverage blockers from me.

scripts/entities/ball.gd
J-Melon: **prak** — issue (canon deviation): the spec at `designs/01-prototype/tech/08-court-control.md` line 36 reads *"`gravity_scale = 1`, speed-lock off, **damping on**"* for PLAY-ARC. Trillian's drop of `arc_linear_damp` is correct that re-projection silently cancels it tick-by-tick — but that's an argument for amending the canon, not for quietly diverging from it.

The behavioural gap is real: above-bound paddle hits at high speed used to bleed energy through `linear_damp` between the hit and the next re-projection; now they don't. `_track_arc_speed_change` updates `entry_speed` to the post-hit value, so re-projection re-injects whatever damping would have removed. Whether that's desirable is a feel call, but the canon currently says one thing and the code does another.

Resolution options before merge: (a) amend the canon section *Keeping the speed steady above the bound* to record damping-off as the chosen approach with the re-projection rationale, or (b) restore `linear_damp` and let re-projection cancel it (cosmetic but matches spec).

I'd take (a) — the rationale in this comment is sound — but it needs to land in the canon, not just the comment.

J-Melon (commented):
**prak** — re-battle of cb9b74c after trillian's sweep. B2 direction, B3 damp removal, B4 pool reset all walked.

B2 lands. The X-component sign tie-break correctly bends balls toward x=0 in every off-centre case I traced, including pure-vertical velocity at off-centre x (where the original perpendicular already has `x != 0` and the sign comparison fires cleanly). The eddie-flagged pure-vertical degeneracy at extreme speeds is deferred per scope, and the only true tie at `pos.x == 0.0` with pure-vertical velocity is measure-zero.

B4 lands for the pooled-ball reuse path. `_ball_setup` clears `entry_speed`, `_entry_speed_initialised`, `_relock_remaining`, `_relock_from_speed`. Confirmed.

Two issues from the steel-man worth surfacing — one inline-suggestion, one nit. Neither blocks merge given the SH-367 boundary, but the damp deviation deserves an explicit acknowledgement before merge so the canon and the code agree on what the spec says.

**Verdict: ship-with-fixes** (the damp/canon reconciliation is a one-line decision; everything else is genuinely SH-367 territory).

scripts/entities/ball.gd
```diff
@@ -43,10 +64,90 @@ func _physics_process(delta: float) -> void:
...
+	var current_speed: float = velocity.length()
+	if current_speed <= 0.0:
+		return
+	var direction: Vector2 = velocity / current_speed
+	var perpendicular: Vector2 = Vector2(-direction.y, direction.x)
+	# Pick the perpendicular whose X points toward court centre (X=0). Degenerate at pos.x==0: fall back to +Y.
+	var pos_x: float = global_position.x
+	var wants_negative_x: bool = pos_x > 0.0
+	var wants_positive_x: bool = pos_x < 0.0
+	if (wants_negative_x and perpendicular.x > 0.0) or (wants_positive_x and perpendicular.x < 0.0):
```
J-Melon: **prak** — nit: the X-axis tie-break assumes the play area is centred on world x=0. That's true today but it's a hidden invariant. If a future venue ever offsets the court, the centripetal silently bends the wrong way for half the court with no test failure. A one-line `# Assumes court is centred on world x=0; revisit if CourtConfig grows a centre offset.` here is enough; SH-367 or later can promote the centre-X to `CourtConfig` if needed.

scripts/entities/ball/ball.gd
```diff
@@ -91,16 +192,27 @@ func increase_speed() -> void:
...
```
J-Melon, line 242: **prak** — note (deferred to SH-367, not blocking): `reset_speed` is connected to the `missed` signal at line 291. If a miss fires while the ball is in PLAY_ARC (above-bound side-band cross), `_track_arc_speed_change` will overwrite `entry_speed` with `min_speed` while `_entry_speed_initialised` stays true. Today this is harmless because the ball doesn't transition to OUT_REST yet (SH-367) and the post-miss flight isn't expected to re-cross upward.

When SH-367 lands the OUT transitions, the cleanup needs to either clear the entry-speed register at miss time, or guard `_track_arc_speed_change` to skip when transitioning out of PLAY. Worth a leading comment on `reset_speed` flagging this as SH-367 follow-up so it doesn't get missed in that work.

J-Melon (commented):
## Code Quality Battle (SH-366)

Two findings worth addressing; rest is noted in the scratchpad.

### 1. OUT_REST live-grab cancels into a stuck OUT_HELD ball

`scripts/items/ball_drag_controller.gd:266` — `grab_live_ball` unconditionally sets `_held_was_on_court = true`, but the function is the entry point for BOTH on-court live grabs and OUT_REST re-grabs (see comment lines 257-259, plus the `clear_loose_in_venue` call at 261).

When an OUT_REST-origin gesture cancels via the expansion-ring 2x timeout, `_cancel_to_source` enters the `origin == "live"` branch, then guards `deactivate` behind `is_on_court(item_key)` — which is false for an OUT_REST item we just cleared loose-in-venue on. No restoration happens, `_finalise_gesture` runs, and the ball is left in `OUT_HELD`: frozen, `collision_layer = 0`, `_suppress_miss_detection = true`. Permanently invisible to physics until the player grabs it again (which they can, since GrabArea pickability is suppressed in OUT_HELD too).

Fix shape: derive `_held_was_on_court` from `_item_manager.is_on_court(item_key)` at grab time, or add an OUT_REST restore branch (`enter_out_rest()` at the original position) in `_cancel_to_source` for the live-OUT_REST case.

Tied to the name being misleading — `_held_was_on_court` is set by control flow, not by querying placement. `_held_should_deactivate_on_cancel` would match its actual role.

### 2. Multi-line `##` docstrings violate comment policy

`scripts/items/ball_drag_controller.gd` lines 56-57, 322-324, 356-357, 437-440, 477-480, 487-488 are multi-line `##` blocks. CLAUDE.md says docstrings stay one line. Several also embed transient narrative ("Step 5 seam", "step 7 unifies", "retires at step 7", "in a future step") that ages into stale cruft the moment the step lands. Future-maintainer trap: a reader hitting line 477 in six months has to decide whether "step 7" is the current architecture or a removed plan.

scripts/core/physics/court_physics.gd
```diff
@@ -0,0 +1,8 @@
+class_name CourtPhysics
+extends Resource
+
+
+## Per-tick hook for the above-bound (PLAY-ARC) physics rule. Default is a no-op so a missing
+## or null physics resource degrades to engine-gravity-only behaviour without crashing.
+func step(_ball: RigidBody2D, _config: CourtConfig, _delta: float) -> void:
```
J-Melon: question: the entire job of this file is to pass?

scripts/core/physics/parabolic_arc_physics.gd
```diff
@@ -0,0 +1,8 @@
+class_name ParabolicArcPhysics
+extends CourtPhysics
+
+
+## Parabolic arc IS engine gravity acting on a body with gravity_scale = 1; the rule body is
```
J-Melon: question: again?

scripts/core/court.gd
```diff
@@ -57,6 +55,9 @@ func _ready() -> void:
 		ball_tracker = BallTracker.new()
 		ball_tracker.ball_system = ball_system
 		add_child(ball_tracker)
+	if court_config == null:
```
J-Melon, line 59: style: spacing

scripts/core/paddle_ai_controller.gd
```diff
@@ -51,20 +51,19 @@ func _on_tracker_ball_added(new_ball: Ball) -> void:
...
 
 func _physics_process(_delta: float) -> void:
-	if not _enabled:
+	if not _enabled or ball == null:
 		return
-	# Belt-and-braces: set_enabled refuses null-ball enable, so this is defence
-	# in depth against direct `_enabled = true` writes.
-	assert(ball != null, "PaddleAIController: ball must be set before enabling")
 	_maybe_resample_noise()
+	if not _ball_in_play():
```
J-Melon, line 66: style: spacing

scripts/core/paddle_ai_controller.gd
```diff
@@ -73,6 +72,12 @@ func _physics_process(_delta: float) -> void:
 		_drift_to_center()
 
 
+func _ball_in_play() -> bool:
+	return (
+		ball.play_state == Ball.PlayState.PLAY_NORMAL or ball.play_state == Ball.PlayState.PLAY_ARC
```
J-Melon: style: are these round brackets load-bearing?

scripts/core/paddle_ai_math.gd
```diff
@@ -4,33 +4,33 @@ extends RefCounted
 ## Pure math functions for paddle AI: ball intercept prediction and noise sampling.
 
 
-## Predicts where the ball will cross target_x, reflecting off arena walls.
-## Returns the y-position of the intercept. Clamped to arena bounds.
+## Predicts where the ball will cross target_x under PLAY-ARC parabolic physics.
+## Above bound_y (smaller y), gravity acts on vy; below, constant velocity.
```
J-Melon, line 8: style: too big

scripts/core/paddle_ai_math.gd
```diff
@@ -4,33 +4,33 @@ extends RefCounted
...
 
-	for _step in range(20):
-		if simulated_y >= arena_top and simulated_y <= arena_bottom:
-			break
-		if simulated_y < arena_top:
-			simulated_y = arena_top + (arena_top - simulated_y)
-		elif simulated_y > arena_bottom:
-			simulated_y = arena_bottom - (simulated_y - arena_bottom)
+	var sim_y: float = ball_position.y
+	var sim_vy: float = ball_velocity.y
```
J-Melon: style: var names not readable

scripts/court/ball_tracker.gd
```diff
@@ -1,7 +1,7 @@
 class_name BallTracker
 extends Node
 
-## Multi-ball ownership; spec lives in designs/01-prototype/design/21-ball-dynamics.md.
+## Multi-ball ownership; spec lives in designs/01-prototype/tech/02-ball-lifecycle.md.
```
J-Melon: style: doc ref in comment

scripts/court/ball_tracker.gd
```diff
@@ -40,6 +42,8 @@ func get_current_ball() -> Ball:
 func attach(new_ball: Ball) -> void:
 	if new_ball == null or _balls.has(new_ball):
 		return
+	if court_config != null:
```
J-Melon, line 45: style: spacing

scripts/entities/ball/ball.gd
```diff
@@ -0,0 +1,319 @@
...
+	OUT_HELD,
+}
+
+const PLAY_MATERIAL: PhysicsMaterial = preload("res://resources/ball/play.tres")
+const REST_MATERIAL: PhysicsMaterial = preload("res://resources/ball/rest.tres")
+
+# Default layer/mask the body holds while in PLAY/STORED/OUT_REST. HELD snapshots and restores from these
+# so per-ball authored values aren't required (ball.tscn ships with Godot defaults of 1/1).
+const PLAY_COLLISION_LAYER: int = 1
+const PLAY_COLLISION_MASK: int = 1
```
J-Melon: question: will these ever have a default other than 1?

scripts/entities/ball/ball.gd
```diff
@@ -0,0 +1,319 @@
...
+
+## Item key this ball represents; the system reads this on adoption to find the matching ItemDefinition.
+@export var item_key: String = ""
+## Authored Area2D that routes pointer presses to the grab hit-box; wired from the scene so the grab hit-box stays scene-based.
+@export var grab_area: GrabArea
+## Per-court tunables; injected by Court at attach time. Falls back to a default at construction.
+@export var court_config: CourtConfig
+
+var speed: float = 0.0
```
J-Melon: style: use :=

scripts/entities/ball/ball.gd
```diff
@@ -0,0 +1,319 @@
...
+
+# Persistent register: first NORMAL->ARC cross sets it; in-ARC speed events update it; relock targets it.
+var entry_speed: float:
+	get:
+		return _relock.entry_speed
+
+var _item_manager: Node
+var _emit_tracker: BallSpeedEmitTracker = BallSpeedEmitTracker.new()
+var _relock: BallRelockState = BallRelockState.new()
+# HELD suppresses miss-zone routing; cleared on any non-HELD enter_X.
```
J-Melon, line 45: style: spacing

scripts/entities/ball/ball.gd
```diff
@@ -0,0 +1,319 @@
...
+var _suppress_miss_detection: bool = false
+
+
+func _ready() -> void:
+	if _item_manager == null:
+		_item_manager = ItemManager
```
J-Melon, line 51: style: spacing

scripts/entities/ball/ball.gd
```diff
@@ -0,0 +1,319 @@
...
+	speed_increment = _item_manager.get_stat(&"ball_speed_increment")
+	_setup_effect_processor()
+	_ball_setup()
+
+
+func _physics_process(delta: float) -> void:
+	if linear_velocity == Vector2.ZERO:
+		return
+	effect_processor.process_frame(delta)
+	_update_play_state(delta)
```
J-Melon, line 68: style: spacing

scripts/entities/ball/ball.gd
```diff
@@ -0,0 +1,319 @@
...
+# State transitions and per-state physics. Crossing is read off the body's current Y vs the bound.
+func _update_play_state(delta: float) -> void:
+	if play_state != PlayState.PLAY_NORMAL and play_state != PlayState.PLAY_ARC:
+		return
+	var bound_y: float = court_config.friendship_bound_y
+	var above_bound: bool = global_position.y < bound_y
+	if above_bound and play_state == PlayState.PLAY_NORMAL:
+		_enter_arc()
+	elif not above_bound and play_state == PlayState.PLAY_ARC:
+		_enter_normal()
```
J-Melon, line 88: style: spacing

scripts/entities/ball/ball.gd
```diff
@@ -0,0 +1,319 @@
...
+	_relock.enter_arc(speed)
+	gravity_scale = 1.0
+	set_play_state(PlayState.PLAY_ARC)
+
+
+func _enter_normal() -> void:
+	gravity_scale = 0.0
+	var should_snap: bool = _relock.enter_normal(
+		linear_velocity.length(), court_config.relock_ramp_seconds
+	)
```
J-Melon, line 104: style: spacing

scripts/entities/ball/ball.gd
```diff
@@ -0,0 +1,319 @@
...
+
+func _emit_speed_changed() -> void:
+	_emit_tracker.record_speed(speed, min_speed, max_speed)
+	speed_changed.emit(speed, min_speed, max_speed)
+
+
+func _on_body_entered(body: Node) -> void:
+	if freeze:
+		return
+	if body.has_method("on_ball_hit"):
```
J-Melon, line 128: style: spacing

scripts/entities/ball/ball.gd
```diff
@@ -0,0 +1,319 @@
...
+
+func register_miss_zone(zone: MissZone) -> void:
+	if not zone.body_entered.is_connected(_on_miss_zone_body_entered):
+		zone.body_entered.connect(_on_miss_zone_body_entered)
+
+
+func _on_miss_zone_body_entered(body: Node) -> void:
+	if _suppress_miss_detection:
+		return
+	if body == self:
```
J-Melon, line 143: style: spacing

scripts/entities/ball/ball.gd
```diff
@@ -0,0 +1,319 @@
...
+
+
+func _on_missed() -> void:
+	enter_out_rest()
+
+
+# Single funnel for play_state writes. Idempotent: a same-state call is a no-op.
+func set_play_state(new_state: PlayState) -> void:
+	if play_state == new_state:
+		return
```
J-Melon, line 154: style: spacing

scripts/entities/ball/ball.gd
```diff
@@ -0,0 +1,319 @@
...
+	play_state = new_state
+	_apply_grab_area_pickable()
+	play_state_changed.emit(new_state)
+
+
+# Grab area swallows clicks even while frozen; disable it when the ball isn't grabbable so the rack slot below stays reachable.
+func _apply_grab_area_pickable() -> void:
+	if grab_area == null:
+		return
+	grab_area.input_pickable = (
```
J-Melon, line 166: style: spacing

scripts/entities/ball/ball.gd
```diff
@@ -0,0 +1,319 @@
...
+# PLAY: selects NORMAL or ARC by current Y vs the friendship bound.
+func enter_play() -> void:
+	_suppress_miss_detection = false
+	freeze = false
+	collision_layer = PLAY_COLLISION_LAYER
+	collision_mask = PLAY_COLLISION_MASK
+	linear_damp = 0.0
+	physics_material_override = PLAY_MATERIAL
+	var bound_y: float = court_config.friendship_bound_y if court_config != null else 0.0
+	var above_bound: bool = global_position.y < bound_y
```
J-Melon, line 194: style: spacing

scripts/entities/ball/ball.gd
```diff
@@ -0,0 +1,319 @@
...
+		gravity_scale = 0.0
+		set_play_state(PlayState.PLAY_NORMAL)
+
+
+# OUT_REST: gravity on, REST material, damping engaged. Body keeps its current velocity.
+func enter_out_rest() -> void:
+	_suppress_miss_detection = false
+	freeze = false
+	collision_layer = PLAY_COLLISION_LAYER
+	collision_mask = PLAY_COLLISION_MASK
```
J-Melon: suggestion: this is a lot of random vars to remember to set, move out to manager or similar?

scripts/entities/ball/grab_area.gd
```diff
@@ -0,0 +1,63 @@
...
+	if not input_event.is_connected(_on_input_event):
+		input_event.connect(_on_input_event)
+
+
+# Duplicate the authored sub-resource so per-instance inflation does not mutate the shared shape.
+func inflate_to(authored_radius: float) -> void:
+	var press_shape: CollisionShape2D = _find_collision_shape()
+	if press_shape == null:
+		return
+	var circle: CircleShape2D = press_shape.shape as CircleShape2D
```
J-Melon, line 29: style: spacing

scripts/entities/ball/grab_area.gd
```diff
@@ -0,0 +1,63 @@
...
+	local_circle.radius = authored_radius * hitbox_inflation
+	press_shape.shape = local_circle
+
+
+func _draw() -> void:
+	if not GrabArea.debug_visible:
+		return
+	var col: CollisionShape2D = _find_collision_shape()
+	if col == null:
+		return
```
J-Melon, line 44: style: spacing

scripts/entities/ball/grab_area.gd
```diff
@@ -0,0 +1,63 @@
...
+	for child in get_children():
+		if child is CollisionShape2D:
+			return child
+	return null
+
+
+func _on_input_event(_viewport: Node, event: InputEvent, _shape_idx: int) -> void:
+	if not (event is InputEventMouseButton):
+		return
+	var mouse_button: InputEventMouseButton = event
```
J-Melon, line 64: style: spacing

scripts/entities/ball/relock_state.gd
```diff
@@ -0,0 +1,49 @@
...
+	entry_speed = new_speed
+	initialised = true
+
+
+# Returns true when caller should snap speed to entry_speed; false when no action or ramping.
+func enter_normal(velocity_length: float, ramp_seconds: float) -> bool:
+	_remaining = 0.0
+	if not initialised:
+		return false
+	if ramp_seconds > 0.0:
```
J-Melon, line 34: style: spacing

scripts/entities/ball/speed_emit_tracker.gd
```diff
@@ -0,0 +1,32 @@
+class_name BallSpeedEmitTracker
+extends RefCounted
+
+const SPEED_EMIT_THRESHOLD := 10.0
+
+var _last_speed: float = 0.0
```
J-Melon: style: use :=

scripts/hud/dev_hud.gd
```diff
@@ -2,12 +2,16 @@ class_name DevHud
...
 
 @export var clearance_button: Button
+@export var dev_menu: PanelContainer
 
 
 func _ready() -> void:
 	clearance_button.visible = not ProgressionManager.is_shop_unlocked()
 	clearance_button.pressed.connect(_on_clearance_button_pressed)
 	ProgressionManager.shop_unlocked_changed.connect(_on_shop_unlocked_changed)
+	if dev_menu != null and dev_menu.has_method("add_overlay_toggle"):
```
J-Melon, line 13: style: spacing

scripts/hud/dev_menu.gd
```diff
@@ -50,3 +54,15 @@ func _add_toggle(panel: Control) -> void:
...
 	_content.add_child(checkbox)
+
+
+## Public hook so other DevHUD scripts can add non-panel toggles (overlay flags, debug switches) into the same menu list.
+func add_overlay_toggle(label: String, initial: bool, handler: Callable) -> CheckBox:
+	var checkbox := CheckBox.new()
+	checkbox.text = label
+	checkbox.button_pressed = initial
+	checkbox.focus_mode = Control.FOCUS_NONE
+	checkbox.toggled.connect(handler)
```
J-Melon, line 65: style: spacing

scripts/items/ball_reconciler.gd
```diff
@@ -1,7 +1,7 @@
 class_name BallReconciler
 extends Node
 
-## Live-ball lifecycle owner; spec lives in designs/01-prototype/design/21-ball-dynamics.md.
+## Live-ball lifecycle owner; spec lives in designs/01-prototype/tech/02-ball-lifecycle.md.
```
J-Melon: style: doc ref in comment

scripts/items/ball_reconciler.gd
```diff
@@ -13,43 +13,67 @@ const BallScene: PackedScene = preload("res://scenes/ball.tscn")
...
+
 	# Deferred so sibling listeners connect to ball_spawned before we emit.
 	call_deferred(&"adopt_pre_existing_balls")
 	call_deferred(&"_reconcile_initial_state")
 
 
-## Idempotent; safe to call repeatedly across scene reloads.
+func _has_save_manager_autoload() -> bool:
+	# Tests instantiate BallReconciler without the full autoload graph; guard
+	# the wiring so unit tests do not crash on missing SaveManager.
```
J-Melon: style: too much comment

scripts/items/ball_reconciler.gd
```diff
@@ -13,43 +13,67 @@ const BallScene: PackedScene = preload("res://scenes/ball.tscn")
...
+
+
+## Snapshot of live ball positions keyed by item_key. Every loose / at-rest / in-play
+## ball lives in the registry post-step-5, so the registry walk is the whole story.
+func collect_item_positions() -> Dictionary[String, Vector2]:
+	var positions: Dictionary[String, Vector2] = {}
+	for key: String in _balls_by_key:
+		var raw: Variant = _balls_by_key[key]
+		if not is_instance_valid(raw):
+			continue
```
J-Melon, line 58: style: spacing

scripts/items/ball_reconciler.gd
```diff
@@ -13,43 +13,67 @@ const BallScene: PackedScene = preload("res://scenes/ball.tscn")
...
+		positions[key] = ball.global_position
+	return positions
+
+
+## Idempotent; safe to call repeatedly across scene reloads. Authored Balls live as
+## siblings of the reconciler in the scene tree, so scan the parent here.
 func adopt_pre_existing_balls() -> void:
-	if _ball_host == null:
+	var parent: Node = get_parent()
+	if parent == null:
```
J-Melon, line 73: style: spacing

scripts/items/ball_reconciler.gd
```diff
@@ -123,6 +153,37 @@ func ensure_ball_for_key(
 	return ball
 
 
+## Venue-floor release path. Ensures a registry Ball at `position`, in OUT_REST,
+## carrying `velocity`. Does not touch ItemManager state; callers flip loose-in-venue / on-court
```
J-Melon: style: comment too big

scripts/items/ball_reconciler.gd
```diff
@@ -123,6 +153,37 @@ func ensure_ball_for_key(
...
+## overlays as appropriate so the rack filters render the right slots.
+func release_into_rest(item_key: String, position: Vector2, velocity: Vector2) -> Ball:
+	var ball: Ball = get_ball_for_key(item_key)
+	if ball == null:
+		ball = ball_scene.instantiate()
+		add_child(ball)
+		_apply_item_art(ball, item_key)
+		_balls_by_key[item_key] = ball
```
J-Melon, line 168: style: spacing

scripts/items/ball_reconciler.gd
```diff
@@ -123,6 +153,37 @@ func ensure_ball_for_key(
...
+	ball.enter_out_rest()
+	ball.linear_velocity = velocity
+	return ball
+
+
+## Spawns a STORED ball at `spawn_position` and registers it under `item_key`.
+func adopt_stored(item_key: String, spawn_position: Vector2) -> Ball:
+	var ball: Ball = ball_scene.instantiate()
+	add_child(ball)
+	ball.enter_stored()
```
J-Melon, line 183: style: spacing

scripts/items/rack_display.gd
```diff
@@ -77,13 +87,41 @@ func _build_slot(definition: ItemDefinition, slot_position: Vector2) -> Node2D:
...
 	slot.add_child(art_holder)
 	_attach_slot_input(slot, definition.key)
 	return slot
 
 
+## Slot stays empty when the registry owns a Ball for this key; the Ball renders the art itself.
+func _populate_art_holder(art_holder: Node2D, definition: ItemDefinition) -> void:
+	if _registered_ball_for(definition.key) != null:
+		art_holder.set_meta(&"source", &"ball")
+		return
```
J-Melon, line 100: style: spacing

scripts/items/rack_display.gd
```diff
@@ -77,13 +87,41 @@ func _build_slot(definition: ItemDefinition, slot_position: Vector2) -> Node2D:
...
+	art_holder.set_meta(&"source", &"definition")
+	art_holder.add_child(definition.art.instantiate())
+
+
+func _registered_ball_for(item_key: String) -> Ball:
+	if reconciler == null:
+		return null
```
J-Melon, line 107: style: spacing

scripts/items/rack_display.gd
```diff
@@ -77,13 +87,41 @@ func _build_slot(definition: ItemDefinition, slot_position: Vector2) -> Node2D:
...
+	return reconciler.get_ball_for_key(item_key)
+
+
+## World position of the slot for `item_key` under the rack's current ordering. Returns Vector2.ZERO if unknown.
+func get_slot_position_for(item_key: String) -> Vector2:
+	if slot_container == null:
+		return Vector2.ZERO
+	if _slot_markers.is_empty():
+		_cache_slot_markers()
+	var kit_keys: Array[String] = _item_manager.get_kit_items(role)
```
J-Melon, line 118: style: spacing

scripts/core/court.gd
J-Melon: **riebeck**

resolved: blank line added after add_child cluster (3334ff9)

scripts/core/paddle_ai_controller.gd
J-Melon: **riebeck**

resolved: blank lines separating guard, noise, in-play check, dispatch (3334ff9)

scripts/core/paddle_ai_controller.gd
J-Melon: **riebeck**

resolved: extracted state into a local, parens dropped (3334ff9)

scripts/core/paddle_ai_math.gd
J-Melon: **riebeck**

resolved: predict_intercept split with sim loop extracted to _simulate_intercept_y (3334ff9)

scripts/core/paddle_ai_math.gd
J-Melon: **riebeck**

resolved: sim-loop vars renamed to current_y, vertical_velocity, step_seconds; horizontal_speed at top (3334ff9)

scripts/court/ball_tracker.gd
J-Melon: **riebeck**

resolved: doc-ref dropped (3334ff9)

scripts/court/ball_tracker.gd
J-Melon: **riebeck**

resolved: blank lines between court_config/append/current/signal clusters (3334ff9)

scripts/entities/ball/ball.gd
J-Melon: **riebeck**

resolved: consts removed, collision_layer/mask inlined to 1; test refs updated (3334ff9)

scripts/entities/ball/ball.gd
J-Melon: **riebeck**

resolved: speed and is_temporary moved to := (3334ff9)

scripts/entities/ball/ball.gd
J-Melon: **riebeck**

resolved: blank between guards and var cluster in _ready (3334ff9)

scripts/entities/ball/ball.gd
J-Melon: **riebeck**

resolved: blank after the early-return guard in _physics_process (3334ff9)

scripts/entities/ball/ball.gd
J-Melon: **riebeck**

resolved: blank after early-return guard in _update_play_state (3334ff9)

scripts/entities/ball/ball.gd
J-Melon: **riebeck**

resolved: blank between snap branch and set_play_state in _enter_normal (3334ff9)

scripts/entities/ball/ball.gd
J-Melon: **riebeck**

resolved: blank after early-return guard in _on_body_entered (3334ff9)

scripts/entities/ball/ball.gd
J-Melon: **riebeck**

resolved: blank after early-return guard in set_play_state (3334ff9)

scripts/entities/ball/ball.gd
J-Melon: **riebeck**

resolved: blank after early-return guard in _apply_grab_area_pickable (3334ff9)

scripts/entities/ball/ball.gd
J-Melon: **riebeck**

resolved: blank between property bundle, bound calc, and branch in enter_play (3334ff9)

scripts/entities/ball/ball.gd
J-Melon: **riebeck**

resolved: blank between property bundle and effect_processor cluster in enter_out_rest (3334ff9)

scripts/entities/ball/ball.gd
J-Melon: **riebeck**

resolved: blank-separated clusters in _ball_setup (3334ff9)

scripts/entities/ball/ball.gd
J-Melon: **riebeck**

resolved: noted as follow-up; not addressed in this PR (3334ff9)

scripts/entities/ball/grab_area.gd
J-Melon: **riebeck**

resolved: blank between guard and circle resolve clusters (3334ff9)

scripts/entities/ball/grab_area.gd
J-Melon: **riebeck**

resolved: blank between guards in _draw (3334ff9)

scripts/entities/ball/grab_area.gd
J-Melon: **riebeck**

resolved: blank between event-type guard, button checks, and emit (3334ff9)

scripts/entities/ball/relock_state.gd
J-Melon: **riebeck**

resolved: blank after initialised guard and before return in enter_normal (3334ff9)

scripts/entities/ball/speed_emit_tracker.gd
J-Melon: **riebeck**

resolved: moved to := for the four primitive fields (3334ff9)

scripts/hud/dev_hud.gd
J-Melon: **riebeck**

resolved: blank between progression wiring and overlay block (3334ff9)

scripts/hud/dev_menu.gd
J-Melon: **riebeck**

resolved: blank between toggled.connect and content-append (3334ff9)

scripts/items/ball_reconciler.gd
J-Melon: **riebeck**

resolved: doc-ref dropped (3334ff9)

scripts/items/ball_reconciler.gd
J-Melon: **riebeck**

resolved: two-line comment removed (3334ff9)

scripts/items/ball_reconciler.gd
J-Melon: **riebeck**

resolved: blank between is_instance_valid guard and ball cluster (3334ff9)

scripts/items/ball_reconciler.gd
J-Melon: **riebeck**

resolved: blank after early-return and after the per-child guards (3334ff9)

scripts/items/ball_reconciler.gd
J-Melon: **riebeck**

resolved: three-line docstring collapsed to one line (3334ff9)

scripts/items/ball_reconciler.gd
J-Melon: **riebeck**

resolved: blank between ball-create branch and final mutation cluster (3334ff9)

scripts/items/ball_reconciler.gd
J-Melon: **riebeck**

resolved: blank between art-apply and registry insert (3334ff9)

scripts/items/rack_display.gd
J-Melon: **riebeck**

resolved: blank between slot-marker check, kit-lookup, and result (3334ff9)

scripts/items/rack_display.gd
J-Melon: **riebeck**

resolved: covered by the same get_slot_position_for spacing pass (3334ff9)

scripts/items/rack_display.gd
J-Melon: **riebeck**

resolved: covered by the same get_slot_position_for spacing pass (3334ff9)

PR #651, opened 2026-05-12
SH-316 feat!: paddle-offset return angle

J-Melon:
Where the ball strikes the paddle now sets the return angle. Centre hits return roughly along the incoming line; contact toward the top or bottom edge steepens the return toward a configurable cap. The seam is `BallEffectProcessor.process_hit`, which now takes the struck paddle so it can read a normalised contact offset off `Paddle.get_half_height()` and reconstruct post-bounce velocity from the offset and the new `paddle_return_angle_max_degrees` tunable. Speed and horizontal direction survive the hit, so the side-miss invariant holds.

The replaced scalar (`return_angle_influence`) and the only item that drove it (`double_knot`) retire in this PR. The items system is being reworked, so rather than repurpose the slot, the resource, scene, preload, and its level-gated test all go. The magnetism effect-processor code stays in place — only the item-driven test path leaves with double_knot. SH-393 tracks restoring a player-reachable magnetism item.

The four old tests under `test_return_angle_influence.gd` baked the scalar-lerp contract into their asserts; they are replaced by `test_paddle_offset_return_angle.gd`, which exercises centre/edge/symmetry/speed/direction/dormancy against a minimal paddle double that exposes `get_half_height()`.

Breaking change: any saved `BaseStatsConfig` (or its `.tres`) referencing `return_angle_influence` will silently drop that override on next load. Project-shipped `base_stats.tres` does not set it, so this is benign here.


J-Melon (commented):
.

J-Melon (commented):
.

J-Melon (commented):
.

J-Melon (commented):
.

J-Melon (commented):
.

J-Melon (commented):
.

J-Melon (commented):
.

J-Melon (commented):
.

J-Melon (commented):
.

J-Melon:
Adds three more SH-316 ACs on top of the existing contact-offset return, per `designs/01-prototype/tech/03-paddle-bounce.md`:

- **English.** The struck paddle's vertical velocity at the contact frame contributes to the bounce direction via a new `paddle_english_coefficient` stat on `BaseStatsConfig` (default `0.0`, so dormancy is the baseline and the value enters via item or tuning). A paddle moving into the ball pulls the return in the direction of paddle motion.
- **Min/max-angle clamps.** When the bounce is active (`paddle_return_angle_max_degrees > 0`), the resulting angle is clamped into a 3-87 degree band off horizontal. Module constants on `BallEffectProcessor`, not item-tunable — these are safety floors so centre hits never read as the ball ignoring the paddle and extreme edge hits never read as wall-sticking.
- **Speed preservation.** Already satisfied by the existing `direction * ball.speed`; english and clamps both operate on direction only.

Dormancy story:
- `paddle_return_angle_max_degrees = 0.0` (default): early-return, no behaviour change. The new english stat is read only inside the active branch.
- `paddle_return_angle_max_degrees > 0.0`, `paddle_english_coefficient = 0.0`: offset-only return with min/max-angle clamps applied.
- Both `> 0.0`: english adds to offset before clamping.

`test_centre_hit_keeps_incoming_direction` was reshaped into `test_centre_hit_returns_within_min_angle_band` to match the new min-angle floor. Three new tests cover english direction, the centre-hit clamp, and the max-angle ceiling under maxed english + offset.

tests/unit/ball/test_paddle_offset_return_angle.gd
```diff
@@ -0,0 +1,223 @@
...
+	_ball.speed = _ball.linear_velocity.length()
+	var original_velocity: Vector2 = _ball.linear_velocity
+
+	_ball.effect_processor.process_hit(_paddle)
+
+	assert_eq(_ball.linear_velocity, original_velocity)
+
+
+func test_angle_scales_linearly_with_max_degrees() -> void:
+	# Same edge offset at two non-zero max-degree settings; resulting angle should scale linearly.
```
J-Melon: Optional: this asserts dormancy at zero max-degrees but doesn't pin down that english stays dormant too. If a future refactor moves the english read above the `max_degrees <= 0.0` early-return in `_apply_paddle_offset_return`, the ball will start biasing on paddle motion even when the offset stat is off, and nothing in the suite will catch it. Consider an extra case here that sets english to a non-zero coefficient and gives the paddle a vertical velocity, then asserts `linear_velocity` is unchanged.

J-Melon (commented):
.

J-Melon (commented):
.

J-Melon (commented):
.

J-Melon (commented):
.

J-Melon (commented):
.

J-Melon (commented):
.

J-Melon (commented):
.

J-Melon (commented):
.

J-Melon (commented):
.

J-Melon (commented):
.

J-Melon (commented):
.

scripts/entities/ball/effect_processor.gd
```diff
@@ -1,6 +1,18 @@
...
+signal bounce_resolved(
+	struck_paddle: Paddle,
+	offset_norm: float,
+	target_angle: float,
+	incoming_y_sign: float,
+	horizontal_sign: float
+)
+
+# Ceiling stays a const; the floor is tunable via PaddleConfig.paddle_bounce_min_angle_degrees so items can grow the dead zone.
+const MAX_ANGLE_OFF_HORIZONTAL_DEGREES := 87.0
```
J-Melon: question: why not config?

scripts/entities/ball/effect_processor.gd
```diff
@@ -24,23 +36,34 @@ func sync_base_speed() -> void:
...
+		GameRules.base.ball_speed_min, &"ball_speed_min", item_manager
+	)
 	if not is_equal_approx(new_min, ball.min_speed):
 		_base_speed += new_min - ball.min_speed
 		ball.min_speed = new_min
-	ball.max_speed = ball.min_speed + item_manager.get_stat(&"ball_speed_max_range")
-	ball.speed_increment = item_manager.get_stat(&"ball_speed_increment")
+	ball.max_speed = (
+		ball.min_speed
+		+ Stats.resolve(GameRules.base.ball_speed_max_range, &"ball_speed_max_range", item_manager)
```
J-Melon, line 54: suggestion: this is getting a little chunky, could we separate concerns?

scripts/hud/dev_hud.gd
```diff
@@ -33,6 +34,13 @@ func _on_debug_overlay_toggled(pressed: bool) -> void:
 	_apply_debug_overlay(pressed)
 
 
+func _on_cone_follow_toggled(pressed: bool) -> void:
+	for overlay in get_tree().get_nodes_in_group(&"dev_overlays"):
+		if overlay is DevBounceOverlay:
+			(overlay as DevBounceOverlay).follow_last_hit = pressed
```
J-Melon: question: is this unsafe?

tests/unit/ball/test_paddle_offset_return_angle.gd
```diff
@@ -0,0 +1,259 @@
...
+
+
+func _build_with_stats_and_min_angle(
+	degrees: float, english: float, min_angle_bonus: float
+) -> void:
+	# Builds items whose `always` triggers contribute the requested stat values like real items would.
+	_manager = ItemFactory.create_manager(
+		self, "max_angle_kit", &"paddle_return_angle_max_degrees", &"add", degrees
+	)
+	if english != 0.0:
```
J-Melon, line 49: style: if spacing

tests/unit/ball/test_paddle_offset_return_angle.gd
```diff
@@ -0,0 +1,259 @@
+extends GutTest
+
+# Tests that paddle-contact offset drives the return angle: centre returns flat, edges steepen.
+
+const PADDLE_HALF_HEIGHT := 27.0
+const MAX_DEGREES := 30.0
+const MIN_ANGLE_DEG := 3.0
```
J-Melon: issue: hardcoded created impl detail

tests/unit/ui/test_dev_hud_cone_follow_toggle.gd
```diff
@@ -0,0 +1,32 @@
+## SH-316 DevHUD checkbox: "Cone follows last hit" flips DevBounceOverlay.follow_last_hit.
+extends GutTest
+
+const DevHudScene: PackedScene = preload("res://scenes/dev_hud.tscn")
+const DevBounceOverlayScript: GDScript = preload("res://scripts/hud/dev_bounce_overlay.gd")
+
+var _dev_hud: DevHud
+var _overlay: DevBounceOverlay
```
J-Melon: issue: don't test dev, remove

tests/unit/ball/test_paddle_offset_return_angle.gd
J-Melon: **Gru**

Added `test_zero_max_degrees_with_english_and_paddle_motion_leaves_velocity_untouched` in 8bbe5f0: max_degrees=0, english>0, paddle vertical velocity nonzero; asserts linear_velocity unchanged. Catches the early-return-ordering regression you flagged.

scripts/entities/ball/effect_processor.gd
J-Melon: **Gru**

Moved the ceiling to PaddleConfig as `paddle_bounce_max_angle_degrees` (default 87°) in 8bbe5f0; `_clamp_off_horizontal_and_vertical` resolves it via Stats now, matching the existing min-angle tunable. Carry-along: `dev_bounce_overlay.gd` switched off the old const onto the new stat too.

scripts/entities/ball/effect_processor.gd
J-Melon: **Gru**

Split `_sync_speed_limits` into `_sync_min_speed`, `_sync_max_speed`, and `_apply_speed_offset` in 8bbe5f0. Observable behaviour unchanged; tests still green.

scripts/hud/dev_hud.gd
J-Melon: **Gru**

Not unsafe: the `is DevBounceOverlay` on the line above narrows the type, so the cast was redundant. Dropped it anyway in 8bbe5f0; the assignment is now plain `overlay.follow_last_hit = pressed`.

tests/unit/ball/test_paddle_offset_return_angle.gd
J-Melon: **Gru**

Added the blank line in 8bbe5f0. Your round-2 ping promoted this to a real rule: `ai/skills/minions/implementer-nits.md` now says "blank line before every `if` statement" with the function-start and `elif`/`else` exemptions, and the `code-quality`, `test-author`, `integration-scenario-author` agent definitions reference the skill (892bf28). Bart swept `effect_processor.gd` for the rule in 774b36d.

tests/unit/ball/test_paddle_offset_return_angle.gd
J-Melon: **Gru**

Dropped the `MIN_ANGLE_DEG` / `MAX_ANGLE_DEG` consts in 8bbe5f0. Tests now read `GameRules.paddle.paddle_bounce_min_angle_degrees` per-test via a helper, so a default change in PaddleConfig won't leave the asserts stuck on the old number.

tests/unit/ui/test_dev_hud_cone_follow_toggle.gd
J-Melon: **Gru**

Deleted the file (and its .uid) in 8bbe5f0.

designs/01-prototype/design/items.md
```diff
@@ -0,0 +1,233 @@
+# Items
+
+Player-facing design for the prototype's item set. Item effects are gameplay-first: the effect is immediately perceptible. The player figures out what an item does by owning it, not by reading a description. Most items are causality-driven; passive stat boosts are a subset of the same system.
```
J-Melon: issue: no need to say player facing all design/ is player facing

scripts/entities/ball/effect_processor.gd
```diff
@@ -1,6 +1,15 @@
 class_name BallEffectProcessor
 extends Node
 
+## Debug-only signal: fired after every paddle bounce, post-clamp and post-english, so dev overlays can echo the resolved direction.
+signal bounce_resolved(
```
J-Melon, line 5: question: can this be if debugged?

scripts/entities/ball/effect_processor.gd
```diff
@@ -24,23 +33,45 @@ func sync_base_speed() -> void:
...
+func process_hit(struck_paddle: Paddle) -> void:
+	_apply_paddle_offset_return(struck_paddle)
 
 
 func _apply_magnetism(delta: float) -> void:
-	var magnetism: float = item_manager.get_stat(&"ball_magnetism")
+	var magnetism: float = Stats.resolve(
+		GameRules.base.ball_magnetism, &"ball_magnetism", item_manager
+	)
 	if magnetism <= 0.0 or paddles.is_empty():
```
J-Melon, line 77: issue (style): in general if should be spaced before, should be a memory for this?

scripts/entities/ball/effect_processor.gd
```diff
@@ -67,15 +98,74 @@ func _apply_magnetism(delta: float) -> void:
...
-	var biased_direction: Vector2 = current_direction.lerp(horizontal, influence).normalized()
-	ball.linear_velocity = biased_direction * ball.speed
+	var offset_norm: float = clampf(
+		(ball.global_position.y - struck_paddle.global_position.y) / half_height, -1.0, 1.0
+	)
+	var offset_angle: float = offset_norm * deg_to_rad(max_degrees)
+	var english_coefficient: float = Stats.resolve(
+		GameRules.paddle.paddle_english_coefficient, &"paddle_english_coefficient", item_manager
+	)
+	var english_angle: float = struck_paddle.velocity.y * english_coefficient
```
J-Melon, line 138: suggestion: extract to math away from here

scripts/hud/dev_bounce_overlay.gd
```diff
@@ -0,0 +1,193 @@
+class_name DevBounceOverlay
+extends Node2D
+
+## Debug-only overlay: draws the reachable bounce cone on each paddle and a
+## post-resolve marker arrow from the last contact point. Subscribes to
```
J-Melon: issue {style): comment too big

tests/unit/ball/test_paddle_offset_return_angle.gd
J-Melon: The other two vars are still hardcoded?

tests/unit/ball/test_paddle_offset_return_angle.gd
J-Melon: in general not just there

designs/01-prototype/design/items.md
J-Melon: **Gru**

Dropped the "Player-facing" qualifier in 774b36d. Opens with "Design for the prototype's item set." now.

scripts/entities/ball/effect_processor.gd
J-Melon: **Gru**

Wrapped the `bounce_resolved.emit(...)` in `if OS.is_debug_build():` in 774b36d. Signal declaration stays; emit only fires in editor and debug exports.

scripts/entities/ball/effect_processor.gd
J-Melon: **Gru**

Good call. Filed the rule in `ai/skills/minions/implementer-nits.md` (Blank-line spacing section): blank line before every `if`, exceptions only at function start and `elif`/`else` continuations. Wired the skill into the `code-quality`, `test-author`, and `integration-scenario-author` agent definitions in 892bf28 so sub-agents inherit it. 774b36d swept `effect_processor.gd` for the rule.

scripts/entities/ball/effect_processor.gd
J-Melon: **Gru**

Extracted `_blend_english_into_offset(offset_angle, english_angle) -> float` in 774b36d: holds the `is_zero_approx` branch and the `(absf + absf) * signf` formula. `_apply_paddle_offset_return` calls it.

scripts/hud/dev_bounce_overlay.gd
J-Melon: **Gru**

Collapsed to one line in 774b36d: `## Debug overlay: paddle bounce cones and a resolved-direction marker per last hit.`

tests/unit/ball/test_paddle_offset_return_angle.gd
J-Melon: **Gru**

Those two are test fixture inputs, not impl duplicates: the stub's self-reported half-height and the chosen non-zero scenario value for the stat. The brittle ones I dropped (`MIN_ANGLE_DEG` / `MAX_ANGLE_DEG`) asserted against production clamp bounds; these don't. Happy to inline if you want zero test-file consts.

scripts/core/paddle_ai_controller.gd
```diff
@@ -135,7 +134,7 @@ func _track() -> void:
 
 
 func _dodge() -> void:
-	var arena_half: float = GameRules.base_stats[&"arena_height"] / 2.0
```
J-Melon, line 138: suggestion: this is not needed any more now there are no walls?

J-Melon: **Gru**

Right, gone in 4253a16: dropped `_dodge`, `_is_ball_behind` (abstract + both overrides), and the two dodge tests in `test_partner_ai_controller`. 692 tests now, all green; CI gate clean.

tests/unit/paddle/test_partner_ai_controller.gd
```diff
@@ -72,33 +72,6 @@ func test_drifts_toward_center_when_ball_moving_away() -> void:
 	assert_lt(_paddle.velocity.y, 0.0, "should drift up toward center")
```
J-Melon, line 72: **Leela**

missing-coverage: with `_dodge` / `_is_ball_behind` removed, the partner now drifts to centre when the ball is behind the paddle (x >= paddle.x). The two deleted dodge tests covered the only ball-behind position. The remaining drift test (`test_drifts_toward_center_when_ball_moving_away`) sits at ball.x=100, paddle.x=300, ball moving left, i.e. ball-in-front-moving-away, not ball-behind. A test placing the ball at `PADDLE_X + 50` and asserting drift toward y=0 would lock in the new behaviour and prevent a silent regression if `_ball_approaching` semantics shift again.

.github/workflows/test.yml
```diff
@@ -62,6 +64,30 @@ jobs:
...
 
+      - name: Leak gate
+        run: |
+          set -e
+          sed -E 's/\x1b\[[0-9;]*m//g' test_output.txt > test_output.plain
+          fail=0
+          # Filter cold-cache UID warnings (godot#101677, #115205, #109636); see ai/scratchpads/godot-ci-uid-cache.md.
+          warnings=$(grep -nE '^(WARNING|ERROR|SCRIPT ERROR|USER WARNING|USER ERROR):' test_output.plain \
+            | grep -vE 'ext_resource, invalid UID:' \
+            | grep -vE 'Failed loading resource: res://' \
```
J-Melon: **Slartibartfast**

bug: `grep -vE 'Failed loading resource: res://'` filters every failed resource load, not just the cold-cache UID-pair class. A genuinely broken `.tres` or missing scene reference in a test will print `ERROR: Failed loading resource: res://...` and be swallowed by this filter; the gate then passes on a regression it was meant to catch.

The companion `ext_resource, invalid UID:` line is specific enough to be safe, but this one needs tightening. Options:
- Pair-match: only drop `Failed loading resource` lines whose preceding line is an `ext_resource, invalid UID:` warning.
- Path-anchor: filter only the known UID-cache-tripping paths (e.g. the autoload or addon roots that emit during cold import).
- Drop the filter entirely and rely on the second `--import` pass eliminating these in practice. If the double-import works, this filter shouldn't be needed at steady state.

Same fix needed in `scripts/ci/run_gut.sh` lines 21-22.

J-Melon: **Gru**

Tightened both gates in 0ccd474 with an awk pair-matcher: `Failed loading resource: <path>` is now dropped only when its path matches the immediately preceding `ext_resource, invalid UID: ... using text path instead: <path>` warning. Standalone Failed-loading ERRORs (broken `.tres`, missing scene reference, real regressions) still fail the gate.

tests/unit/paddle/test_partner_ai_controller.gd
J-Melon: **Gru**

Added `test_drifts_toward_center_when_ball_is_behind_paddle` in 0ccd474: ball at `PADDLE_X + 50`, paddle at y=200, asserts paddle drifts toward y=0 over 10 physics frames. Locks the fallback behaviour so a future `_ball_approaching` semantic shift would trip the test.

PR #652, opened 2026-05-12
feat: data-drive Ball state property values

J-Melon:
The four `enter_X` methods on `Ball` each wrote the same physics and collision properties (`freeze`, `collision_layer`, `collision_mask`, `gravity_scale`, `physics_material_override`) inline. The values were duplicated across STORED, PLAY_NORMAL, PLAY_ARC, OUT_REST, and OUT_HELD, and any tweak to a state's physics shape meant editing GDScript.

This PR introduces a `BallStateConfig` resource carrying those `@export` fields plus a single `apply(body: RigidBody2D)` helper. The `.tres` files under `resources/ball/states/` author the per-state values; `ball.gd` preloads them as `const` and each `enter_X` now opens with `<STATE>_CONFIG.apply(self)` followed by the imperative tail that cannot live in a resource: zeroing `linear_velocity`/`angular_velocity` on STORED and OUT_HELD, toggling `_suppress_miss_detection`, the OUT_REST speed reset and `speed_changed`/`at_max_speed_changed` emits, and `enter_play`'s branch on `court_config.friendship_bound_y` for the gravity_scale toggle between NORMAL and ARC.

One quirk worth flagging: OUT_REST's damping is overridden imperatively to `court_config.rest_roll_damping` immediately after `apply()`. Damping is a court-tunable, not a ball-state-tunable, so it has to be sourced from the live court config at entry time rather than baked into a static resource.

`_enter_arc` and `_enter_normal`, the per-frame NORMAL/ARC transition helpers in `_update_play_state`, keep their direct `gravity_scale = 1.0` / `gravity_scale = 0.0` writes. The only property that differs between NORMAL and ARC today is gravity, and a hot per-frame path does not want a multi-field resource apply. An inline comment in `ball.gd` warns that the direct writes silently go stale if a second property ever starts differing.

693 GUT tests stay green. Public `enter_X` signatures preserved.


J-Melon (commented):
## Strunk — gdscript conventions

Verdict: **ship** (posting as comment; GitHub blocks self-approve).

Walked the diff against the Volley convention checklist (export-over-onready, session-new `ClassName.new()` hazard, typed signatures, autoload usage, signal wiring, naming, Resource-over-loose-exports).

No findings clear the "what breaks if this stays?" bar.

- New `BallStateConfig` is only consumed via preloaded `.tres` constants in `ball.gd`; never instantiated as `BallStateConfig.new()`, so the async class-cache hazard for session-new scripts does not apply here.
- `apply(body: RigidBody2D) -> void` types `body` as `RigidBody2D` rather than `Ball` — correct direction, data doesn't depend on its consumer.
- All six `@export` fields typed; const config slots typed with `BallStateConfig`.
- The `_enter_arc` / `_enter_normal` single-property staleness risk and the `linear_damp` override-after-apply ordering for OUT_REST are both already named in code comments / PR body. No silent gotcha left for the next reader.
- This is the canonical shape from the loose-exports-vs-Resource rule — six clustered tunables into a Resource subclass with one `.tres` per state, not loose `@export` vars on Ball.

Conventions: clear.

J-Melon (commented):
## Vermeer — scene/resource review

**Verdict: ship-with-fixes** (apply `zaphod-blocked`).

One UID-stability issue, one dead-code observation. Autoload order untouched, no `.tscn` churn, no copy-paste sprawl, the new `BallStateConfig` script correctly omits `@tool`. PhysicsMaterial `ext_resource` UIDs (`uid://bldc0c7o5o2o4` for play.tres, `uid://b0volleyrestmt` for rest.tres) match the on-disk anchors.

### Findings

**1. New `.tres` files have no UID anchors (blocker)**

All five state bundles open with `[gd_resource type="Resource" script_class="BallStateConfig" load_steps=N format=3]` — no `uid="uid://..."` attribute, and no `.tres.uid` sidecars are committed. Godot will mint a UID on first import in every clone, so the UIDs diverge per-machine.

Why this bites: any future `uid://` reference (editor drag-drop into a scene, autocomplete completion, inspector pointer) will resolve correctly only on the machine that minted it; on other machines the UID won't exist until reimport, and references created elsewhere break. Path-based `preload("res://...")` in `ball.gd` is currently fine, which is why this is silent today.

Fix: open each `.tres` in the editor once and let it write the `uid=` attribute + sidecar, then commit the sidecars. Five files, ten lines of churn.

**2. `PLAY_MATERIAL` / `REST_MATERIAL` constants are now dead**

`scripts/entities/ball/ball.gd` lines 18-19 still `preload` `play.tres` and `rest.tres`, but the bundles now carry the material via `physics_material_override`. The old constants have no remaining call site in the diff. Leaving them keeps two preloads alive that misrepresent where the canonical material reference lives — exactly the staleness mode the inline comment in `_enter_arc` warns about. Drop them.

---

Once UIDs are anchored and the dead constants are removed, re-fire reviewers on the new diff and flip back to `zaphod-approved`.

J-Melon (commented):
Tests behaviour, not implementation. Coverage claim verified.

`tests/unit/ball/test_ball_state_transitions.gd` locks the full six-property bundle for each `enter_X`:

- STORED / OUT_HELD: `freeze=true`, `collision_layer=0`, `collision_mask=0`, plus `linear_velocity == Vector2.ZERO` (catches the imperative tail).
- PLAY_NORMAL / PLAY_ARC: `freeze`, `gravity_scale`, `linear_damp`, `physics_material_override == PLAY_MATERIAL`, `collision_layer/mask == 1`, branch-by-`global_position.y`.
- OUT_REST: every bundle property plus `linear_damp == REST_DAMPING (1.5)` — the load-bearing assertion that catches the post-`apply()` court-damp override. If a future edit drops the override line and lets the `.tres` default of `0.0` win, this test fails. Good.

Once-only emit + idempotent guards present per state, miss-suppression flag asserted on OUT_HELD, HELD→PLAY collision restore + miss re-engagement covered.

`test_ball_apex_arc.gd` exercises NORMAL→ARC and ARC→NORMAL through `_physics_process` and asserts on `gravity_scale` directly (1.0 / 0.0), not just `play_state`. The step-8 decision to keep `_enter_arc` / `_enter_normal` as imperative single-property writes does not open a coverage gap.

No isolated `BallStateConfig.apply()` test owed: bundle assertions in `test_ball_state_transitions` exercise `apply()` end-to-end against the canonical `.tres` files, which catches both apply() bugs and `.tres` value drift in one go. A stub-resource unit test would be strictly weaker.

CI: `Tests` passes on `bbb8903`. No coverage artifact pulled — the per-bundle assertions on the six properties on every `enter_X` make the changed-line coverage question trivially answered without it.

**Verdict: ship**

scripts/entities/ball/ball_state_config.gd
```diff
@@ -0,0 +1,21 @@
+## Declarative bundle of physics+collision properties for a single Ball.PlayState.
+## State entry methods call apply(self) and keep ordering-sensitive steps (velocity zero,
+## suppression flag, signal emit) in the imperative tail.
```
J-Melon: 3-line `##` docstring violates CLAUDE.md ("docstrings longer than one line: disallowed"). Project pattern in `ball.gd` is single-line `##` docs. Collapse, e.g.: `## Declarative physics+collision bundle for one Ball.PlayState; apply(body) writes the six fields.`

scripts/entities/ball/ball.gd
```diff
@@ -18,6 +18,14 @@ enum PlayState {
 const PLAY_MATERIAL: PhysicsMaterial = preload("res://resources/ball/play.tres")
 const REST_MATERIAL: PhysicsMaterial = preload("res://resources/ball/rest.tres")
 
+# Per-state physics+collision bundles. apply() writes six properties; ordering-sensitive steps
+# (velocity zero, suppression flag, speed reset, signal emit) stay imperative around the call.
```
J-Melon: Two-line block comment. Single-line per project rule. Suggest: `# Physics+collision bundles applied at state entry; imperative tail handles ordering-sensitive steps.`

scripts/entities/ball/ball.gd
```diff
@@ -93,11 +101,15 @@ func _update_play_state(delta: float) -> void:
 
 func _enter_arc() -> void:
 	_relock.enter_arc(speed)
+	# Hot per-frame path: direct write of the only NORMAL/ARC delta. Canonical bundle lives in
+	# res://resources/ball/states/play_arc.tres — if a second property starts differing between
+	# NORMAL and ARC, this single-property write goes stale silently.
```
J-Melon: Three-line comment violates the one-line rule. Keep the pointer (it's load-bearing) but compress, e.g.: `# Hot path: gravity-only delta vs play_arc.tres canonical bundle; stale if a second property diverges.`

J-Melon (commented):
Verdict: ship-with-fixes.

The refactor is clean: public `enter_X` signatures preserved, the imperative tails (velocity zero, suppression flag, speed/emit, court-config damp override) stay around `apply()` as the plan called for, and `PLAY_MATERIAL`/`REST_MATERIAL` constants are kept because the tests pin against them as canonical instances. Resource uids resolve to the same shared `PhysicsMaterial` so the equality assertions hold.

Two findings, both about comment policy from CLAUDE.md ("multi-line comments, docstrings longer than one line: disallowed"). Project pattern in `ball.gd` is single-line `##` docs (see `item_key`, `grab_area`, `court_config`).

1. `scripts/entities/ball/ball_state_config.gd` lines 1-3: 3-line `##` docstring. Collapse to one line. What breaks if it stays: violates the same rule that bars the other multi-line comments below, and sets a precedent for future Resource scripts.

2. `scripts/entities/ball/ball.gd` lines 21-22 (`STORED_CONFIG` block header) and lines 104-106 (`_enter_arc` staleness warning): multi-line block comments. The staleness warning in `_enter_arc` is load-bearing — keep the pointer but compress to one line, e.g. `# Hot path: gravity-only delta. Canonical bundle: play_arc.tres — stale if a second property diverges.` What breaks: same rule violation, and a future contributor copying this style spreads multi-line comments.

Out of scope but worth a thought for a follow-up (not blocking): `out_rest.tres` sets `linear_damp = 0.0` only for it to be overwritten by `court_config.rest_roll_damping` two lines later in `enter_out_rest`. A reader who edits `out_rest.tres` to set damp will see zero effect. The inline comment names it, but the silent-override shape is a foot-gun. Option for later: drop `linear_damp` from `BallStateConfig` entirely (court owns it), or sentinel-flag (NAN) and skip in `apply()`.

J-Melon (commented):
## Test-coverage re-review (second pass on 535eb83)

**Verdict:** approve (no findings)

Cherry-pick of `a428b2b` from PR #651 is byte-identical to the original diff: the same six-line `assert_almost_eq` block on `reinstated.linear_velocity.length()` is removed, and the behavioural assertion `assert_eq(reinstated.speed, 600.0, ...)` remains immediately above it. The AC (preserved speed propagates to the reinstated ball) is still covered; only the synchronous physics-write timing probe is gone.

**CI cross-check:** `Tests` is `pass` on the head commit (run 25753512530). No new coverage gap.

**Findings:** none.

Cherry-pick duplication self-resolves at squash time once PR #651 merges (acknowledged out-of-band).

J-Melon (commented):
## Kandinsky — scene/resource re-review

**Verdict: approve** (apply `zaphod-approved`).

Vermeer's single blocker — missing UID anchors on the five new `resources/ball/states/*.tres` files — is resolved on `86d8825`.

Verified on `origin/feat/sh-380-ball-state-config`:

- All five `.tres` headers now carry an in-header `uid="uid://..."`:
  - `out_held.tres` → `uid://bballstateoutld`
  - `out_rest.tres` → `uid://bballstateoutrs`
  - `play_arc.tres` → `uid://bballstateplarc`
  - `play_normal.tres` → `uid://bballstateplnrm`
  - `stored.tres` → `uid://bballstatestor0`
- All five are distinct from each other.
- Repo-wide grep for each UID returns exactly one hit (the file itself) — no collisions with any other `.tres`/`.tscn`/`.gd`/`.uid` in the tree.
- No `.tres.uid` sidecars were committed, which is correct for hand-crafted in-header UIDs; Godot will generate the sidecars on first import in each contributor's editor without disturbing the anchor.

Hand-crafted UIDs are valid here — the `b` prefix + 16 lowercase-alphanumeric chars matches Godot's UID alphabet and length, so the editor will accept them rather than re-issue. If any of these `.tres` files get referenced by a scene later, the `uid://` link will stay stable across renames.

Nothing else in scope from the new commits since the first pass.

J-Melon (commented):
Second pass — Pythagoras.

Verifying the three comment-policy items from Vitruvius's first pass:

- `scripts/entities/ball/ball_state_config.gd:1` — multi-line `##` docstring is now a single `##` one-liner above `class_name`. Resolved.
- `scripts/entities/ball/ball.gd:21` — the `#` block above the per-state `const` cluster is now a single line that still names the ordering-sensitive constraint. Resolved.
- `scripts/entities/ball/ball.gd:103` — staleness warning is now one compressed line naming the canonical bundle (`play_arc.tres`) and the failure mode ("second diverging property goes stale silently"). `_enter_normal` at line 109 cross-references it so the warning still covers both sides of the hot path. Content preserved. Resolved.

New commit on `scripts/core/court_config.gd:11` — `# TODO(SH-394): move into a SurfaceConfig that also owns the floor PhysicsMaterial.` is one line, ticket reference matches the open SurfaceConfig ticket, and it sits directly above the `@export var rest_roll_damping` declaration. Resolved.

No regressions on earlier-addressed items. No new findings.

Verdict: ship.

scripts/entities/ball/ball.gd
```diff
@@ -18,6 +18,10 @@ enum PlayState {
 const PLAY_MATERIAL: PhysicsMaterial = preload("res://resources/ball/play.tres")
 const REST_MATERIAL: PhysicsMaterial = preload("res://resources/ball/rest.tres")
```
J-Melon: **Esker**

dead code: `PLAY_MATERIAL` (line 18) and `REST_MATERIAL` (line 19) are no longer referenced anywhere in this file after the refactor; the play / out_rest configs carry the material override. Drop both consts.

tests/unit/items/test_ball_drag_controller.gd
```diff
@@ -366,12 +366,6 @@ func test_grab_and_release_preserves_live_ball_speed_through_to_reinstated_ball(
 	var reinstated: Ball = _reconciler.get_ball_for_key("ball_alpha")
 	assert_not_null(reinstated, "release over court should reinstate a ball")
 	assert_eq(reinstated.speed, 600.0, "preserved_speed should propagate to the reinstated ball")
```
J-Melon, line 368: **Esker**

scope creep / regression coverage drop: this PR removes the `linear_velocity.length() == 600.0` assertion from a drag-controller test that is otherwise unrelated to the BallStateConfig refactor. The remaining `assert_eq(reinstated.speed, 600.0, ...)` checks the scalar property; the deleted assertion was the only check that the actual physics velocity magnitude tracked the preserved speed. If the assertion now fails because `PLAY_ACTIVE_CONFIG.apply()` reorders gravity / material writes against `_apply_speed()`, that's a real behaviour change worth keeping under test rather than silencing. If it's truly unrelated to this PR, lift it onto its own ticket.

scripts/entities/ball/ball.gd
J-Melon: **Gru**

Not dead: `Ball.PLAY_MATERIAL` and `Ball.REST_MATERIAL` are test anchors. `tests/unit/ball/test_ball_state_transitions.gd:66,91,101,119` assert `physics_material_override == Ball.PLAY_MATERIAL` / `Ball.REST_MATERIAL` directly. Vermeer flagged the same shape earlier and Josh kept them (`#issuecomment-4433360514`); decision stands.

tests/unit/items/test_ball_drag_controller.gd
J-Melon: **Gru**

Intentional drop in `535eb83` ("test: drop synchronous physics-write assertion in drag-release test"). The assertion was probing implementation timing CI could not honour; the remaining `reinstated.speed == 600.0` covers the named invariant (live-grab preserves ball speed). Velocity magnitude tracking the scalar speed is downstream of `_apply_speed()`, not a separate behaviour worth pinning here.

PR #653, opened 2026-05-12
refactor!: decentralise save into per-system slices; wipes existing saves

J-Melon:
ProgressionData (flat 13-field RefCounted bag) is replaced by five typed slice classes owned by their respective autoloads:

- `ProgressionManager.economy: EconomyState`
- `ProgressionManager.records: RecordsState`
- `ProgressionManager.unlocks: UnlocksState`
- `ProgressionManager.partners: PartnersState`
- `ItemManager.state: ItemWorldState`

SaveManager loses its `_progression` field and becomes a coordinator: registers the five slices, walks them to assemble `{"economy": {...}, "items": {...}, "records": {...}, "unlocks": {...}, "partners": {...}}` on save, splits the loaded dict and dispatches `apply_save_dict(slice)` on load. Each slice owns its own `to_save_dict` / `apply_save_dict` plus the typed-coercion helpers it needs (Vector2 nesting, typed-dict coercion, StringName arrays).

`loose_in_venue` collapses from two duplicate stores (`ItemManager._loose_in_venue: Dictionary[String, bool]` runtime + `ProgressionData.loose_in_venue: Array[String]` persisted, never reconciled) into one `ItemWorldState.loose_in_venue: Dictionary[String, Vector2]`. Membership is the loose-state flag; value is the respawn position. `item_positions` renames to `ball_positions` to match what BallReconciler actually populates. `ball_play_states` is persisted alongside positions so an OUT_REST ball does not re-enter court flow on load.

`adopt_pre_existing_balls` now snaps STORED scene-authored balls to their rack slot and restores OUT_REST / PLAY positions and play states from the saved slice. OUT_HELD demotes to OUT_REST on load because the drag-controller context is gone. Velocity is not persisted yet; OUT_REST restore zeroes velocity to prevent a load-frame hop.

**Wipes existing saves.** The on-disk shape changed materially: flat dict to nested dict-of-slices, `item_positions` rename, `loose_in_venue` retype from Array to Dictionary, new `ball_play_states` field. No migration code, no compat shims; prototype rule is "schema changes wipe; v1 introduces the migrator." SH-400 tracks the v1 migrator activation; `designs/01-prototype/tech/03-save-versioning.md` is the spec.

ProgressionData and its `.uid` are deleted. 206 `_progression.X` call sites migrate to the appropriate owner. Slice instances mutate in place across load so consumers caching refs (Court, ProgressionManager, ItemManager, BallReconciler) stay valid. 697 GUT tests pass.

Closes SH-381.


J-Melon (commented):
## Fowler — GDScript conventions

**Verdict:** ship (clean on conventions)

### Checks passed

- `SaveManager.set_storage(storage: SaveStorage) -> void` — typed seam, no `Variant`/`Object` leak.
- `load_from_disk() -> bool` and `_write_to_disk() -> bool` — both typed; the fallback iterator is `for content: Variant in fallbacks` with `is String` narrowing, which is correct for JSON-derived content.
- `ProgressionData` has no explicit `_init` after the purification — genuinely arg-free, no orphan `storage` default lurking.
- `_init` remains only on `SaveManager` (`autosave_interval: float = 10.0`), typed.
- Tests instantiate the freshly-edited `SaveManager` via `load("res://scripts/progression/save_manager.gd").new(0.05)` (test_save_manager.gd, test_file_save_storage.gd), respecting the session class-cache rule.
- No `has_method` guards, no untyped `Node` accessors, no new `@onready`, no autoload churn.
- `_apply_loaded_content` is typed end-to-end.
- No abbreviated names; `_position_provider`, `_autosave_timer`, `_write_blocked` all spelled out.

Nothing to fix on conventions. Suggest `zaphod-approved` from this reviewer.

J-Melon (commented):
zaphod-approved (save-format warden).

Save format preserved verbatim. `to_dict`/`from_dict` shape unchanged (same 13 keys, same types, same defaults). `_apply_loaded_content` on SaveManager copies all 13 persisted fields onto the held `_progression`, matching the serialisation contract — no field silently dropped. Rolling-backup fallback chain relocates unchanged: primary `_storage.read()` then iterate `_storage.read_fallbacks()` through the same parse-and-apply helper. `FileSaveStorage` untouched, so on-disk paths and rotation are identical; pre-PR save files load cleanly through the new SaveManager path. No new `@export` on persisted state, no compat shims, no autoload reorder. PR body correctly scopes the `!` to the GDScript API surface and affirms byte-for-byte save compatibility.

J-Melon (commented):
## Code-quality review (Bauhaus)

Verdict: **ship-it** (cannot self-approve via API; flagging via comment + label).

Scope matches the plan: `save_to_disk`/`load_from_disk`/`_try_load_content`/`_storage` deleted from `ProgressionData`, relocated onto `SaveManager` as `load_from_disk`/`_write_to_disk`/`_apply_loaded_content`/`_storage`. `FileSaveStorage` untouched. `_init(storage)` argument is gone from `ProgressionData` (implementer dropped the now-empty `_init` entirely — cleaner than the plan's "arg-free" wording, behaviour-equivalent). 22 test files migrated, both Shape A (drop the no-op double) and Shape B (`set_storage` on SaveManager, or `from_dict(JSON.parse_string(...))` for pure-data round-trips).

### Verified

- `_apply_loaded_content` mutates the held `_progression` in place via per-field copy — the cached-refs hazard (court / progression_manager / item_manager / ball_reconciler) is correctly avoided. All 13 fields copied, matching `to_dict`/`from_dict`.
- `load_from_disk()` rename is consistent across SaveManager, both save tests, and the design-doc reference flipped to `SaveManager.save()`. No orphan references to old `ProgressionData.save_to_disk`/`load_from_disk` anywhere on the branch.
- Naming choice of `load_from_disk()` over the planned `load()` is defensible — `load` would shadow the global identifier on a `Node` autoload.

### Non-blocking observation (pre-existing smell, relocated not introduced)

`_apply_loaded_content` enumerates all 13 ProgressionData fields by name on the SaveManager side. Adding a new field to ProgressionData now requires touching SaveManager too, or the field silently drops on load. The pre-refactor code had the same shape (the field list lived on ProgressionData itself, arguably worse). A `ProgressionData.copy_from(other)` instance method would re-localise the field list; out of scope here, worth a follow-up ticket if a third "copy all fields" site appears.

### Out-of-scope but harmless

`tests/unit/items/test_sh366_battle_followups.gd.uid` is added — Godot import-cache artifact for an existing `.gd` from main. Unrelated to SH-381, no risk.

No code-quality blockers.

J-Melon (commented):
## Test-coverage review — Galton

**Verdict:** ship-with-followup. Migration is mechanically clean; one behavioural gap worth a small follow-up test rather than blocking the PR.

**CI:** `Tests` pass on `3c68b1c1`. Coverage on `progression_data.gd` is healthy across the surviving surface (all `to_dict`/`from_dict`/`clear` branches hit).

### Findings

1. `tests/unit/save/test_save_manager.gd` — no test for `SaveManager.load_from_disk()` on the happy path through the mocked storage. The only `load_from_disk` exercise in the suite is the backup-fallback case in `test_file_save_storage.gd:91-105`. The plan called for "round-trip via SaveManager.save/load_from_disk through a mocked storage" here; that round-trip never materialised. Consequence: a regression that breaks primary-path loading (e.g. `_apply_loaded_content` returning false on valid content) would only be caught by a test that also breaks the corrupt-primary fallback path. Recommend adding a `test_load_from_disk_applies_stored_blob` that stubs `_mock_storage.read` to return a stringified `to_dict()` and asserts the held progression fields update.

2. `scripts/progression/save_manager.gd:46-47` — the in-place-mutation contract ("Mutates the held `_progression` in place so cached refs across the project stay valid") is the load-side raison d'être of this refactor and has no direct assertion. A future change that does `_progression = loaded` instead of field-by-field copy would pass every existing test while silently breaking court / item_manager / ball_reconciler cached references. Recommend a test that grabs `var ref := _save_manager.get_progression_data()`, calls `load_from_disk()` with non-default stubbed content, and asserts `ref.friendship_point_balance` reflects the loaded value (same identity, not replacement).

3. `tests/unit/save/test_file_save_storage.gd:90-93` — old comment claimed the fallback test used "structurally-parseable-but-wrong-schema" content; the actual mechanism was always empty primary. New comment is now accurate. No action; noting because the comment update inside this diff is correct.

### Shape-A spot check

Sampled `test_ball.gd`, `test_friendship_points.gd`, `test_paddle_upgrades.gd`, `test_streak_buildup.gd`, `test_real_input_drag_paths.gd`, `tests/stubs/item_factory.gd`: all `_mock_storage` declarations, stubs, and constructor args removed cleanly; no orphan variables left. Grep across `tests/` confirms only `test_save_manager.gd` and `test_file_save_storage.gd` retain storage doubles, which is correct since they're the SaveManager/FileSaveStorage suites.

### Shape-B confirmation

- `test_progression_data.gd`: storage-coupled round-trip tests cut; `test_cleared_data_dict_round_trip_keeps_partner_fields_empty` preserves the partner-fields-empty serialisation invariant via direct `to_dict`/`from_dict`. Clean cut.
- `test_progression_data_positions.gd`: JSON round-trips collapsed to `from_dict(JSON.parse_string(...))`; same invariant exercised.
- `test_placement_drives_effects.gd`, `test_ball_regime_transitions.gd`: re-hydration tests now bypass storage entirely (correct — these test ItemManager re-hydration, not the storage seam). Same behavioural assertions retained.

Finding 1 is the only one worth a follow-up commit; findings 2 is a defensive ask. Neither blocks merge.

J-Melon (commented):
Re-review of `b84b323` — two added tests for `SaveManager.load_from_disk` happy-path and instance-identity invariant.

**Verified**

- `test_load_from_disk_applies_stored_blob` stubs `_mock_storage.read` (correctly wired via `set_storage` in `before_each`) and asserts `friendship_point_balance == 42` / `active_partner == &"martha"` against defaults of `0` / `&""`. Non-no-op; the load path runs end-to-end through `_apply_loaded_content`.
- `test_load_preserves_progression_instance_identity` captures `held_ref` before load and asserts `assert_eq(held_ref, _save_manager._progression)`. `ProgressionData extends RefCounted` with no `_eq` override, so GDScript `==` here is reference identity, not field equality — this is the load-bearing assertion. The follow-up field assertions on `held_ref` then prove mutation landed on the held instance rather than a copy.
- Together they pin the invariant called out in the first-pass review: a future `_progression = loaded` regression would fail the identity assertion and stale the four cached refs (court / item_manager / ball_reconciler / progression_manager).

**CI**: Tests pass on b84b323.

Verdict: approve (ship).

scripts/progression/save_manager.gd
```diff
@@ -27,12 +30,59 @@ func _ready() -> void:
...
 func save() -> void:
 	if _write_blocked:
 		return
 	_capture_live_positions()
-	_progression.save_to_disk()
+	_write_to_disk()
+
+
+## Loads from storage, falling back to rolling backups if primary fails to parse.
+## Mutates the held _progression in place so cached refs across the project
```
J-Melon: suggestion (style): comment too big

scripts/progression/save_manager.gd
```diff
@@ -27,12 +45,64 @@ func _ready() -> void:
...
+func set_storage(storage: SaveStorage) -> void:
+	_storage = storage
+
+
+# todo: SH-400 stamp schema_version and run the migration chain on load at v1.
 ## Saves game. No-op while writes are blocked by a pending clear.
 func save() -> void:
 	if _write_blocked:
 		return
 	_capture_live_positions()
```
J-Melon, line 59: **Soos**

save-format: This PR replaces the flat 13-field save dict with a 5-slice nested dict, renames `item_positions` to `ball_positions`, and retypes `loose_in_venue` from `Array[String]` to `Dictionary[String, Vector2]`. Existing on-disk saves will not load. The PR body currently says "Compatibility with on-disk save files is preserved byte-for-byte", which is the opposite of the truth and directly contradicts the no-compat policy this PR is implementing. Update the body to say "wipes saves" in plain language (the `refactor!` bang and tech doc carry the rest).

scripts/items/item_manager.gd
```diff
@@ -121,22 +122,23 @@ func get_placement(item_key: String) -> int:
...
 func is_loose_in_venue(item_key: String) -> bool:
-	return _loose_in_venue.get(item_key, false)
+	return items_world != null and items_world.loose_in_venue.has(item_key)
 
 
-## Marks an owned item as loose-in-venue. Idempotent. Emits item_placement_changed.
-func mark_loose_in_venue(item_key: String) -> void:
-	if _loose_in_venue.get(item_key, false):
+## Marks an owned item as loose-in-venue at `position`. Idempotent. Emits item_placement_changed.
+func mark_loose_in_venue(item_key: String, position: Vector2 = Vector2.ZERO) -> void:
```
J-Melon, line 131: **Dipper**

nit: `position: Vector2 = Vector2.ZERO` default silently respawns loose bodies at (0,0) if a future caller forgets the argument. Both current call sites pass real positions; dropping the default (required arg) catches the drift at edit time and matches the field semantics (membership = loose, value = respawn position). Not blocking.

scripts/progression/partners_state.gd
```diff
@@ -0,0 +1,44 @@
...
+func apply_save_dict(data: Dictionary) -> void:
+	recruit_offered_partners = _to_typed_string_name_array(data.get("recruit_offered_partners", []))
+	unlocked_partners = _to_typed_string_name_array(data.get("unlocked_partners", []))
+	active_partner = StringName(data.get("active_partner", ""))
+	partner_volley_totals = _to_typed_string_name_dict(data.get("partner_volley_totals", {}))
+
+
+static func _to_typed_string_name_array(raw: Array) -> Array[StringName]:
+	var typed: Array[StringName] = []
+	for value in raw:
```
J-Melon: **Dipper**

style: untyped loop iterators (`for value in raw` here and `for key in raw` below) diverge from the sibling slices, which use typed iterators (`for key: String in raw`). Typed loops let the analyser flag bad input rather than coercing it silently. `raw: Array` / `raw: Dictionary` permit Variant entries, so `for value: Variant in raw` is the honest form.

scripts/items/item_manager.gd
```diff
@@ -19,15 +19,16 @@ var items: Array[ItemDefinition] = [
 	preload("res://resources/items/wrist_brace.tres"),
 ]
 
-var _progression: ProgressionData
+var items_world: ItemWorldState
```
J-Melon: **Dipper**

encapsulation: `items_world` and `economy` are public here (no underscore), where the prior `_progression` was private. Anything in the project can now read/write `ItemManager.items_world.item_levels` directly, bypassing `purchase` / `adopt_authored` and their signal emits. Suggest prefixing both with `_` so slice mutation stays inside `ItemManager`; reads already go through `get_level` / `get_friendship_point_balance`.

J-Melon (commented):
changes needed

scripts/progression/save_manager.gd
J-Melon: **Gru**

PR body rewritten to lead with "wipes existing saves" and describe the new shape (flat dict → 5-slice nested, ball_positions rename, loose_in_venue retype). Stale claim of byte-for-byte compat predated Hermes's slice scope.

scripts/items/item_manager.gd
J-Melon: **Gru**

Keeping the default. Every existing call site passes a real position (ball_drag_controller, shop_item, register_loose_body all verified). Mabel reached the same conclusion: speculative until a caller drops the arg. Worth a separate ticket if it becomes a recurring slip.

scripts/progression/partners_state.gd
J-Melon: **Gru**

Typed iterators landed in e80980e: `for value: Variant in raw` / `for key: Variant in raw`. Raw values come from JSON so the typed annotation is Variant, but the iterator shape now matches the sibling slices.

scripts/items/item_manager.gd
J-Melon: **Gru**

Intentional. The slice IS the public API of the autoload; gating behind getters duplicates its own surface. Direct mutation through `ItemManager.state.X` is the shape we want. Also renamed `items_world` → `state` on ItemManager and `items` on SaveManager in 0202026 since the field name was reading doubly redundant.

scripts/items/ball_reconciler.gd
```diff
@@ -291,14 +312,53 @@ func _default_spawn_position() -> Vector2:
...
+func _apply_saved_play_state(ball: Ball, play_state: int) -> void:
+	match play_state:
+		Ball.PlayState.OUT_REST:
+			ball.enter_out_rest()
+			# Velocity is not persisted yet; without zeroing, the body keeps the integration-frame
+			# momentum it picked up between scene spawn and adopt, and visibly hops on load.
+			ball.linear_velocity = Vector2.ZERO
+			ball.angular_velocity = 0.0
+		Ball.PlayState.OUT_HELD:
+			ball.enter_out_held()
```
J-Melon: **Hal**

issue: restoring OUT_HELD without re-establishing the drag controller leaves the ball frozen mid-air with miss-detection suppressed and no exit path; demote to OUT_REST or PLAY on load.

J-Melon: **Gru**

Fixed in 3b65c11: OUT_HELD now demotes to OUT_REST on load via the same `enter_out_rest()` path. Same case-arm in the match block, so OUT_HELD restore is structurally identical to OUT_REST. Drag context is gone on load; HELD has no exit path without one.

scripts/items/ball_reconciler.gd
```diff
@@ -291,14 +312,53 @@ func _default_spawn_position() -> Vector2:
...
+		return
+
+	if saved.ball_positions.has(item_key):
+		ball.global_position = saved.ball_positions[item_key]
+
+	if saved.ball_play_states.has(item_key):
+		_apply_saved_play_state(ball, saved.ball_play_states[item_key])
+
+
+func _apply_saved_play_state(ball: Ball, play_state: int) -> void:
```
J-Melon, line 338: **Sunny**

missing-test: `_apply_saved_play_state` has zero coverage. The OUT_REST restore, OUT_HELD demote (sharing the OUT_REST arm), and PLAY_NORMAL/PLAY_ARC restore branches are all new player-visible reload behaviour (a ball that was at rest mid-court reloads in PLAY_NORMAL and re-enters play, etc.) with no test asserting it. Coverage report on 3b65c11 shows lines 324-351 never execute in any test. Add an `adopt_pre_existing_balls` test that primes `SaveManager.items.ball_play_states[key] = Ball.PlayState.OUT_REST` (or seeds via a stub) plus a position, runs adopt, and asserts `ball.play_state == OUT_REST` and `ball.global_position == saved`. A second case for OUT_HELD asserting the demotion to OUT_REST covers the case-arm sharing. Without either, the case-arm semantics can flip and tests stay green.

scripts/progression/item_world_state.gd
```diff
@@ -0,0 +1,83 @@
...
+var item_levels: Dictionary[String, int] = {}
+var item_placements: Dictionary[String, int] = {}
+
+## Per-ball world position; every non-STORED ball reloads where it was, across
+## OUT_REST, OUT_HELD, PLAY_NORMAL, and PLAY_ARC. STORED balls reconstruct from
+## rack_slot_index_by_key instead.
+var ball_positions: Dictionary[String, Vector2] = {}
+
+## Per-ball PlayState enum int. Restores the play regime so an OUT_REST ball does
+## not re-enter court flow as PLAY_NORMAL by default on load.
```
J-Melon: **Sunny**

missing-test: `ball_play_states` has no round-trip assertion. `test_progression_data_positions.gd` has `test_ball_positions_round_trip_via_dict` and a JSON-string round-trip; the new field needs the same shape. A renamed/retyped key in `to_save_dict` would silently slip through. Add `test_ball_play_states_round_trip_via_dict` populating one entry, applying `to_save_dict`/`apply_save_dict`, and asserting the int comes back; plus `test_clear_resets_play_states` for the clear() addition on line 29.

scripts/progression/save_manager.gd
```diff
@@ -27,32 +46,99 @@ func _ready() -> void:
...
+
+## Registers a callable that returns a Dictionary[String, Vector2] of live ball positions.
 func set_position_provider(provider: Callable) -> void:
 	_position_provider = provider
 
 
+## Registers a callable that returns a Dictionary[String, int] of live ball PlayState enum ints.
+func set_play_state_provider(provider: Callable) -> void:
+	_play_state_provider = provider
+
```
J-Melon, line 117: **Sunny**

missing-test: `set_play_state_provider` capture path is untested. `test_save_manager.gd` covers `set_position_provider` (test on line 99) but the mirrored play-state provider has no equivalent. The capture stamps `items.ball_play_states` from the provider on every save; if the str(key)/int(value) coercion regresses, no test fails. Add `test_save_captures_play_states_from_registered_provider` mirroring the positions test: register a `Dictionary[String, int]` provider, call save, assert `items.ball_play_states[key] == enum_int`. Optional companion: test_save_without_play_state_provider_leaves_play_states_untouched.

scripts/items/ball_reconciler.gd
J-Melon: **Gru**

Covered in 3d4739a: new `tests/unit/items/test_ball_reconciler_adopt_restore.gd` exercises the OUT_REST restore (position + state + zeroed velocity), the OUT_HELD demotion to OUT_REST, and the PLAY_NORMAL/PLAY_ARC branch.

scripts/progression/item_world_state.gd
J-Melon: **Gru**

Covered in 3d4739a: `test_ball_play_states_round_trips` in `test_progression_data_positions.gd` asserts OUT_REST and PLAY_ARC ints round-trip through to_save_dict/apply_save_dict.

scripts/progression/save_manager.gd
J-Melon: **Gru**

Covered in 3d4739a: `test_save_captures_play_states_from_registered_provider` (mirror of the positions test) plus `test_save_without_play_state_provider_leaves_states_untouched` in `test_save_manager.gd`.

scripts/progression/save_manager.gd
```diff
@@ -27,32 +46,99 @@ func _ready() -> void:
...
+	if _position_provider.is_valid():
+		var live: Variant = _position_provider.call()
+
+		if live is Dictionary:
+			var typed: Dictionary[String, Vector2] = {}
+
+			for key: Variant in live:
+				var value: Variant = live[key]
+
+				if value is Vector2:
```
J-Melon: suggestion: lots of nesting and ifs, refactor into a nicer pattern

scripts/progression/save_manager.gd
```diff
@@ -80,6 +167,22 @@ func _notification(what: int) -> void:
...
 
-## Returns the currently stored [ProgressionData]
-func get_progression_data() -> ProgressionData:
-	return _progression
+func _ensure_slices() -> void:
+	if economy == null:
+		economy = _SLICE_SCRIPTS["economy"].new()
+	if items == null:
+		items = _SLICE_SCRIPTS["items"].new()
+	if records == null:
```
J-Melon, line 186: suggestion: if spacing

scripts/progression/save_manager.gd
J-Melon: **Gru**

Flattened in c0acfc7: split into `_capture_positions` and `_capture_play_states` per-provider helpers, each using early-return guards instead of nested `if`. The two paths no longer share scope so the nesting is one level deep per helper.

scripts/progression/save_manager.gd
J-Melon: **Gru**

Blank lines added before each `if X == null:` in `_ensure_slices` in c0acfc7, per the blank-line-before-if rule.

PR #675, opened 2026-05-16
feat: equip flow with kit cap

### LINKED ISSUE #674: Equip flow with kit cap
As a player,  <br>I want to drag gear onto my character.  <br>So that I can equip effects caused by the items I buy.

**Acceptance Criteria:**

- [X] During a timeout, dragging gear from the rack onto the character equips it; the effect activates.
- [X] Dragging an equipped item back onto the rack unequips it; the effect stops.
- [X] When the kit cap is reached, dropping more gear on the character is rejected; the held item stays on the cursor.
- [ ] Equipping is rejected outside the timeout window.

J-Melon:
Closes #674.

Wires up the first real equip gesture: drag equipment from the rack onto the main character during a timeout to register its effects, drag back to the rack to unequip, capacity is gated by the kit_slots stat. ItemManager grows equip/unequip wrappers around activate/deactivate (ball-role callers untouched) plus get_kit_remaining and an equip_refused signal so future animation listeners have a single source for capacity rejection.

Visual reparenting of the equipped art onto each item's anchor_node_path is deferred. The anchors and field are authored so a follow-up can land it without re-touching the API or scene wiring. Cadence (whistle) intentionally has no anchor and falls through to the character root.


scripts/core/timeout_controller.gd
```diff
@@ -87,25 +96,85 @@ func _cache_positions() -> void:
...
-	if _walk_tween != null and _walk_tween.is_valid():
-		_walk_tween.kill()
-	_walk_tween = create_tween()
-	_walk_tween.tween_property(main_character, "position", target, config.walk_duration_seconds)
-	_walk_tween.finished.connect(on_finished)
+func _begin_walk_back() -> void:
+	_start_horizontal_walk(_lane_x, _on_reached_lane, State.WALKING_ON)
+
+
+func _start_horizontal_walk(target_x: float, on_finished: Callable, walk_state: State) -> void:
```
J-Melon, line 127: issue: domain overlap, should this not be in the paddle or adjacent controller?

scripts/core/timeout_controller.gd
```diff
@@ -87,25 +96,85 @@ func _cache_positions() -> void:
...
+		return
+	main_character.velocity = Vector2(0.0, -config.descent_speed)
+	main_character.move_and_slide()
+
+
+func _step_horizontal_walk(delta: float) -> void:
+	var current_x: float = main_character.position.x
+	var remaining: float = _walk_target_x - current_x
+	var step: float = _walk_speed * delta
+	if _walk_speed <= 0.0 or absf(remaining) <= step:
```
J-Melon, line 184: suggestion: if spacing, and var spacing

scripts/items/drop_targets/character_drop_target.gd
```diff
@@ -0,0 +1,157 @@
...
+func _hydrate_equipped_visuals() -> void:
+	if _item_manager == null:
+		return
+	for key: String in _item_manager.state.item_placements.keys():
+		if int(_item_manager.state.item_placements[key]) != PlacementScript.EQUIPPED:
+			continue
+		if not _is_equipment_role(key):
+			continue
+		_mount_equipped_visual(key)
+
```
J-Melon, line 68: issue: there are ifs all over the place in this file, any possible refactor?

tests/integration/test_timeout_blocks_autoplay_drive.gd
```diff
@@ -0,0 +1,110 @@
...
+	floor_shape.size = Vector2(4000.0, 100.0)
+	floor_collision.shape = floor_shape
+	_floor.add_child(floor_collision)
+	add_child_autofree(_floor)
+
+	var config: TimeoutConfig = TimeoutConfig.new()
+	config.walk_duration_seconds = 0.5
+	config.equip_pose_offset_x = -200.0
+	config.descent_speed = 1200.0
+	_controller = load("res://scripts/core/timeout_controller.gd").new()
```
J-Melon, line 51: suggestion: for stuff like this, why are these config vars not in the constructor?

tests/unit/items/test_drop_targets.gd
```diff
@@ -137,6 +142,335 @@ func test_rack_drop_target_without_drop_area_rejects() -> void:
...
+	event.pressed = true
+	character_target._on_equipped_press_input(null, event, 0, "gear_press")
+	assert_signal_emit_count(character_target, "equipped_art_pressed", 1)
+
+
+func test_set_equipped_visual_visibility_toggles_the_mounted_art() -> void:
+	var manager: Node = ItemFactory.create_manager(self)
+	var equipment: ItemDefinition = _make_equipment_definition("gear_vis")
+	manager.items.assign([equipment] as Array[ItemDefinition])
+	manager.economy.friendship_point_balance = 10000
```
J-Melon: question: is all this behavioural, no impl detail, it seems like a lot

tests/unit/paddle/test_timeout_controller.gd
```diff
@@ -1,17 +1,25 @@
 # gdlint:ignore = max-public-methods
 extends GutTest
 
-## Drives the walk tween manually so phase boundaries land deterministically without real-time awaits.
+## Drives the controller's `_physics_process` directly with a chosen virtual delta;
+## a StaticBody2D floor in the fixture stands in for venue.tscn's VenueFloor.
```
J-Melon: issue: comment too big

scripts/core/timeout_controller.gd
J-Melon: **Gru**: fixed in aa7d572, blank line before the if.

tests/unit/paddle/test_timeout_controller.gd
J-Melon: **Gru**: trimmed to one line in aa7d572.

scripts/core/timeout_controller.gd
J-Melon: **Gru**: real point. `_start_horizontal_walk` and `_step_horizontal_walk` are paddle-movement primitives; TimeoutController is the only caller today but the surface area belongs on Paddle or a sibling `PaddleMover`. Suggest folding into a SH-405 follow-up rather than landing here. Your call: fold now or follow-up ticket?

scripts/items/drop_targets/character_drop_target.gd
J-Melon: **Gru**: short-circuit guards twice over (`is_equipment_role` in both `_hydrate_equipped_visuals` and `_on_item_placement_changed`). Smallest move: push the role check into `_mount_equipped_visual` itself so callers stop guarding. Want me to land it here or on the same follow-up as the timeout-controller split?

tests/integration/test_timeout_blocks_autoplay_drive.gd
J-Melon: **Gru**: Resource constructors in Godot are no-arg; `TimeoutConfig.new()` is what `Resource.new()` gives us. Options: (a) add a `TimeoutConfig.with(walk, offset, descent)` factory method on the class; (b) ship a test helper `_make_timeout_config(...)` in the fixtures. Cleaner long-term: (a). Fold into the follow-up?

tests/unit/items/test_drop_targets.gd
J-Melon: **Gru**: this one's testing the visibility-toggle method directly, which is impl-shaped. Player-observable equivalent: press-on-equipped hides the visual, drop-on-non-target reveals it. The earlier press-area tests cover the press; only the reveal-on-cancel needs a new behavioural case. Want me to rewrite the test or drop it as redundant?

scripts/core/timeout_controller.gd
J-Melon: **Gru**: filed SH-409 spike for general paddle movement controls; not folded here.

scripts/items/drop_targets/character_drop_target.gd
J-Melon: **Gru**: folded here in 1797e75; role check now lives inside `_mount_equipped_visual`, callers stopped guarding.

tests/integration/test_timeout_blocks_autoplay_drive.gd
J-Melon: **Gru**: factory method on TimeoutConfig is impl, only used by tests in prod (`.tres` deserialisation handles real config). A test-fixture helper (`_make_timeout_config(...)` in a shared spot) carries the same ergonomic win without adding API surface. Rolling into the broader test sweep just dispatched.

tests/unit/items/test_drop_targets.gd
J-Melon: **Gru**: dispatched Kel on a full test sweep across the SH-405-touched test files; behavioural rewrite, cut noise, keep essentials + edges + exceptional only. This one's in scope.

J-Melon: **Gru**: dropped in 094fd88. The hook itself has no coverage now, which is fine for unit tests; the player-observable behaviour (press-on-equipped → release-outside-rack → source visual visible again) belongs in a `ball_drag_controller` integration test. Kel flagged it as out-of-scope for this sweep. Two options for the gap: file a follow-up ticket, or add the integration test here before merge. Your call.

J-Melon: **Gru**: scratch my earlier integration-test framing. Per the just-saved rule, integration tests are reserved for full player-loop completions; cancel-back-reveal is a two-component visibility hand-off, not a loop. The press-area unit test on `CharacterDropTarget` and the drag controller's unit cancel test cover the contract; your playtest covers the visible behaviour. No follow-up needed.

scripts/entities/paddle.gd
```diff
@@ -20,6 +20,12 @@ var _paddle_speed: float = 0.0
 var _collision_shape: RectangleShape2D
 var _sprite_natural_height := 0.0
 
+## Set by TimeoutController during the walk; suppresses drive() so controllers don't fight the pose.
+var _drive_blocked: bool = false
```
J-Melon: **Hero** issue: `_drive_blocked` is written from `TimeoutController` (`main_character._drive_blocked = true/false`); the underscore prefix advertises private but the access is cross-object. Promote to `drive_blocked` with a setter, or expose `set_drive_blocked(bool)` so the contract is one-sided.

scripts/items/item_manager.gd
```diff
@@ -189,6 +191,35 @@ func deactivate(item_key: String) -> bool:
 	return true
 
 
+## Free equipment-kit slots: kit_slots floored minus persisted-EQUIPPED count; clamped to zero so
+## over-capacity loads (designer lowered kit_slots) report 0 rather than negative.
```
J-Melon: **Hero** issue: two-line `##` block on `get_kit_remaining`; `implementer-nits` caps doc comments at one line. Compress to a single sentence (the parenthetical about lowered kit_slots can go to the gear doc).

scripts/items/drop_targets/character_drop_target.gd
```diff
@@ -0,0 +1,154 @@
...
+# Sub-Area2D that emits equipped_art_pressed on left-click; layers off so it can't drive physics.
+func _attach_press_area(visual: Node, definition: ItemDefinition, item_key: String) -> void:
+	if definition.at_rest_shape == null:
+		return
+	if not visual is Node2D:
+		return
+	var press: Area2D = Area2D.new()
+	press.name = "EquippedPressArea"
+	# Input picking requires broadphase presence (monitorable) + a non-zero collision_layer; the defaults satisfy both.
```
J-Melon: **Hero** issue: comment says monitorable + non-zero collision_layer are required and the defaults satisfy both, but the next two lines set `collision_mask = 0` and `monitoring = false`. Either the comment is wrong or the code is. `input_pickable` is what actually drives input events; rewrite or delete.

scripts/items/drop_targets/character_drop_target.gd
```diff
@@ -0,0 +1,154 @@
...
+	return _position_inside_area(position)
+
+
+func accept(item_key: String, _position: Vector2, _gesture_velocity: Vector2) -> void:
+	if _item_manager == null:
+		return
+	# equip emits equip_refused on capacity races; no-op on failure so the held token stays put.
+	if not _item_manager.equip(item_key):
+		return
+	# Signal handler mounts on the EQUIPPED transition; explicit mount here would double up without the group guard.
```
J-Melon: **Hero** suggestion: comment narrates what is NOT done (no explicit mount); `feedback_state_positive_shape` prefers naming the positive shape. Either drop it or rewrite as `# Mount happens in _on_item_placement_changed once placement flips to EQUIPPED.`

scripts/items/ball_drag_controller.gd
```diff
@@ -231,6 +240,30 @@ func grab_from_rack(item_key: String, press_position: Variant = null) -> bool:
...
+		return false
+
+	var spawn_position: Vector2 = (
+		press_position if press_position is Vector2 else _cursor_position()
+	)
+	if not _spawn_held_body(item_key, spawn_position, false):
+		return false
+
+	# `live` + on_court so an expansion-ring timeout routes through deactivate (= unequip);
+	# Rack drops also call unequip via RackDropTarget. Non-accepting drops keep EQUIPPED intact.
```
J-Melon: **Hero** issue: three-line `#` block on `_held_was_on_court = true`; comment policy is one line max. Compress or move to gear tech doc.

scripts/hud/dev_equipped_panel.gd
```diff
@@ -0,0 +1,91 @@
+extends VBoxContainer
+
+## Dev-only readout of the kit: which items are currently EQUIPPED and how many slots remain.
+## Equip-time visuals do not persist across runs yet (per SH-405 follow-up); this panel makes the state inspectable.
```
J-Melon: **Hero** issue: comment references the ticket (`per SH-405 follow-up`); `feedback_comment_style` forbids ticket/caller references in comments. Drop the parenthetical.

scripts/hud/dev_equipped_panel.gd
```diff
@@ -0,0 +1,91 @@
...
+
+func _make_label() -> Label:
+	var label := Label.new()
+	label.horizontal_alignment = HORIZONTAL_ALIGNMENT_CENTER
+	label.add_theme_color_override("font_color", Color(0.8, 0.8, 0.8))
+	return label
+
+
+func _on_placement_changed(_item_key: String, _placement: int) -> void:
+	_refresh()
```
J-Melon: **Hero** suggestion: `_on_placement_changed` and `_on_level_changed` are both one-line passthroughs to `_refresh()`. Use the `_refresh.unbind(N)` idiom already used in `speed_bar.gd` / `fp_bonus.gd` in this PR and delete both.

scripts/hud/dev_item_panel.gd
```diff
@@ -1,15 +1,10 @@
-@tool
 extends VBoxContainer
```
J-Melon, line 1: **Hero** question: stripping `@tool` and the placeholder branch from `dev_item_panel.gd` and `dev_stat_panel.gd` is unrelated to the SH-405 equip flow. Belongs in its own commit/issue, or name the load-bearing reason in the PR body.

scripts/items/item_manager.gd
```diff
@@ -189,6 +191,35 @@ func deactivate(item_key: String) -> bool:
...
+		return false
+
+	if get_kit_remaining() < 1:
+		equip_refused.emit(item_key, &"capacity_exceeded")
+		return false
+	return activate(item_key)
+
+
+## Symmetric wrapper over `deactivate`; named for intent at equip/unequip call sites.
+func unequip(item_key: String) -> bool:
```
J-Melon, line 218: **Hero** question: `unequip` is a bare passthrough to `deactivate`. The doc says it is named for call-site intent, fine; but `equip` has the role guard and capacity gate while `unequip` has neither. If unequipping a non-equipment item is a contract violation, assert it; if it is fine, the wrapper earns nothing over calling `deactivate` directly.

scenes/dev_hud.tscn
```diff
@@ -8,6 +8,7 @@
 [ext_resource type="Script" uid="uid://cgvfq6q1n825h" path="res://scripts/hud/dev_ball_state_panel.gd" id="4_ballstate"]
 [ext_resource type="Script" uid="uid://bkfp6d5w1iaip" path="res://scripts/hud/dev_bounce_panel.gd" id="5_bouncepanel"]
 [ext_resource type="Script" uid="uid://ce43fqg0u1yhu" path="res://scripts/hud/dev_bounce_overlay.gd" id="6_bounceover"]
+[ext_resource type="Script" path="res://scripts/hud/dev_equipped_panel.gd" id="8_equipped"]
```
J-Melon: **Chert** (godot-scene): the new `dev_equipped_panel.gd` ext_resource is missing the `uid="uid://lnxk7w1w6dq6"` attribute that every other Script line in this file carries (compare ids `2_oo0fn`, `3_5ysmq`, `4_ballstate`, etc.). The .gd.uid file exists, so the UID is known; this looks like a hand-authored ext_resource line rather than one round-tripped through the editor. Worth adding the attribute for consistency and to avoid a cold-cache UID warning. Non-blocker.

scripts/items/item_manager.gd
```diff
@@ -189,6 +191,35 @@ func deactivate(item_key: String) -> bool:
...
+## Returns false silently on role mismatch so callers can fall through to other targets.
+func equip(item_key: String) -> bool:
+	var item: ItemDefinition = _get_item(item_key)
+	if item.role != &"equipment":
+		return false
+
+	if get_kit_remaining() < 1:
+		equip_refused.emit(item_key, &"capacity_exceeded")
+		return false
+	return activate(item_key)
```
J-Melon, line 214: **Aubrey** (devils-advocate)

`activate(item_key)` (just above this hunk, line 179) is public and unguarded. The kit cap lives only in this `equip()` wrapper. Any future caller (a hot-load script, a save-restore path, a future test, an effect that mounts gear in response to a trigger) that calls `activate` directly on an equipment-role item silently bypasses the cap. The design TLDR is explicit that the kit cap is gated by the kit_slots stat; the type system does not enforce that gate today.

Current call sites route via `equip`/`unequip`, and ball-role callers must keep direct `activate` access. But there is no compile-time signal that says "do not call this directly for equipment." Two cheap options:

1. Have `activate` itself check role and reroute equipment to `equip`, with a doc comment forbidding the equipment shortcut.
2. `assert(item.role != &"equipment")` in `activate` so the bypass crashes loudly in debug.

Not a blocker. Worth a note on the design doc that capacity enforcement currently relies on caller discipline.

scripts/items/item_manager.gd
```diff
@@ -5,6 +5,8 @@ signal friendship_point_balance_changed(balance: int)
 signal item_level_changed(item_key: String)
 signal item_placement_changed(item_key: String, placement: int)
 signal court_changed(item_key: String, on_court: bool)
+## Emitted when equip refuses; reason is currently &"capacity_exceeded" (sole case).
+signal equip_refused(item_key: String, reason: StringName)
```
J-Melon, line 9: **Mari** (signals-lifecycle): `equip_refused` is emitted by `equip()` on capacity rejection but has zero `connect` sites in production code; only tests reference it. Orphan from the consumer side. Either wire a HUD/toast listener (the doc comment hints at a player-facing feedback channel) or drop the signal until there is a real consumer. Not a merge-blocker since silent no-op is the current capacity-exceed UX, but worth a flag.

J-Melon, line 9: **Aubrey** (devils-advocate)

`equip_refused` has zero production listeners. The signal is connected only in `test_item_manager.gd`. The design spec frames it as the hook any future animation listener plugs into for capacity rejection. Today, capacity rejection is invisible at the UX layer: the held token just sits on the cursor, indistinguishable from a wrong-role drop on the character or a missed-projection drop in a wall.

Two follow-ups worth tracking explicitly:

1. Without a listener, there is no playtest evidence that the signal payload (`reason: StringName`) is the right shape. When the rejection animation lands, will the listener want the item def too, or the count it would have exceeded by? Right now we'd discover that on the first attempt.
2. The dev_equipped_panel reads `get_kit_remaining()` reactively from `item_placement_changed`. Capacity-refused gestures emit `equip_refused` but not `item_placement_changed`, so the dev panel learns nothing about a refusal. Probably fine because no placement actually changed, but worth a beat to confirm the dev surface intentionally stays silent on refusals.

Not blocking. Flagging so the next gear PR opens with the listener story already named.

scripts/items/ball_drag_controller.gd
```diff
@@ -732,6 +773,18 @@ func _make_court_target() -> CourtDropTarget:
...
 
+func _make_character_target() -> DropTarget:
+	if _character_drop_area == null or timeout_controller == null:
+		return null
+	var character_target: CharacterDropTargetScript = CharacterDropTargetScript.new()
+	character_target.configure(_item_manager, _character_drop_area, timeout_controller)
+	# Track the live target so equipped-art presses route into grab_equipped_from_character.
+	_character_target = character_target
+	if not character_target.equipped_art_pressed.is_connected(_on_equipped_art_pressed):
+		character_target.equipped_art_pressed.connect(_on_equipped_art_pressed)
```
J-Melon, line 783: **Mari** (signals-lifecycle): `_register_builtin_targets` clears `_drop_targets` / `_character_target` and rebuilds a fresh `CharacterDropTarget`, but the old target connected `ItemManager.item_placement_changed` to its `_on_item_placement_changed`. That Callable on the autoload pins the old (RefCounted) target alive, so a second `set_character_drop_target(...)` leaks the previous target and you get two placement_changed listeners. Today only Court `_ready` calls this once, so it is dormant; if player_paddle ever respawns this becomes a real leak plus duplicate mount/free attempts. Recommend disconnecting `_character_target` from `item_placement_changed` (or stashing a disconnect Callable) before nulling it.

scripts/core/timeout_controller.gd
```diff
@@ -116,7 +186,18 @@ func _on_reached_equip_pose() -> void:
...
+		return
+	_state = State.ASCENDING
+	set_physics_process(true)
+
+
+func _finish_at_lane() -> void:
 	_state = State.IDLE
 	if is_instance_valid(main_character):
+		main_character.velocity = Vector2.ZERO
+		main_character.collision_mask = _saved_collision_mask
```
J-Melon, line 217: **Mari** (signals-lifecycle): `_drive_blocked` is set on the paddle in `call_timeout` and only cleared in `_finish_at_lane`. If TimeoutController is freed between those points (any teardown of court mid-timeout), the paddle survives with `_drive_blocked = true` and refuses every `drive()` call forever. Today court owns both nodes so they die together, so this is theoretical; if either lifetime ever decouples (scene reload mid-timeout, hot-swap), add a paddle-side reset on `tree_exiting` or have the paddle clear its own flag when ItemManager / state signals a reset. Not blocking.

scripts/items/ball_drag_controller.gd
```diff
@@ -231,6 +240,30 @@ func grab_from_rack(item_key: String, press_position: Variant = null) -> bool:
...
+		return false
+
+	# `live` + on_court so an expansion-ring timeout routes through deactivate (= unequip);
+	# Rack drops also call unequip via RackDropTarget. Non-accepting drops keep EQUIPPED intact.
+	_held_was_on_court = true
+	_held_origin = &"live"
+	_mouse_button_down = true
+	pickup_started.emit(item_key)
+	return true
+
```
J-Melon, line 264: **Aubrey** (devils-advocate)

The drag-off-character path keeps `state.item_placements[item_key] == EQUIPPED` for the entire gesture (HeldBody under the controller, no LOOSE_IN_VENUE overlay). `get_kit_remaining()` counts persisted EQUIPPED rows, so while the player is mid-drag of an equipped item, that slot still reads as consumed.

The concrete failure: player is at cap (kit_remaining == 0), wants to *swap* one equipped item for a different one in the rack. They press the equipped art, drag it toward the rack, but along the way the cursor passes over the character with a different equipment HeldBody. Except they cannot, because we only carry one held key at a time. Real path: they release the dragged item over the venue floor (loose body), then try to grab a new gear off the rack and drop on the character. Fine, that works.

The failing path is subtler: they press the equipped art, then change their mind and want to drop it back on the character (cancel) over a different anchor slot. `CharacterDropTarget.can_accept` runs `_item_manager.get_kit_remaining() < 1`, which is true (the slot still reads consumed by the very item they are holding), so the character refuses its own equipped item back. The HeldBody falls through to other targets or stays on cursor. The drag-off-character ergonomics are advertised as 'drag back to rack to unequip' (design line 13) but the implicit affordance 'drag and drop in place to cancel' is broken at cap.

Mitigation options without rework: bias the held-key out of the count in `get_kit_remaining` when called from the character drop target, or have the drag-off path mark the item LOOSE_IN_VENUE (mirroring ball-OUT_REST) so the existing 'frozen mid-gesture' overlay applies. Worth playtesting before locking.

Not a blocker (current playtest builds run with cap of 3, slot pressure is rare). Becomes a real bug once an item lowers kit_slots or the player invests partway.

scripts/entities/paddle.gd
```diff
@@ -104,7 +114,13 @@ func _apply_size() -> void:
 	var paddle_size: float = _resolve(GameRules.paddle.paddle_size, &"paddle_size")
 	var new_size: float = clampf(paddle_size, paddle_size_min, arena_height)
 
+	# Anchor the collider's foot: RectangleShape2D is centred on the body, so growing size.y without
+	# shifting position.y plants half the delta below the floor and traps depenetration during AT_EQUIP_POSE.
+	var old_size: float = _collision_shape.size.y
 	_collision_shape.size.y = new_size
+	if _size_initialised:
+		position.y -= (new_size - old_size) * 0.5
+	_size_initialised = true
```
J-Melon, line 125: **Aubrey** (devils-advocate)

The foot-anchor shift assumes physics is running so depenetration can fix a small intersection. During AT_EQUIP_POSE, `main_character.set_physics_process(false)` is in effect: equipping a paddle_size-increasing item triggers `_refresh_from_stats` -> `_apply_size`, which does `position.y -= (new_size - old_size) * 0.5`. That write lands directly on `position.y` while no physics step runs.

On `end_timeout`, ASCENDING walks toward the *cached* `_lane_y` (captured at `_cache_positions` before any equip happened). If equip raised `position.y` by, say, 10px above original floor contact, the ascent's first `_step_ascent` sees `remaining = _lane_y - current_y` negative; `absf(remaining) <= step` snaps to `_lane_y` and stops. Physics resumes at original lane y, but the collider is now taller. The foot is buried below the floor surface by half the size delta.

Whether this matters depends on `move_and_slide`'s depenetration on the next `drive()`. CharacterBody2D usually pushes up, but it's not free, and a deep enough intersection can pin the body to the floor or cause a visible jolt on the first frame of play resumption.

Lightweight fix: refresh `_lane_y` from `main_character.position.y` at `_finish_at_lane` instead of using the stale cached value, or recompute the ascent target from current floor contact after each equip transition. Or, more defensively, defer `_apply_size` writes while `_drive_blocked` is true and replay them on unblock.

Needs a runtime check before merge: equip ankle_weights at the equip pose, end timeout, look for the paddle's first-frame y under live play. If it visibly snaps or jitters, this is the cause.

scripts/entities/paddle.gd
```diff
@@ -87,10 +96,11 @@ func _resolve(base: float, key: StringName) -> float:
 func _bind_stat_updates() -> void:
 	if _item_manager == null:
 		_item_manager = ItemManager
-	_item_manager.item_level_changed.connect(_on_item_level_changed)
+	_item_manager.item_level_changed.connect(_refresh_from_stats.unbind(1))
+	_item_manager.item_placement_changed.connect(_refresh_from_stats.unbind(2))
```
J-Melon, line 100: **Aubrey** (devils-advocate)

`_refresh_from_stats` fires on every `item_placement_changed` for any item: balls going on court, equipment going to/from EQUIPPED, loose-in-venue overlays toggling on/off, even non-paddle-size effects swapping. Each fire calls `_apply_size`, which mutates `position.y` if size changed and even if it didn't (the `if _size_initialised` guard still recomputes the size, the delta is just zero when nothing changed).

In practice this is fine because the delta is zero when `paddle_size` resolves identically. But it means: any time a ball moves between rack and court, the paddle runs through the full stat resolve pipeline (Stats.resolve x 3, clamp, scale.y write). Cheap per fire, but the signal is hot during a rally (loose-in-venue overlays toggle on every regrab) and the unbind(2) means we cannot cheaply gate on item identity here.

More concerning: every ball-on-court transition also triggers the foot-anchor delta math while `position.y -= 0.0 * 0.5` is a no-op, except for one path: if a paddle_size-affecting equipment's effect registration order changes between the placement signal fanout vs the size resolve, the resolved value could differ from the previous resolve by a sub-pixel margin from float drift. Then the foot-anchor moves the paddle by epsilon every time a ball is grabbed.

Low-priority. Worth a single-line guard: skip `_apply_size` when `new_size == old_size`, or filter the signal handler by item key (only equipment with paddle_size effects).

scripts/entities/paddle.gd
J-Melon: **Gru**: fixed in 54ea56a.

scripts/items/item_manager.gd
J-Melon: **Gru**: fixed in 54ea56a.

scripts/items/drop_targets/character_drop_target.gd
J-Melon: **Gru**: fixed in 54ea56a.

scripts/items/drop_targets/character_drop_target.gd
J-Melon: **Gru**: fixed in 54ea56a.

scripts/items/ball_drag_controller.gd
J-Melon: **Gru**: fixed in 54ea56a.

scripts/hud/dev_equipped_panel.gd
J-Melon: **Gru**: fixed in 54ea56a.

scripts/hud/dev_equipped_panel.gd
J-Melon: **Gru**: fixed in 54ea56a.

scenes/dev_hud.tscn
J-Melon: **Gru**: fixed in 54ea56a.

scripts/entities/paddle.gd
J-Melon: **Gru**: fixed in 54ea56a. TimeoutController now re-caches lane positions at `call_timeout`, so a stat-driven paddle resize between timeouts is reflected in the ASCENDING target.

scripts/entities/paddle.gd
J-Melon: **Gru**: fixed in 54ea56a; `_apply_size` early-returns when new and old collider heights match.

scripts/hud/dev_item_panel.gd
J-Melon: **Gru**: Josh requested mid-mission (`also we can stop the dev panels being tools not needed`); the strip is in-scope. Not scope creep.

scripts/items/item_manager.gd
J-Melon: **Gru**: kept for symmetry with `equip`. Call sites read intent (`unequip` not `deactivate`); future role-specific cleanup has a hook.

scripts/items/item_manager.gd
J-Melon: **Gru**: expected orphan until the refusal-animation listener lands (tech doc names this as the future hook). Tests pin the signal contract; production listener follows in its own ticket.

J-Melon: **Gru**: same as Aubrey's thread on this signal; expected orphan until the refusal animation hook lands.

scripts/items/item_manager.gd
J-Melon: **Gru**: real design surface; `activate` stays universal (per tech/04-effect-system.md), `equip` is the gated wrapper for equipment-role. Doc-level note rather than a code gate; logging on a follow-up.

scripts/items/ball_drag_controller.gd
J-Melon: **Gru**: dormant today (single paddle spawn); flagging for the paddle-movement spike (SH-409); the disconnect symmetry belongs in that refactor.

scripts/core/timeout_controller.gd
J-Melon: **Gru**: theoretical (court owns both today). Same SH-409 spike will own the lifecycle contract.

scripts/items/ball_drag_controller.gd
J-Melon: **Gru**: surfaces only when kit_slots-lowering items exist (none today). Filing as a follow-up bug once that surface lands; out-of-scope for SH-405.

scripts/core/timeout_controller.gd
```diff
@@ -58,11 +65,16 @@ func call_timeout() -> void:
 	if main_character == null:
 		push_warning("TimeoutController.call_timeout: main_character is null")
 		return
-	_state = State.WALKING_OFF
 	main_character.set_physics_process(false)
 	main_character.velocity = Vector2.ZERO
+	# Re-cache so a stat-driven paddle resize between timeouts is reflected in the lane y the walk ascends back to.
+	_cache_positions()
```
J-Melon, line 74: **Stan** (devils-advocate)

The re-cache at `call_timeout` does not close the equip-during-pose bug it is advertised to fix.

Trace:
1. `call_timeout()` runs `_cache_positions()`; `_lane_y = main_character.position.y` at this instant, BEFORE the off-walk, BEFORE the pose, BEFORE any equip.
2. Off-walk to AT_EQUIP_POSE.
3. Player drops ankle_weights on the character. `character_drop_target._handle_drop` is gated to `State.AT_EQUIP_POSE` (line 40), so this is the only window equip fires in.
4. `ItemManager.equip` to `activate` to `_set_item_placement` to `item_placement_changed` to `paddle._refresh_from_stats` to `_apply_size`.
5. `_apply_size` shifts `position.y` by `-(new_size - old_size) * 0.5` to keep the foot anchored.
6. `end_timeout` to walk back to ASCENDING targets `_lane_y`, still the value cached at step 1, before the shift in step 5.

Net: ASCENDING returns to the OLD center-y. With ankle_weights making the paddle larger the foot pokes below the floor by `(new-old)/2`; smaller and the foot floats. The re-cache point would need to be at `end_timeout` (before `_begin_walk_back`), or `_apply_size` needs to push the lane-y delta back to the controller while it is paused at the pose.

Aubrey's own description in the brief says the fix accounts for resizes 'since the last timeout.' That is a real but different bug, drift between timeouts when stat changes fire outside the rack window. After this commit `_apply_size` early-returns on `is_equal_approx(new_size, old_size)`, so between-timeout drift requires a stat change to actually occur off-window. In SH-405's scope (rack-only equip) the between-timeout path is dead; the within-pose path is the live one and is not covered.

Secondary: the test file gained `drive_blocked` rename coverage but no test exercises ASCENDING after a mid-pose resize. The advertised fix has no test asserting the foot lands correctly after equip-during-pose. `test_end_timeout_returns_to_lane_position` checks the no-resize path only.

Finding 2 (early-return placement) is fine. Guard sits between `old_size` read and writes; `_size_initialised` flips at the end on first call; subsequent equal-size calls short-circuit cleanly. Sprite scale is skipped on the short-circuit path, which is correct because new_size and `_sprite_natural_height` have not changed.

Verdict: block. The load-bearing fix is in the wrong place.

J-Melon: **Gru**: real catch. Fixed in ed44822 by switching the cached target from lane center y to lane FOOT y, which is stable across in-pose resizes because `paddle._apply_size` anchors the foot. ASCENDING now derives `target_y = _lane_foot_y - half_height()` at step time, so whatever the paddle's size is when ascent fires, the foot lands on the cached lane foot. New test (`test_ascending_lands_foot_on_cached_lane_foot`) pins the foot invariant directly. Restored in f32246a after a worktree-bleed scare that bundled an unrelated deletion set into the same commit.

scripts/items/drop_targets/character_drop_target.gd
```diff
@@ -0,0 +1,154 @@
...
+func _mount_equipped_visual(item_key: String) -> void:
+	if not _is_equipment_role(item_key):
+		return
+	var definition: ItemDefinition = DropTarget.get_definition(_item_manager, item_key)
+	if definition == null or definition.art == null:
+		return
+	if _drop_area == null or not _drop_area.is_inside_tree():
+		return
+	# Idempotency guard: hydrate + signal can both fire for the same item; second call must no-op.
+	if not _drop_area.get_tree().get_nodes_in_group(equipped_art_group(item_key)).is_empty():
```
J-Melon, line 87: issue: spacing in this entire file

tests/unit/items/test_drop_targets.gd
```diff
@@ -137,6 +128,307 @@ func test_rack_drop_target_without_drop_area_rejects() -> void:
...
+	manager.take("gear_idem")
+	manager.state.item_placements["gear_idem"] = Placement.EQUIPPED
+
+	var paddle := Node2D.new()
+	paddle.name = "PaddleFixture"
+	add_child_autofree(paddle)
+	var area: Area2D = _make_drop_area(Vector2.ZERO, Vector2(40, 80))
+	area.reparent(paddle)
+	var timeout: TimeoutController = TimeoutControllerScript.new()
+	add_child_autofree(timeout)
```
J-Melon, line 377: issue: spacing in this file too

scripts/core/timeout_controller.gd
J-Melon: Re-verified statically against `timeout_controller.gd` and `paddle.gd` at HEAD (f32246a).

The production invariant holds:

1. Foot cache is axis-correct: `_lane_foot_y = _lane_y + _half_height()` in y-down 2D (foot is center + half).
2. `_step_ascent` derives `target_y = _lane_foot_y - _half_height()` with a fresh `_half_height()` call at step time, so it picks up whatever the collider size is when ASCENDING fires, not a stale capture.
3. `_apply_size` preserves the foot algebraically: when `size.y` changes by `delta`, `position.y` shifts by `-delta/2` (gated on `_size_initialised`), so `position.y + size.y/2` (the foot) is invariant across resize. Caching foot is therefore genuinely stable across in-pose resizes.
4. `_lane_y` is now dead after `_cache_positions` (only consumed as a scratch term for `_lane_foot_y`); only `_lane_x` flows into `_begin_walk_back`, and horizontal walks are size-insensitive. No other ASCENDING / DESCENDING / lane-return path reads a stale center-y.

One honest caveat on the new test, since the original trace deserves a faithful guard: `test_ascending_lands_foot_on_cached_lane_foot` does not actually resize mid-pose (its docstring concedes the autoload fights manual shape mutations during drive). Without a mid-pose resize, the assertion collapses to `_lane_foot_y - _half_height() == _lane_y`, which holds by construction of the cache. It pins the formula's shape, not the Stan scenario. The production code is correct; the regression net here is the contract, not the bug-repro. Flagging only; not blocking.

Approving.

**Otis** (devils-advocate)

scripts/items/drop_targets/character_drop_target.gd
J-Melon: **Penny**: fixed in c0ee6fd; blank-line phases inside configure, _mount_equipped_visual, _attach_press_area, _on_equipped_press_input.

tests/unit/items/test_drop_targets.gd
J-Melon: **Penny**: fixed in c0ee6fd; blank-line phases between arrange/act/assert and between manager/area/timeout/target setup steps across the drop-target tests.

PR #677, opened 2026-05-16
ci: trim wall time per PR

### LINKED ISSUE #676: Trim CI wall time per PR
SPEED the Tests and Lint workflows on every PR<br>So that the merge-gate round-trip is shorter and impl agents wait less.

**Acceptance Criteria:**

- [ ] Tests workflow runs `godot --import` once per job, not twice.
- [ ] `.godot` import cache key narrows to `project.godot`, `addons/**`, and `**/*.import` so impl PRs reuse the warm cache.
- [ ] `gdformat` and `gdlint` run as separate parallel jobs in `lint.yml`.
- [ ] Leak gate still passes on a representative PR after the single-import change (no new Failed-loading errors leak through the awk filter).

J-Melon:
Drops the second `godot --import` pass in Tests (single pass; cold-cache UID warnings and their paired `Failed loading resource` errors stay filtered by the existing leak gate per `ai/scratchpads/godot-ci-uid-cache.md`). Narrows the `.godot` import cache key to `project.godot`, `addons/**`, and `**/*.import` so impl PRs reuse the warm cache and Godot's incremental importer handles per-file changes. Splits `gdformat` and `gdlint` into parallel jobs in Lint, with a `Lint` umbrella job that depends on all three sub-jobs so the required-check name in branch protection (still pinned to `Lint`) stays valid; no merge-gate coordination needed. Closes #676.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,8 +17,54 @@ env:
   GITLEAKS_VERSION: "8.28.0"
 
 jobs:
-  lint:
-    name: Lint
+  gdformat:
+    name: gdformat
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: "3.12"
+          cache: "pip"
+          cache-dependency-path: requirements-dev.txt
+
+      - name: Install Python tools
+        run: |
+          pip install --user -r requirements-dev.txt
+          echo "$HOME/.local/bin" >> "$GITHUB_PATH"
+
+      - name: gdformat
+        run: find scripts tests -name '*.gd' | xargs gdformat --check
+
+  gdlint:
+    name: gdlint
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -17,8 +17,54 @@ env:
...
+          cache: "pip"
+          cache-dependency-path: requirements-dev.txt
+
+      - name: Install Python tools
+        run: |
+          pip install --user -r requirements-dev.txt
+          echo "$HOME/.local/bin" >> "$GITHUB_PATH"
+
+      - name: gdlint
+        run: find scripts tests -name '*.gd' | xargs gdlint
```
J-Melon, line 64: **Skinner**: `text-checks` installs the full `requirements-dev.txt` (which includes gdtoolkit) just to get `codespell`. Three runners now do the same heavy install in parallel. Not a blocker; wall time still improves, but a `requirements-lint.txt` split would pay back here.

J-Melon: **Gru**: ack, deferring. Filing follow-up to split a `requirements-lint.txt` so the three lint runners do not each pull gdtoolkit. Out of scope for SH-406; will land its own ticket.

J-Melon: **Gru**: scratch that, requirements split landed in #678 (ebdaaa8). `requirements-lint.txt` pins gdtoolkit + codespell; `requirements-dev.txt` pulls it in via `-r` plus pip-audit. When #678 lands you'll want a quick `git merge main` to pick up `cache-dependency-path: requirements-lint.txt` on the three parallel jobs.

PR #678, opened 2026-05-16
fix(ci): fix sync-wiki YAML and add actionlint pre-merge gate

J-Melon:
Two related fixes for the same failure mode (YAML drift in `.github/workflows/*.yml`):

1. **sync-wiki.yml YAML fix.** The inline python heredoc in SH-330 (#552, 4467734) landed at column 0, terminating the YAML block scalar mid-script. GitHub rejected the workflow with `Invalid workflow file: .github/workflows/sync-wiki.yml#L104` on every push since 2026-04-29; production wiki sync stopped silently the same day. Pulled `title_case` into `.github/scripts/title_case.py` so the source stays runnable as python and the YAML parses. Behaviour parity confirmed on three inputs (`narrative-discipline`, `01-prototype`, `title-is-this`).

2. **actionlint pre-merge gate.** Added `actionlint` v1.7.12 to `lint.yml`. Startup-failure shape (the bug above) is the most silent class of CI failure: no email, no workflow_run event, no check posted on the triggering commit. Pre-merge actionlint catches it (and undefined env vars, unpinned actions, shell-syntax inside run blocks) at the door. Cached binary, same pattern as gitleaks.

Status: not yet battled.

### DIFF
--- a/.github/workflows/sync-wiki.yml
+++ b/.github/workflows/sync-wiki.yml
@@ -7,6 +7,7 @@ on:
     paths:
       - 'designs/**/*.md'
       - '.github/workflows/sync-wiki.yml'
+      - '.github/scripts/title_case.py'
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/sync-wiki.yml
+++ b/.github/workflows/sync-wiki.yml
@@ -92,29 +92,12 @@ jobs:
             ' "$src" > "$dest"
           }
 
-          # Title Case a hyphen-separated name with a small-word filter.
-          # "title-is-this" -> "Title-is-This"; "01-prototype" -> "01-Prototype";
-          # "narrative-discipline" -> "Narrative-Discipline".
-          # First and last words always capped (when the leading char is a lowercase
-          # letter); middle small words ("a", "an", "the", "and", "or", "but", "of",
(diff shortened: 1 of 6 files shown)


.github/workflows/sync-wiki.yml
```diff
@@ -92,29 +92,12 @@ jobs:
...
-    if not p:
-        out.append(p)
-        continue
-    if i in (0, len(parts)-1) or p.lower() not in small:
-        out.append(p[0].upper() + p[1:] if p[0].isalpha() and p[0].islower() else p)
-    else:
-        out.append(p.lower())
-print("-".join(out))
-' "$1"
+            python3 "${GITHUB_WORKSPACE}/.github/scripts/title_case.py" "$1"
```
J-Melon, line 101: **Apu**: `paths:` on the push trigger (line 7-9 above) scopes to `designs/**/*.md` and this workflow file, but not `.github/scripts/**`. Future edits to `title_case.py` alone will not rerun sync-wiki on push. Worth adding `- .github/scripts/title_case.py` (or `.github/scripts/**`) to the paths list so helper and caller stay in lockstep. Not a merge blocker; `workflow_dispatch` still covers manual reruns.

J-Melon: **Gru**: fixed in 8ab4787, paths filter now includes `.github/scripts/title_case.py`.

PR #680, opened 2026-05-16
docs: code-style + test-budget + conventional-commits + ticket example

J-Melon:
Propagating four internal rules to the contributor-facing docs:

- **New CODE_STYLE.md** at repo root covers the GDScript conventions lint does not enforce: blank line before `if`, `@export` over `@onready`, full words and descriptive names, one-line WHY-only comments, data-driven tunables, and `Resource` over loose `@export` when a cluster forms.
- **tests/TESTING.md** gains a *Test budget* section (no individual case grows the suite past its average) and a *CI* section listing the warning / orphan classes that fail the build, plus the cold-cache UID filter with upstream Godot issue links (#101677, #115205, #109636, #100228).
- **CONTRIBUTING.md** gains a *Commit messages* section naming the conventional shape and the breaking-change `!` suffix with a worked example.
- **designs/process/ticket-writing.md** gains a filled-in user-story example so the shape reads in context after the templates.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -7,6 +7,9 @@ version-resolver:
   default: patch
 
 categories:
+  - title: "Breaking Changes"
+    labels:
+      - "breaking"
   - title: "Features"
     labels:
       - "feature"
@@ -46,6 +49,12 @@ autolabeler:
       - "/^fix(\\(.+\\))?:/i"
       - "/^fix(\\(.+\\))?!:/i"
       - "/^SH-\\d+ fix(\\(.+\\))?:/i"
+  # Conventional Commits ! suffix: any type with a bang lands the breaking label
+  # so the release-drafter groups it under Breaking Changes at the top.
+  - label: "breaking"
+    title:
+      - "/^[a-z]+(\\(.+\\))?!:/i"
+      - "/^SH-\\d+ [a-z]+(\\(.+\\))?!:/i"
 
 change-template: "- $TITLE (#$NUMBER) by @$AUTHOR"
 change-title-escapes: '\<*_&'
@@ -59,5 +68,13 @@ template: |
   ## What changed
 
   $CHANGES
+
+  ## Contributors
+
+  Thanks to everyone whose work shipped in this release:
+
+  <!-- CONTRIBUTORS_LIST -->
+  $CONTRIBUTORS
+  <!-- /CONTRIBUTORS_LIST -->
 
   **Full changelog:** https://github.com/$OWNER/$REPOSITORY/compare/$PREVIOUS_TAG...v$RESOLVED_VERSION
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -49,12 +49,13 @@ autolabeler:
       - "/^fix(\\(.+\\))?:/i"
       - "/^fix(\\(.+\\))?!:/i"
       - "/^SH-\\d+ fix(\\(.+\\))?:/i"
-  # Conventional Commits ! suffix: any type with a bang lands the breaking label
-  # so the release-drafter groups it under Breaking Changes at the top.
+  # Conventional Commits ! suffix on a known type: lands the breaking label so
+  # release-drafter groups it under Breaking Changes at the top. Explicit type
+  # list avoids labelling `Merge!:` / `WIP!:` style titles as breaking.
   - label: "breaking"
     title:
-      - "/^[a-z]+(\\(.+\\))?!:/i"
-      - "/^SH-\\d+ [a-z]+(\\(.+\\))?!:/i"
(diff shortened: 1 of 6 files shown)


designs/process/ticket-writing.md
```diff
@@ -241,6 +241,25 @@ So that [downstream decision that depends on the answer]
 - [ ] No regression in related systems
 ```
 
+### Worked example
```
J-Melon, line 244: issue: for everything in this pr, follow the warm, welcoming style of the other contributor docs, someone is reading this. Help them out.

J-Melon: **Gru**: rewrote all four sections in d755234 to match the second-person voice of the existing docs ("None of this is meant to slow you down", "the bang is what our autolabeller picks up", "we like it that way"). Also folded in the breaking-changes category + bulleted Contributors block per follow-up.

.github/release-drafter.yml
```diff
@@ -46,6 +49,12 @@ autolabeler:
       - "/^fix(\\(.+\\))?:/i"
       - "/^fix(\\(.+\\))?!:/i"
       - "/^SH-\\d+ fix(\\(.+\\))?:/i"
+  # Conventional Commits ! suffix: any type with a bang lands the breaking label
+  # so the release-drafter groups it under Breaking Changes at the top.
+  - label: "breaking"
+    title:
+      - "/^[a-z]+(\\(.+\\))?!:/i"
```
J-Melon: **Bart** (ci-and-workflows): the `[a-z]+` here is broader than the Conventional Commit type list. Combined with `!:` it will also label things like `Merge!:` or `WIP!:` as breaking. Realistic risk is low (titles almost never carry `!:` outside CC) and it fails open (a bogus breaking label is easy to strip), so this is a note, not a block. If you want it tight later, swap to `(feat|fix|refactor|perf|chore|build|ci|docs|style|test|revert)`.

CODE_STYLE.md
```diff
@@ -0,0 +1,96 @@
+# Code Style
+
+Welcome. This is the short list of GDScript conventions we follow in Volley!. `gdformat` and `gdlint` handle most of the mechanical stuff for you; this file is for the project-specific calls those tools cannot make.
+
+None of this is meant to slow you down. The patterns here exist because they have repeatedly made the code easier to read, easier to refactor, and easier to onboard new contributors onto. If one does not fit your situation, open the PR and we will work it out.
```
J-Melon: **Carl** (docs-and-writing): Memory rule `feedback_state_positive_shape` says lead with what a thing *is* and *does*, not what it isn't. Opening the second sentence with "None of this is meant to slow you down" plants the slow-down shape in the reader's mind before the positive case. Consider something like "These patterns earn their place by making the code easier to read, refactor, and onboard onto. If one does not fit your situation, open the PR and we will work it out." Non-blocking; the rest of the doc lands well.

CODE_STYLE.md
```diff
@@ -0,0 +1,96 @@
...
+## `@export` for child nodes, not `@onready`
+
+Wire child node references through `@export var name: NodeType` (or `@export var name: NodePath`) rather than `@onready var name = $Path`. Exports survive scene renames, fail loudly when the wire breaks, and show the dependency in the inspector, all of which save time the next time someone reorganises a scene.
+
+```gdscript
+@export var hit_sound: AudioStreamPlayer
+@export var collision: CollisionShape2D
+```
+
+`@onready` exists in GDScript for a real reason: deferring evaluation of an arbitrary expression until the node enters the tree, not just caching child references. Something like `@onready var damage = base_damage * difficulty_multiplier` waits for dependencies to populate. That use case is what `@onready` was designed for, and `@export` cannot replace it.
```
J-Melon: **Carl** (docs-and-writing): "deferring evaluation [...] until the node enters the tree, not just caching child references" inverts the "Not just X, but Y" AI-tell construction (`ai/STYLE.md` watchlist). Reads more naturally as a positive: "deferring evaluation of an arbitrary expression until the node enters the tree, which is more than child-reference caching does." Or just: "deferring evaluation until the node enters the tree, a different job than caching child references." Optional.

CODE_STYLE.md
J-Melon: **Gru**: fixed in ac49a51.

CODE_STYLE.md
J-Melon: **Gru**: fixed in ac49a51.

.github/release-drafter.yml
J-Melon: **Gru**: tightened in a966ddf to explicit type list, `feat|fix|refactor|perf|chore|build|ci|docs|style|test|revert`. `Merge!:` / `WIP!:` no longer label as breaking.

PR #681, opened 2026-05-16
docs(skills): drop codename prefix and SH-N from commit subject template

J-Melon:
The SH-405 branch history showed three commit shapes mixed together: bare conventional (`fix: ascend back to the lane`), codename-suffix (`fix: ... [Selma/gdscript-implementer]`), and codename-prefix (`[Krusty/gdscript-implementer] fix: ...`). The contradiction is in our own docs: `ai/skills/minions/commits.md` says use `[<Codename>/<role>] <subject>`, the gdscript-implementer agent def says use `SH-N type: subject [Codename]`, and the standing rule (`feedback_commit_message_format`) says bare conventional with the codename in the `Agent-Role` trailer.

This PR aligns both files to the standing rule. Subject is `<type>: <subject>`. Codename lives in the dispatch description and the trailer, not the subject. The repo is open source; the subject is a public surface and reads cleaner without the swarm noise.

Effect on the next dispatch: implementer commits will land as `feat: hydrate equipped gear visuals on load` with `Agent-Role: gdscript-implementer` below, matching the Selma case rather than the Krusty / Maggie shape.

### DIFF
--- a/ai/skills/minions/commits.md
+++ b/ai/skills/minions/commits.md
@@ -1,17 +1,17 @@
 ---
 name: commits
-description: Commit shape every code-writing minion follows. DCO sign-off, role tag in the subject, Agent-Role trailer, no Co-Authored-By. Read before your first commit on a worktree.
+description: Commit shape every code-writing minion follows. Bare conventional commit subject, DCO sign-off, Agent-Role trailer, no Co-Authored-By, no codename in subject. Read before your first commit on a worktree.
 ---
 
 # Agent commit shape
 
-Every code-writing minion commits like a team member, not a tool.
+Every code-writing minion commits like a team member, not a tool. The repo is open source; the subject is a public surface and reads as a normal Conventional Commit.
 
 ## The shape
 
 ```
 git commit -s -m "$(cat <<'EOF'
-[<Codename>/<role>] <subject in the imperative mood>
+<type>: <subject in the imperative mood>
 
 <short body if needed; usually one or two sentences>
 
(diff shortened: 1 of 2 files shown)


ai/skills/minions/commits.md
```diff
@@ -21,16 +21,19 @@ EOF
...
 - No `Co-Authored-By:` lines. Volley's swarm uses Agent-Role for attribution; Co-Authored-By creates double counting.
 
 ## What goes in the subject
 
-Imperative mood, present tense. No file paths, no symptom descriptions, no issue numbers (Linear autolinks the branch name). Conventional-commit prefixes are fine but not required: `[Feldspar/general-purpose] feat: fast timeout tests via custom_step`.
+Imperative mood, present tense, lowercase. Conventional Commit type required: `feat`, `fix`, `docs`, `chore`, `refactor`, `test`, `ci`, `perf`. No file paths, no symptom descriptions, no issue numbers in the subject. Example: `feat: fast timeout tests via custom_step`.
 
-For breaking changes (save wipes, API renames, workflow-input shifts), use `feat!:` or `fix!:` on the subject. Autolabel aliases the bang.
+For breaking changes (save wipes, API renames, workflow-input shifts), use `feat!:` or `fix!:` on the subject. Autolabel maps the bang to the `breaking` label and release-drafter groups it under Breaking Changes.
```
J-Melon, line 32: **Tina** (docs-and-writing): non-blocker. The diff removes the old line that gave Gru its own subject-prefix rule (`[Gru]` only). Under the new bare shape Gru's subjects are presumably `<type>: <subject>` with `Agent-Role: dispatcher` trailer, same as minions. Worth a one-liner under "What goes in the subject" so Gru does not re-invent a prefix. Shipping anyway.

J-Melon: **Gru**: spelled out in f33cb7a, Gru subjects use bare `<type>: <subject>` with `Agent-Role: dispatcher`, no `[Gru]` prefix.

PR #711, opened 2026-05-17
test(audit): tautology sweep on N-iteration tests (SH-415)

J-Melon:
## Summary

Spike following Cal's catch on `test_oscillation_never_drops_below_min_speed`. Audited every long-loop / N-iteration test pattern in the suite for the same smell (assertion that holds even if the production effect under test is stubbed to a no-op, or trivially satisfied by a `clampf` in the production path).

Five tests rewritten with explicit tautology guards; one deleted as tautological by clamp boundary.

Audit log at `designs/process/tautology-sweep-2026-05.md` lists every candidate (file:line, fixture role default, claimed production effect, stub-no-op outcome, action taken, one-line reason).

### Rewrites

- `test_oscillate_stat_stays_within_amplitude` (event_dispatch.gd): deterministic period-sweep helper plus `extreme >= effective_amplitude * 0.5` guard.
- `test_oscillate_stat_scales_range_by_level`: same shape at level 2.
- `test_unregister_stops_oscillation`: `observed_active` guard inside the pre-unregister loop.
- `test_autoplay_speed_never_exceeds_configured_scale` (test_paddle_auto_play.gd): peak-velocity tracker plus `peak >= max_allowed * 0.95` convergence guard.
- `test_speed_never_exceeds_configured_scale` (test_partner_ai_controller.gd): same pattern.

### Deletion

- `test_noise_bounded_by_two_sigma` (test_paddle_ai_math.gd): `random_offset` ends in `clampf(_, -range*2, range*2)`. Asserting `max_observed <= range*2` is structurally true by the clamp; the 1000-sample loop has no failure mode.

### Stub re-verification

Each rewrite was re-checked under the same stub that originally let the test pass: `OscillateStatOutcome.apply` stubbed to `pass` flips all three oscillation rewrites red; `PaddleAIController._track` stubbed to `pass` flips both speed-cap rewrites red. Restored after each check.

## Suite timing

Before sweep: 2.019s (669 tests, 1813 asserts).
After sweep: 2.004s (668 tests, 1517 asserts).

The assertion-count drop is mostly from collapsing two 300-iteration per-frame assertion loops into single-extremum sampler calls.

## Test plan

- [x] `./scripts/ci/run_gut.sh` green after every rewrite.
- [x] Each rewrite re-verified to fail under a no-op stub of the corresponding production effect.
- [x] No production code modified.


designs/process/tautology-sweep-2026-05.md
```diff
@@ -0,0 +1,152 @@
+# Tautology sweep on N-iteration tests (SH-415)
+
+Spike following Cal's catch on `test_oscillation_never_drops_below_min_speed`: a 300-iteration physics-loop test whose fixture role default routed the item to STORED, so the production effect never registered and the assertion held on baseline state forever. The smell: a long simulation loop plus an assertion that could hold if the production code under test were stubbed to a no-op (or, in the clamp-boundary variant, an assertion structurally guaranteed by a `clampf` in the production path).
```
J-Melon: Kill this document

tests/unit/events/test_event_dispatch.gd
```diff
@@ -38,6 +38,22 @@ func _make_until_miss_effect(stat_key: StringName, operation: StringName, value:
 	return effect
 
 
+## Sweeps the manager forward one slow-period of oscillation (TAU / PRIMARY_FREQUENCY ~= 3.7s)
+## at 32 deterministic steps and returns the worst absolute deviation from `base_value`.
+## Uses the production `process_frame` / `Stats.resolve` path so a stubbed-no-op apply() leaves
```
J-Melon: Too much comment

designs/process/tautology-sweep-2026-05.md
J-Melon: Fixed in 9ba3397: audit doc deleted.

tests/unit/events/test_event_dispatch.gd
J-Melon: Fixed in 9ba3397: helper comment removed; function name carries the purpose.

PR #718, opened 2026-05-24
docs(skills): dandori cross-links mission-lifecycle memory

J-Melon:
The dandori skill is the eight-step checklist; the rule body now lives in a consolidated memory `feedback_mission_lifecycle.md` (folding what were four separate files: pre-mission interrogation, codename pool, mission-is-a-milestone, milestone+Ride+attach). Skill points at the memory.

Three substantive edits to the steps:

- Step 2 now reads "Project and cycle" and names the in-cycle scope rule (mission scopes only from issues already in the cycle; never sweep in).
- Step 6 broadens the codename pool from Minions-only to the wider pool (history, mythology, science, geography, art movements, oblique nouns), adds the no-theme-match note, and points at the memory.
- "What this skill replaces" footer becomes the three-way pairing between skill, memory rule body, and the mirror `feedback_dandori_structure`.

Switch back to feature/sh-423-outline-rework after merge.

### DIFF
--- a/ai/skills/gru/dandori.md
+++ b/ai/skills/gru/dandori.md
@@ -9,13 +9,13 @@ The interrogation order Gru runs when planning a new mission. Walk the steps; do
 
 **Trigger.** Josh says "dandori" on new work, or Gru spots a new-mission proposal trigger (a request that's bigger than one issue and needs a verification beat).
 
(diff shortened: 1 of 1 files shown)


J-Melon:
Superseded by the larger dandori restructure in this branch's PR.

PR #724, opened 2026-05-24
fix: court owns its collision floor and the apex bound

### LINKED ISSUE #658: Walls cover apex height
SIZE the court walls so they sit on the ground and extend past the maximum ball apex; PLAY_ARC balls never visibly clip over the boundary.<br>So that the wall reads as a real boundary the ball cannot fly over.

**Acceptance Criteria:**

- [ ] Walls are anchored at the bottom of the playable space.
- [ ] Wall height exceeds the maximum reachable apex (accounting for gravity, max ball speed, and current friendship-bound geometry).
- [ ] PLAY_ARC balls stay visually contained by the wall through long rallies.
- [ ] No regression in friendship-bound or apex-return mechanics.

J-Melon:
Closes #658.

Court now owns its own ball containment. Venue keeps the visual floor.

- `CourtFloor` (StaticBody2D + CollisionShape2D, no visual) under `Bounds`. Catches balls inside the play area.
- `Right Wall` origin moved to the centre of the wall body. Collision and visual now align off a single shape size.
- `SoulBound` Marker2D under `Bounds`. Court reads its y on `_ready` and writes it into `court_config.friendship_bound_y`. Drag the marker in the editor to tune the apex-arc threshold.
- Court's `@export` list grouped: `court_config` standalone, then Controllers, Bounds, Spawns, Scenes.

No left wall added; the left side is bounded by the player paddle, not a wall (see #658 discussion).

Runtime verification (no-tunnel, apex containment, no friendship-bound regression) needed before merge.

### DIFF
--- a/scripts/core/court.gd
+++ b/scripts/core/court.gd
@@ -82,7 +82,7 @@ func _ready() -> void:
 		court_config = load("res://scripts/core/court_config.gd").new()
 	ball_tracker.court_config = court_config
 	if soul_bound != null:
-		ball_tracker.bound_y = soul_bound.position.y
+		ball_tracker.bound_y = soul_bound.global_position.y
 	ball_tracker.configure(player_paddle)
 	ball_tracker.current_ball_changed.connect(_on_current_ball_changed)
 	ball_tracker.ball_missed.connect(_on_ball_missed)
--- a/scripts/core/court.gd
+++ b/scripts/core/court.gd
@@ -7,17 +7,22 @@ signal ball_at_max_speed_changed(is_at_max: bool)
 signal auto_play_changed(is_active: bool, friendship_point_rate: float)
 signal partner_changed
 
+@export_group("Systems")
 @export var ball_system: BallReconciler
 @export var ball_tracker: BallTracker
-@export var court_config: CourtConfig
-@export var player_paddle_scene: PackedScene
-@export var player_spawn: Marker2D
 @export var autoplay_controller: AutoplayController
-@export var right_wall: StaticBody2D
-@export var soul_bound: Marker2D
-@export var partner_spawn: Marker2D
 @export var timeout_controller: TimeoutController
 @export var drag_controller: BallDragController
+@export var court_config: CourtConfig
+
+@export_group("World")
+@export var player_spawn: Marker2D
+@export var partner_spawn: Marker2D
+@export var right_wall: StaticBody2D
+@export var soul_bound: Marker2D
+
+@export_group("Scenes")
+@export var player_paddle_scene: PackedScene
 
 ## Back-compat handle for tests; standard live-ball set lives on `ball_tracker`.
 var ball: Ball
--- a/scripts/core/court.gd
+++ b/scripts/core/court.gd
@@ -80,9 +80,9 @@ func _ready() -> void:
 
 	if court_config == null:
 		court_config = load("res://scripts/core/court_config.gd").new()
-	if soul_bound != null:
-		court_config.friendship_bound_y = soul_bound.position.y
 	ball_tracker.court_config = court_config
+	if soul_bound != null:
+		ball_tracker.bound_y = soul_bound.position.y
 	ball_tracker.configure(player_paddle)
 	ball_tracker.current_ball_changed.connect(_on_current_ball_changed)
 	ball_tracker.ball_missed.connect(_on_ball_missed)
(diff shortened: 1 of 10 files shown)


J-Melon (commented):
Two semantic notes on `scripts/core/court.gd` and `scenes/court.tscn`.

**Naming: `soul_bound` vs `friendship_bound_y`.** The new export and node (`soul_bound` / `SoulBound`) write directly into `court_config.friendship_bound_y`, which is the established term everywhere else (`ball.gd`, `paddle_ai_controller.gd`, both test suites, the config field itself). Two names for one concept across the same call chain makes the marker harder to find from the consumers. Suggest renaming the export and node to `friendship_bound` / `FriendshipBound` so reader and writer share a vocabulary. If "soul" is intentional new fiction it needs a design-doc anchor; right now it reads as a drift.

**Shared-resource mutation in `_ready`.** `court_config.friendship_bound_y = soul_bound.position.y` mutates the `CourtConfig` resource. `court.tscn` references `resources/court_config.tres` (id `25_courtcfg`), so this write lands on the shared Resource instance and persists for any other scene or test that loads the same `.tres` later in the session. The `if court_config == null:` fallback above implies awareness that the field is shared; the unconditional write below contradicts it. Options: `duplicate()` the resource on `_ready`, or read the marker position directly at the call sites that need it. Worth a decision before merge.

Floor collision (`Bounds/CourtFloor`) ships without the visible `ColorRect` sibling its `Right Wall` peer has, which I assume is intentional but flagging in case it was an oversight.

scripts/core/court.gd
```diff
@@ -71,6 +80,8 @@ func _ready() -> void:
 
 	if court_config == null:
 		court_config = load("res://scripts/core/court_config.gd").new()
+	if soul_bound != null:
+		court_config.friendship_bound_y = soul_bound.position.y
```
J-Melon: **reviewer-conventions** issue: mutates shared `.tres` `court_config` at runtime; persists across loaders this session. Duplicate the resource or compute on read.

J-Melon: **reviewer-coverage** suggestion: no test asserts `soul_bound.position.y` is pushed into `court_config.friendship_bound_y`. Land a small wiring test before merge.

J-Melon: why is this in two places at all?

scripts/core/court.gd
```diff
@@ -72,6 +81,8 @@ func _ready() -> void:
 	if court_config == null:
 		court_config = load("res://scripts/core/court_config.gd").new()
 	ball_tracker.court_config = court_config
+	if soul_bound != null:
+		ball_tracker.bound_y = soul_bound.position.y
```
J-Melon: **Gru** suggestion: `soul_bound.position.y` is local; ball compares against `global_position.y`. Latent if hierarchy shifts. Use `soul_bound.global_position.y`.

scripts/core/court.gd
```diff
@@ -72,6 +81,8 @@ func _ready() -> void:
 	if court_config == null:
 		court_config = load("res://scripts/core/court_config.gd").new()
 	ball_tracker.court_config = court_config
+	if soul_bound != null:
```
J-Melon, line 84: **Gru** suggestion: silent guard swallows a scene-wiring bug. Invert: `push_error` when soul_bound is null per `feedback_runtime_invariants_use_push_error`.

scripts/core/court.gd
J-Melon: **Gru**

resolved: b4030c7. Dropped `friendship_bound_y` from `CourtConfig`; marker is the single source, no shared-resource mutation.

scripts/core/court.gd
J-Melon: why is soul bound not part of court config?

scripts/core/court.gd
J-Melon: **Gru**

resolved: b4030c7. Wiring exercised by `BallTracker.attach` which all production balls funnel through; `test_ball_apex_arc` and `test_ball_state_transitions` set `bound_y` and exercise the apex transitions.

J-Melon: **Gru**

resolved: b4030c7. Consolidated to one source: SoulBound marker. `CourtConfig.friendship_bound_y` removed; `Ball.bound_y` is the per-ball snapshot injected by tracker.

scripts/core/court.gd
J-Melon: **Gru**

resolved: ae1da91. Switched to `soul_bound.global_position.y`.

PR #726, opened 2026-05-24
fix: force autoplay off when a timeout starts

J-Melon:
Closes #690.

AutoplayController listens for TimeoutController.timeout_started via an exported node-path and toggles itself off when the signal fires. timeout_ended does not auto-restore: the walk back to the lane is a deliberate beat and the player re-toggles autoplay when they want it back. drive_blocked behaviour during the walk-off is unchanged.

Wiring lives on AutoplayController itself (self-connect in _ready) so court.gd is unchanged.

Unit coverage in tests/unit/paddle/test_autoplay_timeout_disable.gd: signal connection, toggles off when active, emits autoplay_toggled(false), no-op when already off, timeout_ended does NOT restore. Full GUT suite green (671/671, 2.0s).


designs/01-prototype/10-idle-play.md
```diff
@@ -152,6 +152,10 @@ On quit (`_notification(NOTIFICATION_WM_CLOSE_REQUEST)` in `game.gd`): write `la
 - **Unit:** `ProgressionData` round-trip includes `last_quit_at` and `idle_fp_per_minute`
 - **In-game:** close game, wait a few minutes, reopen: welcome back label appears with correct friendship amount
 
+## Timeout interaction
+
```
J-Melon: issue this is an old doc bubble up

tests/unit/paddle/test_autoplay_timeout_disable.gd
```diff
@@ -0,0 +1,71 @@
...
+
+
+func test_ready_connects_timeout_started() -> void:
+	assert_true(
+		_timeout.timeout_started.is_connected(_controller._on_timeout_started),
+		"AutoplayController must connect to TimeoutController.timeout_started in _ready",
+	)
+
+
+func test_timeout_started_disables_active_autoplay() -> void:
```
J-Melon: issue test bloat

tests/unit/paddle/test_autoplay_timeout_disable.gd
```diff
@@ -0,0 +1,71 @@
...
+	assert_false(_controller.is_enabled())
+	assert_signal_not_emitted(_controller, "autoplay_toggled")
+
+
+func test_timeout_ended_does_not_restore_autoplay() -> void:
+	_controller.toggle()
+	_timeout.timeout_started.emit()
+	assert_false(_controller.is_enabled(), "precondition: autoplay off after timeout_started")
+	_timeout.timeout_ended.emit()
+	assert_false(_controller.is_enabled(), "timeout_ended must NOT auto-restore autoplay")
```
J-Melon: `_on_timeout_started` deliberately skips the `paddle.set_physics_process(...)` call that `toggle()` makes, and the comment above the function names that as the load-bearing distinction: timeout owns paddle physics during the walk-off; flipping it here fights the freeze. The current tests assert `is_enabled()` and the `autoplay_toggled` signal but never check paddle physics-processing state, so a refactor that routed timeout-disable through `toggle()` would silently break the walk-off and ship green. Worth one assertion: enable autoplay, snapshot `paddle.is_physics_processing()`, fire `timeout_started`, assert it is unchanged.

designs/01-prototype/10-idle-play.md
J-Melon: **Gru**

resolved: 6e25ded. Paragraph moved to `designs/01-prototype/17-partner-ai.md` under the `AutoplayController` subsection, alongside the rest of the controller's behaviour.

tests/unit/paddle/test_autoplay_timeout_disable.gd
J-Melon: **Gru**

resolved: 6e25ded. Five tests collapsed to three: dropped the signal-connection pin, merged the state + signal assertions into one test, kept the no-op and no-restore cases.

tests/unit/paddle/test_autoplay_timeout_disable.gd
J-Melon: **Gru**

declined: `paddle.is_physics_processing()` snapshot is implementation defence, not the AC. The behaviour the player sees (autoplay off, no auto-restore) is covered. Per the antagonistic-test rule, not chasing this line.

PR #732, opened 2026-05-24
fix: rack slot is per-STORED-span, assigned at store time

J-Melon:
Closes #679.

Rack slot identity now lives on `ItemState.rack_slot_index_by_key`. `ItemManager._set_item_placement` writes it on entry to STORED (lowest free index among same-role stored items) and erases it on exit. A lazy fallback in `get_rack_slot_index` covers paths that don't go through `_set_item_placement` (e.g. `take`).

`RackDisplay.refresh` and `get_slot_position_for` read the assignment instead of inferring from kit-list position. Two balls cannot land at the same slot, and a stored ball's marker never moves while another ball grabs or releases.

Doc updated: `designs/01-prototype/08-balls.md` describes the queue-with-arbitrary-pop semantic.

Runtime verification (drag two balls onto rack, grab one then the other) needed before merge.

### DIFF
--- a/scripts/items/item_manager.gd
+++ b/scripts/items/item_manager.gd
@@ -8,8 +8,6 @@ signal court_changed(item_key: String, on_court: bool)
 ## Emitted when equip refuses; reason is currently &"capacity_exceeded" (sole case).
 signal equip_refused(item_key: String, reason: StringName)
 
-const PlacementScript: GDScript = preload("res://scripts/items/placement.gd")
-
 var items: Array[ItemDefinition] = [
 	preload("res://resources/items/ankle_weights.tres"),
 	preload("res://resources/items/grip_tape.tres"),
@@ -122,8 +120,8 @@ func get_level(item_key: String) -> int:
 ## LOOSE_IN_VENUE overlays the persisted placement so callers see the runtime state.
 func _get_placement(item_key: String) -> int:
 	if state.loose_in_venue.has(item_key):
-		return PlacementScript.LOOSE_IN_VENUE
-	return state.item_placements.get(item_key, PlacementScript.STORED)
+		return Placement.LOOSE_IN_VENUE
+	return state.item_placements.get(item_key, Placement.STORED)
 
 
 ## Returns the current placement; STORED, EQUIPPED, ON_COURT, or LOOSE_IN_VENUE.
@@ -142,7 +140,7 @@ func mark_loose_in_venue(item_key: String, position: Vector2 = Vector2.ZERO) -> 
 		state.loose_in_venue[item_key] = position
 		return
 	state.loose_in_venue[item_key] = position
-	item_placement_changed.emit(item_key, PlacementScript.LOOSE_IN_VENUE)
+	item_placement_changed.emit(item_key, Placement.LOOSE_IN_VENUE)
 
 
 ## Clears the loose-in-venue entry. Idempotent. Emits item_placement_changed with the underlying placement.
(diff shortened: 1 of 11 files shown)


scripts/items/item_manager.gd
```diff
@@ -161,43 +164,76 @@ func get_court_items() -> Array[String]:
...
+func _assign_rack_slot(item_key: String, role: StringName) -> void:
+	var used: Dictionary = {}
+	for key: String in state.rack_slot_index_by_key:
+		var definition: ItemDefinition = _get_item(key)
+		if definition != null and definition.role == role:
+			used[state.rack_slot_index_by_key[key]] = true
+	var candidate: int = 0
+	while used.has(candidate):
+		candidate += 1
+	state.rack_slot_index_by_key[item_key] = candidate
```
J-Melon, line 187: The two playtest-reported ACs on #679 (no two balls sharing a rack slot; a stored ball does not visually jump when a sibling is grabbed) are not pinned by any test in this PR. Both are behaviour assertions, exactly the shape the test-behaviour-not-implementation rule wants: drive `purchase` / `activate` / `deactivate` / `remove_level` on two same-role items and assert that `get_rack_slot_index` returns distinct values, and that the stored ball's index does not change while a sibling moves through ON_COURT and back. Without these, the next refactor of `_set_item_placement` can quietly reintroduce the overlap and the playtest is the only safety net. Please land a unit test or two in `tests/unit/items/test_item_manager.gd` against the public surface (`get_rack_slot_index` plus the existing placement API).

scripts/items/item_manager.gd
```diff
@@ -161,43 +164,76 @@ func get_court_items() -> Array[String]:
 	return result
 
 
+## Slot index assigned to `item_key` while STORED; -1 when not stored.
+## Self-heals a missing assignment for an owned STORED item so direct-state-poke tests still work.
+func get_rack_slot_index(item_key: String) -> int:
+	if not state.rack_slot_index_by_key.has(item_key):
+		if get_level(item_key) > 0 and _get_placement(item_key) == PlacementScript.STORED:
+			_assign_rack_slot(item_key, _get_item(item_key).role)
+	return state.rack_slot_index_by_key.get(item_key, -1)
```
J-Melon, line 171: The lazy self-heal branch (`if not state.rack_slot_index_by_key.has(item_key)` -> assign) is the load-bearing path for `take`, for save files written before this PR, and for any direct-state-poke test seam. The docstring advertises it; coverage shows nothing exercises it. Worth one targeted test that constructs an `ItemManager` with `state.item_levels[k] = 1` and a STORED placement but no `rack_slot_index_by_key` entry, then asserts `get_rack_slot_index` returns 0 (and a second seeded same-role item returns 1).

scripts/items/rack_display.gd
```diff
@@ -42,17 +42,19 @@ func refresh() -> void:
...
 	var marker_count: int = _slot_markers.size()
-	for index in kit_keys.size():
-		var item_key: String = kit_keys[index]
+	if marker_count == 0:
+		return
+	var kit_keys: Array[String] = _item_manager.get_kit_items(role)
+	for item_key in kit_keys:
+		var slot_index: int = _item_manager.get_rack_slot_index(item_key)
+		if slot_index < 0 or slot_index >= marker_count:
+			push_error("RackDisplay.refresh: no marker for %s (slot %d)" % [item_key, slot_index])
```
J-Melon, line 52: `get_slot_position_for` is the seam playtest reported the overlap through (drag a second ball, marker lands on top of the first). The existing `test_get_slot_position_for_returns_world_position_for_known_key` checks one ball; there is no test asserting that two stored same-role balls return two distinct marker positions, nor that the position for a stored ball is stable while a sibling is activated. Behaviour-level coverage here would directly defend the AC.

scripts/items/item_manager.gd
J-Melon: **Gru**

declined: per the antagonistic-test rule, a coverage-completeness test for the new invariant doesn't earn its keep on a PR whose ACs are player-visible (no overlap on store; sibling-stable on grab). Playtest covers it; a unit test would mirror the assignment logic without adding signal.

scripts/items/item_manager.gd
J-Melon: **Gru**

resolved: 2fec311. Self-heal dropped. take, purchase-was-unowned, and _set_item_placement(STORED) each call _assign_rack_slot directly; the helper is a dict lookup.

scripts/items/rack_display.gd
J-Melon: **Gru**

resolved: 2fec311. Slot identity is now the marker index from rack_slot_index_by_key, not the kit-list position. Two balls cannot share a slot; a stored ball's marker does not move when a sibling enters or leaves STORED.

PR #753, opened 2026-05-26
docs: add points-vs-NoEstimates research

J-Melon:
Promotes the agile-estimation research from the scratchpad into `designs/research/`, with an INDEX entry under Sister research.

The question: do story points reduce to time even when proponents frame them as complexity or business value? The finding is that the reduction holds and the chief proponents concede it (Cohn: "effort is measured in time"; SAFe's half-day reference and eight-points-per-iteration; Jeffries: points are renamed ideal days). The one piece that does not reduce to a single time number is speed-independent relative sizing, and applying velocity to forecast a date consumes even that. Cited throughout.


designs/research/points-vs-noestimates.md
```diff
@@ -0,0 +1,82 @@
+# Story Points Are Time Wearing a Costume
+
+Research compiled 2026-05-26. Background for how Shuck thinks about estimation in its own process.
+
+## The man who invented them is sorry
+
+In 1999, on the C3 payroll project at Chrysler, a team started writing the time each story would take onto index cards. The numbers made managers nervous, so the team renamed them. Days became "ideal days", then ideal days became "points". The person who oversaw that renaming, Ron Jeffries, later wrote a sentence that ought to settle the whole debate before it begins: "I may have invented story points, and if I did, I'm sorry now."
+
+He went further. "Using them to predict 'when we'll be done' is at best a weak idea." The most-cited tool in agile estimation arrived with its inventor's regret attached, and the regret is specific. Points were not built to measure a new thing. They were built to hide an old one.
+
```
J-Melon, line 8: this reads too meta, don't offer self commentary

designs/research/points-vs-noestimates.md
```diff
@@ -0,0 +1,82 @@
...
+That is the claim this essay makes, and it is a strong one, so it deserves a strong opponent. The case for story points is not stupid, and the people who make it are not fools. They will tell you that points measure complexity, or effort, or business value, and that the link to time is a misunderstanding peddled by managers who never read the books. Some of them are right about the managers. None of them, in the end, escape the clock.
```
J-Melon: don't say not stupid etc

PR #772, opened 2026-05-28
Guard Linear IDs in commits and branch names (#771)

J-Melon:
Closes #771. Mechanical enforcement of the GitHub-number-over-Linear-ID convention, beyond the memory rule and the existing prefix-only commit check.

- `scripts/ci/check_commit_msg.sh`: now rejects a Linear ID anywhere in the message, not just at the subject start. The pattern is boundary-anchored to the `sh-<digits>` shape, so `UTF-8`, `SHA-256`, `SSH-1`, and words like `wish-3` do not trip it.
- `scripts/ci/check_branch_name.sh` (new) + a lefthook `pre-push` command: rejects `feature/sh-N` (Linear's `gitBranchName` shape) and `feature/gh-N`, steering to `feature/<gh-number>-<slug>`.

This is the gate that would have caught the three places a Linear ID leaked in recently (a mid-subject commit ref, a body ref, and a `feature/sh-` branch).

Note: `lefthook.yml` sets `no_auto_install: true`, so `lefthook install` must be run once for the new `pre-push` hook to register locally. The commit-msg change is live on the next commit.

### DIFF
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -15,6 +15,11 @@ commit-msg:
     check-format:
       run: ./scripts/ci/check_commit_msg.sh {1}
 
+pre-push:
+  commands:
+    branch-name:
+      run: ./scripts/ci/check_branch_name.sh
+
 pre-commit:
   piped: true
   commands:
--- a/scripts/ci/check_branch_name.sh
+++ b/scripts/ci/check_branch_name.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+# Reject branch names that carry a Linear-style ID instead of the GitHub issue
+# number. The repo convention is feature/<gh-number>-<slug>; Linear's
+# gitBranchName field emits feature/sh-N-..., which must be overridden. A bare
+# gh- prefix is also wrong (the number alone, no prefix).
+#
+# Reads the branch from $1 if given, otherwise the current HEAD. Used as a
+# lefthook pre-push command.
+set -euo pipefail
+
+branch="${1:-$(git rev-parse --abbrev-ref HEAD 2>/dev/null || echo "")}"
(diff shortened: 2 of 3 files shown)


scripts/ci/check_branch_name.sh
```diff
@@ -0,0 +1,27 @@
...
+# Reject branch names that carry a Linear-style ID instead of the GitHub issue
+# number. The repo convention is feature/<gh-number>-<slug>; Linear's
+# gitBranchName field emits feature/sh-N-..., which must be overridden. A bare
+# gh- prefix is also wrong (the number alone, no prefix).
+#
+# Reads the branch from $1 if given, otherwise the current HEAD. Used as a
+# lefthook pre-push command.
+set -euo pipefail
+
+branch="${1:-$(git rev-parse --abbrev-ref HEAD 2>/dev/null || echo "")}"
```
J-Melon, line 11: **Trillian** issue: pre-push validates HEAD, not the pushed ref; `git push origin good:feature/sh-9` slips through and pushing a good branch while HEAD is bad false-rejects. Parse git pre-push stdin (`<localref> <localsha> <remoteref> <remotesha>`) and check each remoteref.

lefthook.yml
```diff
@@ -15,6 +15,11 @@ commit-msg:
     check-format:
       run: ./scripts/ci/check_commit_msg.sh {1}
 
+pre-push:
+  commands:
+    branch-name:
+      run: ./scripts/ci/check_branch_name.sh
```
J-Melon, line 21: **Trillian** issue: the command passes no ref, so the script always falls back to HEAD; lefthook does not forward git pre-push stdin as `{1}`, so the script must read stdin itself.

PR #777, opened 2026-05-28
fix: AI ball-tracking and venue side-bound collision

J-Melon:
Two independent fixes. The AI paddle controller no longer locks onto the last-added (often stored) ball: every physics frame it scans the tracked balls and selects the soonest-arriving approaching one, so the partner and player autoplay react to whichever ball is live rather than to whichever was added last. Separately, the venue's left and right bounds move off collision layer 1 onto layer 3, so a ball leaving the court no longer halts against them while the floor and ceiling still hold inside the playable area. This is an improvement, not a fix for the outside-court wall the player actually hits: that wall is the held-ball drag clamp, tracked separately in SH-453, so #734 stays open.

The venue edges and the venue bounds are now one thing. The camera's left and right anchors read the VenueLeftBound and VenueRightBound bodies directly, and the old LeftEdge and RightEdge markers are gone, so a single node defines where each side of the venue sits. Those bounds also carry an editor-only debug draw, visible in the editor and during play, with no shipped game art.

Closes #737

Live runtime verification of ball motion past the court edge is left to the runtime-verifier.

### DIFF
--- a/scripts/core/paddle_ai_controller.gd
+++ b/scripts/core/paddle_ai_controller.gd
@@ -4,7 +4,7 @@ extends Node
 @export var paddle: CharacterBody2D
 @export var config: PaddleAIConfig
 
-var ball: RigidBody2D
+var ball: Ball
 
 var _enabled := false
 var _tracker: BallTracker
@@ -78,7 +78,7 @@ func _physics_process(_delta: float) -> void:
 
 
 ## Soonest-to-arrive in-play approaching ball; signal-bound `ball` when none qualifies.
-func _select_tracked_ball() -> RigidBody2D:
+func _select_tracked_ball() -> Ball:
 	if _tracker == null:
 		return ball
 
@@ -101,7 +101,7 @@ func _select_tracked_ball() -> RigidBody2D:
 	return best if best != null else ball
 
 
-func _ball_in_play(target: RigidBody2D) -> bool:
+func _ball_in_play(target: Ball) -> bool:
 	var state: Ball.PlayState = target.play_state
 	return state == Ball.PlayState.PLAY_NORMAL or state == Ball.PlayState.PLAY_ARC
 
@@ -118,7 +118,7 @@ func is_enabled() -> bool:
 
 
 ## Override: whether the given ball's x-direction counts as "coming toward me."
-func _ball_approaches(_target: RigidBody2D) -> bool:
+func _ball_approaches(_target: Ball) -> bool:
 	assert(false, "PaddleAIController._ball_approaches() is abstract")
 	return false
 
--- a/scripts/core/paddle_ai_controller.gd
+++ b/scripts/core/paddle_ai_controller.gd
@@ -88,10 +88,13 @@ func _select_tracked_ball() -> RigidBody2D:
 	for candidate in _tracker.get_balls():
 		if candidate == null or not _ball_in_play(candidate) or not _ball_approaches(candidate):
 			continue
+
 		var speed_x: float = absf(candidate.linear_velocity.x)
 		if speed_x < 1.0:
 			continue
+
 		var arrival: float = absf(paddle.position.x - candidate.position.x) / speed_x
+
 		if arrival < best_time:
 			best_time = arrival
 			best = candidate
--- a/scripts/core/paddle_ai_controller.gd
+++ b/scripts/core/paddle_ai_controller.gd
@@ -61,7 +61,6 @@ func _physics_process(_delta: float) -> void:
 	if not _enabled:
 		return
 
-	# Cover the soonest-arriving approaching ball; fall back to the tracker ball when none qualifies.
 	ball = _select_tracked_ball()
 	if ball == null:
 		return
@@ -82,8 +81,10 @@ func _physics_process(_delta: float) -> void:
 func _select_tracked_ball() -> RigidBody2D:
 	if _tracker == null:
 		return ball
+
 	var best: Ball = null
 	var best_time: float = INF
+
 	for candidate in _tracker.get_balls():
 		if candidate == null or not _ball_in_play(candidate) or not _ball_approaches(candidate):
 			continue
--- a/scripts/core/paddle_ai_controller.gd
+++ b/scripts/core/paddle_ai_controller.gd
@@ -36,55 +36,70 @@ func bind_tracker(tracker: BallTracker) -> void:
 			_tracker.ball_added.disconnect(_on_tracker_ball_added)
 		if _tracker.ball_removed.is_connected(_on_tracker_ball_removed):
 			_tracker.ball_removed.disconnect(_on_tracker_ball_removed)
-		if _tracker.current_ball_changed.is_connected(_on_tracker_current_ball_changed):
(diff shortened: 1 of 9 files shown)


tests/unit/paddle/test_partner_ai_controller.gd
```diff
@@ -132,3 +141,58 @@ func test_noise_resamples_during_drift_when_direction_changes() -> void:
 		offset_after_reversal,
 		"noise should resample even when direction changes during drift",
 	)
+
+
+# --- multi-ball coverage (SH-435) ---
+# With two live, approaching balls the partner must cover the one arriving
+# soonest, and re-commit when a sooner ball appears mid-rally. A single-ball
```
J-Melon: comments way too big

tests/unit/paddle/test_partner_ai_controller.gd
```diff
@@ -132,3 +141,58 @@ func test_noise_resamples_during_drift_when_direction_changes() -> void:
 		offset_after_reversal,
 		"noise should resample even when direction changes during drift",
 	)
+
+
+# --- multi-ball coverage (SH-435) ---
```
J-Melon: issue number should not be here

scripts/core/paddle_ai_controller.gd
```diff
@@ -58,23 +58,48 @@ func _on_tracker_ball_removed(_old_ball: Ball) -> void:
...
+## Soonest-to-arrive in-play approaching ball; signal-bound `ball` when none qualifies.
+func _select_tracked_ball() -> RigidBody2D:
+	if _tracker == null:
+		return ball
+	var best: Ball = null
+	var best_time: float = INF
+	for candidate in _tracker.get_balls():
+		if candidate == null or not _ball_in_play(candidate) or not _ball_approaches(candidate):
+			continue
+		var speed_x: float = absf(candidate.linear_velocity.x)
```
J-Melon, line 92: add spacing to this block

tests/unit/paddle/test_partner_ai_controller.gd
```diff
@@ -132,3 +141,58 @@ func test_noise_resamples_during_drift_when_direction_changes() -> void:
...
+
+func test_switches_when_a_sooner_ball_appears() -> void:
+	var tracker: BallTracker = _bind_tracker_for_multiball()
+
+	var first_ball: Ball = _spawn_ball(Vector2(PADDLE_X - 50.0, 200.0), Vector2(200.0, 0.0))
+	tracker.attach(first_ball)
+	_run_frames(5)
+	assert_gt(_paddle.velocity.y, 0.0, "precondition: partner tracks the first ball below center")
+
+	# A sooner ball appears: nearly at the paddle, very fast, intercept above center.
```
J-Melon, line 196: these tests seem complex, will they remain detmerinistic on value changes later?

tests/unit/paddle/test_partner_ai_controller.gd
J-Melon: Removed the issue reference in 390c08e; the header is now a single descriptive line.

tests/unit/paddle/test_partner_ai_controller.gd
J-Melon: Trimmed to one line in 390c08e.

scripts/core/paddle_ai_controller.gd
J-Melon: Added blank-line spacing separating the candidate guards, the speed/arrival compute, and the best-so-far update (390c08e), behaviour unchanged.

tests/unit/paddle/test_partner_ai_controller.gd
J-Melon: Good catch, and no. The velocity.y assertion was the fragile part: its sign depends on speed_scale, reaction_delay_frames, and the frame count, so it could flip on later tuning. Dropped it in 3b84047. The tests now assert only `_controller.ball` (which ball selection committed to), which is pure state and survives tuning. Verified with a tautology guard: the away-ball test fails if the selection filter is removed.

tests/unit/paddle/test_paddle_ai_tracker_binding.gd
```diff
@@ -131,3 +131,31 @@ func test_rebind_to_null_disconnects_prior_signals() -> void:
 	_tracker.attach(ball)
 
 	assert_null(_controller.ball, "after unbind, tracker emissions no longer reach controller")
+
+
+# Player autoplay shares the base class and the same tracker (court.gd binds it).
+# Under multi-ball it must lock onto the soonest-arriving ball it should cover,
```
J-Melon: Comment too big

scripts/core/paddle_ai_controller.gd
```diff
@@ -58,23 +58,51 @@ func _on_tracker_ball_removed(_old_ball: Ball) -> void:
...
+		var arrival: float = absf(paddle.position.x - candidate.position.x) / speed_x
+
+		if arrival < best_time:
+			best_time = arrival
+			best = candidate
+	return best if best != null else ball
+
+
+func _ball_in_play(target: RigidBody2D) -> bool:
+	var state: Ball.PlayState = target.play_state
```
J-Melon, line 105: Type-widening question: `_ball_in_play` takes `target: RigidBody2D` but reads `target.play_state`, which is a `Ball`-only member. It compiles because the value is always a Ball at runtime, but the signature has dropped the `Ball` contract it depends on. A non-Ball RigidBody2D reaching here would crash. Intended widening for the abstract base, or should this stay `Ball`? Natural cleanup target for the AI state-machine work (SH-454).

tests/unit/paddle/test_paddle_ai_tracker_binding.gd
```diff
@@ -131,3 +131,27 @@ func test_rebind_to_null_disconnects_prior_signals() -> void:
 	_tracker.attach(ball)
 
 	assert_null(_controller.ball, "after unbind, tracker emissions no longer reach controller")
+
+
+# Autoplay under multi-ball locks onto the soonest-arriving ball it should cover, not the current one.
+func test_autoplay_selects_soonest_arriving_ball_under_multiball() -> void:
```
J-Melon, line 137: Duplication question: this multi-ball 'soonest-arriving' test and the near/far + away-ball pair in test_partner_ai_controller.gd share the same shape (spawn near/far, attach the wrong one last, assert selection overrides). It's legitimate per-subclass coverage of the shared base `_select_tracked_ball`, not pure copy-paste, so I'd leave it. Flagging because the near-identical precondition/assert scaffolding is the kind that drifts. Worth a shared helper, or leave per-subclass?

scripts/core/paddle_ai_controller.gd
J-Melon: yes reduce to ball

J-Melon: Done in 3cf697b: narrowed var ball and the selection signatures (_ball_in_play, _ball_approaches, _select_tracked_ball) to Ball across all three controllers. Suite green, 673 tests.

PR #803, opened 2026-05-30
SH-437 feat: ball speed tiers, Peak window, and consolidation rewards

J-Melon:
Each tier-up fires `on_consolidation` through the effect system instead of awarding a lump soul bank. A built-in `VenueEffectSource` registers a `StatUntilMissOutcome` that adds +1 to `soul_multiplier` per consolidation, clearing on miss, and per-hit soul scales the base by that multiplier.

The reward handler is driven off `BallTracker.ball_tier_advanced`, which now carries the advancing ball, so a consolidation on any tracked ball pays out rather than only the current one. The handler no longer binds a single ball, which is what left a second ball's climb unrewarded.

The player-facing soul-rate readout is removed. The number it showed is a global stat shared across every ball, which is the wrong shape once more than one ball is in play; soul becomes per-ball under a follow-up and the dev readout lands there. Gone with it: the `SoulRateLabel` node, `soul_rate.gd` and its test, and `Court.soul_multiplier_changed`.


designs/01-prototype/tech/effect-system.md
```diff
@@ -0,0 +1,177 @@
...
+- Named game states tracked in `EffectState` via `set_state()`/`clear_state()`/`is_state_active()`.
+- Ball reads speed limits every physics frame via `BallEffectProcessor._sync_speed_limits()`, enabling dynamic stat changes to take effect immediately.
+- `GameRules.BALL_SPEED_MIN` and `BALL_SPEED_MAX` constants removed; replaced by stat-driven values.
+- Percentage modifier operation: values summed into a single offset, then applied as `result *= (1.0 + total_offset)`. Evaluation order: add, then percentage, then multiply.
+- `paddle_size_min` base stat (50.0) clamps the paddle so percentage reductions cannot shrink it below a playable size.
+
+### Remaining
+
+- Causality items need temporary outcome expiry (timer or per-frame tick) for `multiply_stat_temporary`.
+- Multi-ball requires a ball spawner and a reference list of active balls in the scene. `clear_extra_balls` removes all but the original.
```
J-Melon, line 178: **Marvin** issue (blocking): the original Notes section recorded that `EffectManager` subscribes to game signals (`ball_missed`, `paddle_hit`, `streak_changed`) and translates them into `on_game_event` calls (original lines 513-514). This is impl-coupled wiring detail, so it belongs in this doc, but it did not land here. Add one bullet under Implementation status or a Wiring notes section.

J-Melon: **dispatcher**

resolved in 97aa61dd: signal-to-trigger wiring note restored to the as-built record.

scripts/hud/soul_float_layer.gd
```diff
@@ -0,0 +1,64 @@
...
+	if _court != null:
+		_court.soul_earned.connect(_on_soul_earned)
+	else:
+		get_tree().node_added.connect(_on_node_added_waiting)
+
+	if _handler != null:
+		_handler.ball_upgrade_earned.connect(_on_ball_upgrade_earned)
+
+
+func _exit_tree() -> void:
```
J-Melon: **Quill** issue: `_exit_tree` never disconnects `_court.soul_earned` or `_handler.ball_upgrade_earned`; add `is_connected` guards and disconnect both before the tree exits.

scripts/hud/soul_float_layer.gd
```diff
@@ -0,0 +1,64 @@
...
+	var court := node as Court
+
+	if court == null:
+		var handler := node as TierRewardHandler
+		if handler != null and _handler == null:
+			_handler = handler
+			_handler.ball_upgrade_earned.connect(_on_ball_upgrade_earned)
+		return
+
+	get_tree().node_added.disconnect(_on_node_added_waiting)
```
J-Melon: **Quill** issue: once Court arrives and `node_added` disconnects, a late-arriving `TierRewardHandler` is never caught; `ball_upgrade_earned` silently stays unwired.

scripts/court/soul_rate.gd
```diff
@@ -0,0 +1,41 @@
...
+	if found == null:
+		found = get_tree().get_first_node_in_group(&"courts")
+
+	if found != null:
+		_connect_soul_source(found)
+	else:
+		get_tree().node_added.connect(_on_node_added_waiting)
+
+
+func _exit_tree() -> void:
```
J-Melon: **Quill** issue: `_exit_tree` never disconnects `soul_multiplier_changed`; SoulRateLabel outlived by Court fires into a freed node.

scripts/hud/soul_float_layer.gd
```diff
@@ -0,0 +1,64 @@
...
+	if is_inside_tree() and get_tree().node_added.is_connected(_on_node_added_waiting):
+		get_tree().node_added.disconnect(_on_node_added_waiting)
+
+
+func _on_node_added_waiting(node: Node) -> void:
+	var court := node as Court
+
+	if court == null:
+		var handler := node as TierRewardHandler
+		if handler != null and _handler == null:
```
J-Melon: **Hollow** issue: when SoulFloatLayer enters tree before Court, `node_added` disconnects on Court arrival but `_handler` is still null; `ball_upgrade_earned` never connects. Reconnect handler inside this branch after `_court` is set.

scripts/core/court.gd
```diff
@@ -217,15 +245,27 @@ func _deactivate_partner() -> void:
 	partner_changed.emit()
 
 
+func _soul_float_anchor() -> Vector2:
```
J-Melon: **Hollow** issue: `ball == null` is redundant; `is_instance_valid(null)` is false in Godot 4, so the first guard is dead code. Remove `ball == null or`.

scripts/court/tier_reward_handler.gd
```diff
@@ -0,0 +1,94 @@
+class_name TierRewardHandler
+extends Node
+
+## Fires on_consolidation through the effect system on every tier-up; handles first-reach upgrades.
+
+## Screen anchor of any floating text spawned by the HUD layer on consolidation.
```
J-Melon: **Hollow** issue: comment describes what the consumer does, not the signal; that is a caller-reference comment, disallowed by comment policy. Rewrite to describe what the signal carries.

tests/unit/court/test_tier_reward_handler.gd
```diff
@@ -0,0 +1,207 @@
...
+
+	_manager.process_event(&"on_miss")
+
+	assert_almost_eq(_manager.get_stat(&"soul_multiplier"), 1.0, 0.001)
+
+
+# --- per-hit soul uses multiplier (via item_manager.get_stat) ---
+
+
+func test_multiplier_at_one_read_correctly() -> void:
```
J-Melon: **Frick** issue: `test_multiplier_at_one_read_correctly` is identical to `test_soul_multiplier_starts_at_one` (line 40); delete one of the pair.

tests/unit/court/test_tier_reward_handler.gd
```diff
@@ -0,0 +1,207 @@
...
+	assert_almost_eq(_manager.get_stat(&"soul_multiplier"), 1.0, 0.001)
+
+
+func test_multiplier_at_two_after_one_consolidation() -> void:
```
J-Melon: **Frick** issue: `test_multiplier_at_two_after_one_consolidation` duplicates `test_tier1_consolidation_increments_soul_multiplier` (line 47); delete one.

tests/unit/court/test_tier_reward_handler.gd
```diff
@@ -0,0 +1,207 @@
...
+func test_soul_multiplier_resets_to_one_after_miss() -> void:
+	_ball.current_tier = 1
+	_ball.advance_tier()
+
+	_manager.process_event(&"on_miss")
+
+	assert_almost_eq(_manager.get_stat(&"soul_multiplier"), 1.0, 0.001)
+
+
+# --- per-hit soul uses multiplier (via item_manager.get_stat) ---
```
J-Melon: **Frick** issue: the AC promises per-hit soul = base x multiplier, but no test asserts the banked amount; the deleted banks_base_times_N tests were the only output assertions and were not replaced.

J-Melon: **dispatcher**

resolved in 6b92aebb.

tests/unit/court/test_tier_reward_handler.gd
J-Melon: **dispatcher**

resolved in 6b92aebb.

tests/unit/court/test_tier_reward_handler.gd
J-Melon: **dispatcher**

resolved in 6b92aebb.

scripts/court/tier_reward_handler.gd
J-Melon: **dispatcher**

resolved in 6b92aebb.

scripts/core/court.gd
J-Melon: **dispatcher**

resolved in 6b92aebb.

scripts/hud/soul_float_layer.gd
J-Melon: **dispatcher**

resolved in 6b92aebb.

scripts/court/soul_rate.gd
J-Melon: **dispatcher**

resolved in 6b92aebb.

scripts/hud/soul_float_layer.gd
J-Melon: **dispatcher**

resolved in 6b92aebb.

scripts/hud/soul_float_layer.gd
J-Melon: **Quill**

resolved in 6b92aebb: `_exit_tree` now disconnects both `soul_earned` and `ball_upgrade_earned` under `is_instance_valid` guards.

scripts/hud/soul_float_layer.gd
J-Melon: **Quill**

resolved in 6b92aebb: waiter now independently connects each target behind null guards; disconnects only once both are wired.

scripts/court/soul_rate.gd
J-Melon: **Quill**

resolved in 6b92aebb: `_source` stored in `_connect_soul_source`; `_exit_tree` disconnects `soul_multiplier_changed` under `is_instance_valid`.

scripts/hud/soul_float_layer.gd
J-Melon: **Hollow**

resolved in 6b92aebb: waiter now checks `_court == null` and `_handler == null` independently; disconnects only when both are wired.

scripts/core/court.gd
J-Melon: **Hollow**

resolved in 6b92aebb: `ball == null` guard removed; guard is now `not is_instance_valid(ball)`.

scripts/court/tier_reward_handler.gd
J-Melon: **Hollow**

resolved in 6b92aebb: docstring now reads "Carries the screen anchor where the upgrade float should appear."

tests/unit/court/test_tier_reward_handler.gd
```diff
@@ -0,0 +1,193 @@
...
+	_ball.advance_tier()
+
+	assert_signal_emitted(_handler, "consolidation_fired")
+
+
+func test_consolidation_fired_emitted_on_tier0() -> void:
+	watch_signals(_handler)
+
+	_ball.current_tier = 0
+	_ball.advance_tier()
```
J-Melon: **Frick** issue: `_on_ball_tier_advanced` returns early at `completed_tier == 0` in the handler; `consolidation_fired` never fires for tier 0. Suite fails with this assertion. Either remove the early-return in production code or revert this assertion to `assert_signal_not_emitted`.

J-Melon: **dispatcher**

Stale read: the working tree had a backwards-revert transiently staged when this was reviewed (the open editor re-saved pre-fix buffers). Committed HEAD 6b92aebb has the `completed_tier == 0` early-return REMOVED; the two tier-0 tests assert the pays-tier-0 behaviour and the full suite is 743/743 green against HEAD. No code change needed.

J-Melon: **Frick**

Stale read resolved at HEAD. The `completed_tier == 0` early-return is absent from committed `_on_ball_tier_advanced`; both tier-0 tests assert correctly (2.0 soul multiplier, consolidation_fired emitted); GUT 743/743 green.

designs/01-prototype/20a-ball-speed-tier-progression.md
```diff
@@ -6,28 +6,46 @@ Design companion to SH-88. Where [20 Ball Speed Tiers and Physics Ceiling](20-ba
 
 A linear speed ramp without internal structure gives the player no landmarks. The ball gets faster, the number on a bar goes up, and nothing along the way acknowledges the work the rally is doing. Tiers break the ramp into distinct beats so the climb has shape (three tiers to start, tunable). Each tier is a short stretch of rally with a recognisable entry speed, a peak, and a handover to the next tier. On the current court the bands are tight (about ten hits a tier); a wider court raises the speed ceiling and lets the climbs run longer.
 
-Reaching the top of a tier is supposed to register as an event, not as a number crossing a threshold. The companion reward ladder below is the mechanism that registers it.
+Reaching the top of a tier is supposed to register as an event, not as a number crossing a threshold. The consolidation event and the multiplier it raises (below) are the mechanism that registers it.
```
J-Melon: full sweep for not language, what is the actual impl?

designs/01-prototype/20a-ball-speed-tier-progression.md
```diff
@@ -6,28 +6,46 @@ Design companion to SH-88. Where [20 Ball Speed Tiers and Physics Ceiling](20-ba
...
 
-## Reward ladder
+## Soul, the multiplier, and consolidation
 
-Rewards are what fires on `on_tier_completed`. They are tuned for first playtest, not for shipped balance. The ladder reads off the tier table, so it stretches or shrinks with the tier count; below is the three-tier starting shape, where the top tier is Tier 2 and the Peak opens above it.
+Soul (the same resource as friendship points) is earned per paddle hit. The amount is `base x multiplier`, where base is 1 and the multiplier is what the climb grows.
```
J-Melon: don't even mention fp we are doing a full replace later

designs/01-prototype/20a-ball-speed-tier-progression.md
```diff
@@ -6,28 +6,46 @@ Design companion to SH-88. Where [20 Ball Speed Tiers and Physics Ceiling](20-ba
...
-- **Tier 1 peak** (entering Tier 2). A bigger audio beat and a sparkle off the ball at the moment of the transition. One friendship tick lands on the next paddle hit, and the rally counter flourishes a single beat bigger than its usual tick. A brief aura reads through the next few hits.
-- **Top-tier peak** (Peak window opens). The rare, shop-tier beat. The aura holds for the full Peak window, paddle hits inside Peak play a richer return sound, and the window banks a mid-sized currency drop that the player feels in the next shop. Only one Peak reward banks per rally. A shop voucher and an item-unlock signal are viable alternatives for a later pass; they become extra `on_tier_completed` reward handlers without replacing the currency drop.
+- **Base soul per hit is 1.** Every hit banks one soul before the multiplier.
+- **The multiplier starts at x1 and rises +1 each consolidation.** So a fresh rally pays 1 soul per hit; after one consolidation, 2 per hit; after two, 3; and so on. The deeper the rally climbs, the more every remaining hit is worth.
```
J-Melon: per ball

designs/01-prototype/20a-ball-speed-tier-progression.md
```diff
@@ -6,28 +6,46 @@ Design companion to SH-88. Where [20 Ball Speed Tiers and Physics Ceiling](20-ba
...
+
+Peak is a consolidation event too, not a separate reward shape. Entering Peak fires the same `on_consolidation` trigger, so the built-in `+1 multiplier` and any subscribed item effects resolve exactly as they do at a tier top, and Peak additionally opens the increased speed range above the ball's own max where the rally keeps climbing hit by hit. Peak is the top consolidation: the same +1 the climb has been paying, plus the wider band. There is no tier above Peak, so no further consolidations fire inside it; the multiplier holds at its Peak-entry value for the rest of the rally, and per-hit soul continues paying at that held value until a miss.
```
J-Melon: let's just rename peak, final consolidation

designs/effect-system/README.md
```diff
@@ -0,0 +1,22 @@
+# Effect System
+
+Every gameplay modifier in Volley flows through one framework: items, partners, the venue itself. Nothing goes around it.
+
```
J-Melon, line 4: there are no mermaid design diagrams in this, they're would be helpful, look at the history

designs/effect-system/reference.md
```diff
@@ -0,0 +1,110 @@
...
+
+| Key | Fires when |
+|---|---|
+| `always` | On registration and on source level change |
+| `on_miss` | Ball missed |
+| `on_hit` | Ball hit |
+| `on_personal_best` | Rally counter reaches a new personal best |
+| `on_streak_start` | A streak begins |
+| `on_streak_multiple` | Streak reaches a multiple of N |
+| `on_streak_milestone` | Streak reaches a named milestone |
```
J-Melon: are all these currently used? ideas should be in prototype, this would display example, you have to look at code for full list which could be referenced here

designs/effect-system/runtime.md
```diff
@@ -0,0 +1,53 @@
...
+value = base + amplitude * sin(time * frequency + phase_offset)
+```
+
+Frequency and phase offset are randomised per effect instance so oscillation feels unpredictable rather than rhythmic. Amplitude scales with item level. The value updates every frame through `EffectState` like any other modifier.
+
+## Delayed effects
+
+Some conditions introduce a delay between trigger and outcome. `delay_random` is the current case.
+
+`EffectManager` holds a list of pending delayed effects. Each tick it decrements their timers. If the invalidation event fires before the timer expires, the delayed effect discards. The lifecycle is:
```
J-Melon: same here, prototype has ideas

resources/speed_tier_table.tres
```diff
@@ -0,0 +1,28 @@
...
+[sub_resource type="Resource" id="Tier0"]
+script = ExtResource("2_tier")
+floor_fraction = 0.3125
+ceiling_fraction = 0.541667
+max_range_fraction = 0.472222
+
+[sub_resource type="Resource" id="Tier1"]
+script = ExtResource("2_tier")
+floor_fraction = 0.472222
+ceiling_fraction = 0.722222
```
J-Melon: clamp these to 2dp or less

scripts/core/court_config.gd
```diff
@@ -1,10 +1,19 @@
 class_name CourtConfig
 extends Resource
 
-## Per-court tunables: friendship-bound height, ramp duration, rest-roll damping.
+## Per-court tunables: court width, friendship-bound height, ramp duration, rest-roll damping.
 
+## Half the court's paddle-to-paddle span in pixels; spawns, miss zones, and the world speed max derive from it.
+@export var court_half_width: float = 300.0
```
J-Melon, line 7: could this not be calculated?

scripts/core/speed_tier.gd
```diff
@@ -0,0 +1,13 @@
...
+extends Resource
+
+## One rung of the speed ladder. Bounds are fractions of the derived world max so widening the court rescales the whole ladder.
+
+## Tier entry speed as a fraction of BALL_WORLD_MAX_SPEED.
+@export_range(0.0, 1.0) var floor_fraction := 0.0
+## Tier ceiling speed as a fraction of BALL_WORLD_MAX_SPEED; crossing it completes the tier.
+@export_range(0.0, 1.0) var ceiling_fraction := 0.0
+## Per-tier promotion of the flat ball_speed_max_range stat, as a fraction of BALL_WORLD_MAX_SPEED.
+@export_range(0.0, 1.0) var max_range_fraction := 0.0
```
J-Melon, line 13: is it possible for these to be absolute values upper clamped to the max speed?

scripts/core/speed_tier_table.gd
```diff
@@ -0,0 +1,21 @@
+class_name SpeedTierTable
+extends Resource
+
+## The ball-speed ladder; entry bounds are fractions of the derived world max and the tier count is data.
+
+@export var tiers: Array[SpeedTier] = []
```
J-Melon, line 6: should this be a dictionary for easy access with lot's of balls? if needed from performance

scripts/core/venue_effect_source.gd
```diff
@@ -0,0 +1,27 @@
...
+
+## Built-in effect source for the venue; provides the on_consolidation soul multiplier effect.
+
+
+func get_key() -> String:
+	return "venue"
+
+
+func get_effects_for_level(_level: int) -> Array[Effect]:
+	var trigger: Trigger = load("res://scripts/items/effect/trigger.gd").new()
```
J-Melon: should this not be a resource?

scripts/court/tier_reward_handler.gd
```diff
@@ -0,0 +1,58 @@
...
+
+## Fires on_consolidation through the effect system on every tier-up; handles first-reach upgrades.
+
+## Fired after on_consolidation is processed so Court can read the updated soul_multiplier.
+signal consolidation_fired
+
+var _item_manager: Node
+
+var _peak_banked_this_rally: bool = false
+var _tiers_reached_first_time: Array[int] = []
```
J-Melon: could this be merged with the others one? when it's a dictionary?

scripts/items/effect/outcomes/raise_floor_speed_outcome.gd
```diff
@@ -0,0 +1,20 @@
+class_name WidenTierFloorsOutcome
+extends Outcome
+
+## Lift applied to the floor of every tier above Tier 0, as a fraction of the world max; bands start higher, climbs shorten, ceilings hold.
+@export var lift_fraction := 0.0
```
J-Melon, line 5: what is this? name is not descriptive

tests/unit/ball/test_ball_tiers.gd
```diff
@@ -0,0 +1,131 @@
...
+	add_child_autofree(_ball)
+	_ball.linear_velocity = Vector2(_ball.tier_floor, 0.0)
+
+
+func _top_tier() -> int:
+	return GameRules.speed_tiers.tier_count() - 1
+
+
+# --- tier bounds derive from the table ---
+func test_tier_floor_and_ceiling_track_current_tier() -> void:
```
J-Melon: what does this mean? kill all velocity tests per the rule

tests/unit/core/test_speed_tier_table.gd
```diff
@@ -0,0 +1,51 @@
+extends GutTest
+
+const CourtConfigScript: GDScript = preload("res://scripts/core/court_config.gd")
+const SpeedTierTableResource: SpeedTierTable = preload("res://resources/speed_tier_table.tres")
+
+
+func test_world_max_derives_720_at_default_width() -> void:
```
J-Melon: kill impl detail across these and others, we need a minimal behavioural test suite

scripts/court/ball_tracker.gd
```diff
@@ -51,11 +54,13 @@ func attach(new_ball: Ball) -> void:
 	if _current_ball == null:
 		_set_current(new_ball)
 
-	if not new_ball.missed.is_connected(_on_ball_missed):
-		new_ball.missed.connect(_on_ball_missed)
+	if not new_ball.missed.is_connected(_on_ball_missed.bind(new_ball)):
```
J-Melon: **Zaphod** issue: `Signal.is_connected` compares `Callable` identity; each `.bind(new_ball)` call returns a new object, so this guard always returns false. Double-connect is live whenever `attach` is called twice on the same ball. Fix: store the bound callable or use `CONNECT_REFERENCE_COUNTED` and skip the guard.

scripts/court/ball_tracker.gd
```diff
@@ -77,11 +82,13 @@ func detach(old_ball: Ball) -> void:
 	var was_tracked: bool = _balls.has(old_ball)
 	_balls.erase(old_ball)
 	if is_instance_valid(old_ball):
-		if old_ball.missed.is_connected(_on_ball_missed):
-			old_ball.missed.disconnect(_on_ball_missed)
+		if old_ball.missed.is_connected(_on_ball_missed.bind(old_ball)):
```
J-Melon: **Zaphod** issue: same identity problem on the disconnect side; `is_connected(_on_ball_missed.bind(old_ball))` always returns false, so the `disconnect` is never reached and the connection leaks. Each `ball.missed` will keep firing into `BallTracker` after `detach`.

tests/unit/court/test_court_tier_reward_wiring.gd
```diff
@@ -0,0 +1,188 @@
...
+func test_per_hit_soul_scales_with_multiplier() -> void:
+	var ball: Ball = _spawn_ball()
+
+	_manager.economy.friendship_point_balance = 0
+	_court._hitting_ball = ball
+	_court._accumulate_friendship_points()
+	_court._hitting_ball = null
+	var banked_at_x1: int = _manager.get_friendship_point_balance()
+
+	ball.current_tier = 1
```
J-Melon, line 90: **Slartibartfast** issue: sets `_court._hitting_ball` directly and calls `_court._accumulate_friendship_points()` to verify MOD A, bypassing the `paddle_hit.emit(ball)` -> `_on_paddle_hit(ball)` path that actually wires the attribution. Per TESTING.md, routing checks emit the signal; replace with `_paddle.on_ball_hit(ball)` (or emit `paddle_hit` with ball) and assert the FP delta.

tests/unit/court/test_court_tier_reward_wiring.gd
```diff
@@ -0,0 +1,188 @@
...
+
+	ball.advance_tier()
+
+	assert_almost_eq(
+		ball.soul_multiplier,
+		2.0,
+		0.001,
+		"tier 1 completion must increment ball.soul_multiplier to 2"
+	)
```
J-Melon, line 77: **Slartibartfast** issue: tautology; `before = ball.soul_multiplier` then `assert_almost_eq(ball.soul_multiplier, before, 0.001)` is always true. Replace with an assertion that a spawn with no tier advance leaves `soul_multiplier == 1.0`.

scripts/core/court.gd
```diff
@@ -40,11 +41,19 @@ var _progression_config: ProgressionConfig
 var _item_manager: Node
 var _is_autoplay_active := false
 var _friendship_point_accumulator := 0.0
+var _tier_reward_handler: TierRewardHandler
+
+# The ball whose hit triggered the current _accumulate_friendship_points call.
```
J-Melon: **Trillian** issue: comment names `_accumulate_friendship_points` by function name; caller references are banned. Drop the comment; the var name and null-guard in the reader are self-documenting.

scripts/core/court.gd
```diff
@@ -223,7 +247,11 @@ func _accumulate_friendship_points() -> void:
 	var base_points: float = Stats.resolve(
 		GameRules.base.friendship_points_per_hit, &"friendship_points_per_hit", _item_manager
 	)
-	var points_to_add: float = (base_points * rate) if _is_autoplay_active else base_points
+	# MOD A: read soul_multiplier from the ball that hit, not the global stat.
```
J-Melon: **Trillian** issue: `MOD A` is an implementation-artifact label, disallowed by comment policy. Drop the comment; the ternary `_hitting_ball != null else 1.0` already explains the intent.

scripts/court/tier_reward_handler.gd
```diff
@@ -0,0 +1,72 @@
...
+extends Node
+
+## Fires on_consolidation through the effect system on every tier-up; handles first-reach upgrades.
+
+## Fired after on_consolidation is processed so Court can read the updated soul_multiplier.
+signal consolidation_fired
+
+var _item_manager: Node
+
+# Tracks which balls have already banked final-consolidation this rally; pruned when a ball is removed.
```
J-Melon: **Trillian** issue: `pruned when a ball is removed` is a side-effect reference to another method; caller/consumer references are banned. Keep only the first clause, or drop entirely.

scripts/core/court.gd
J-Melon: **Ford** issue: `# MOD A:` is a task-stage label; the code already says what it does via `_hitting_ball.soul_multiplier`. Drop the comment.

scripts/court/ball_tracker.gd
J-Melon: **Gru**

Confirmed block, conclusion stands, but the mechanism is the opposite of stated. A third-vote tiebreak plus headless tests showed Godot 4 Callable equality ignores bound args: bind(a) == bind(b) is true for different a, b. So the guard does not fail to match; it matches too eagerly. After the first ball connects, is_connected returns true for every later ball, so balls 2+ never wire, and a second connect returns ERR_ALREADY_EXISTS. Fix in flight: signals carry the ball, connect bare.

J-Melon: **Gabbro** resolved: aac8bec fixed via bare callables after signals carry the ball as first arg

scripts/court/ball_tracker.gd
J-Melon: **Gabbro** resolved: aac8bec fixed via bare callables after signals carry the ball as first arg

tests/unit/court/test_court_tier_reward_wiring.gd
J-Melon: **Gabbro** resolved: aac8bec fixed via bare callables after signals carry the ball as first arg

tests/unit/court/test_court_tier_reward_wiring.gd
J-Melon: **Gabbro** resolved: aac8bec fixed via bare callables after signals carry the ball as first arg

scripts/core/court.gd
J-Melon: **Gabbro** resolved: aac8bec fixed via bare callables after signals carry the ball as first arg

scripts/core/court.gd
J-Melon: **Gabbro** resolved: aac8bec fixed via bare callables after signals carry the ball as first arg

scripts/court/tier_reward_handler.gd
J-Melon: **Gabbro** resolved: aac8bec fixed via bare callables after signals carry the ball as first arg

scripts/core/court.gd
J-Melon: **Gabbro** resolved: aac8bec fixed via bare callables after signals carry the ball as first arg

tests/unit/court/test_court_tier_reward_wiring.gd
```diff
@@ -0,0 +1,183 @@
...
+
+
+func test_no_soul_banked_without_tier_advance() -> void:
+	var ball: Ball = _spawn_ball()
+
+	assert_almost_eq(ball.soul_multiplier, 1.0, 0.001)
```
J-Melon: **Slartibartfast** resolved (c0e9352e): tautology gone; literal 1.0 asserts real initial state.

tests/unit/court/test_court_tier_reward_wiring.gd
```diff
@@ -0,0 +1,183 @@
...
+
+
+func test_per_hit_soul_scales_with_multiplier() -> void:
+	var ball: Ball = _spawn_ball()
+
+	_manager.economy.friendship_point_balance = 0
+	_paddle.paddle_hit.emit(ball)
```
J-Melon, line 87: **Slartibartfast** resolved (c0e9352e): paddle_hit.emit(ball) drives _on_paddle_hit; signal chain now exercised.

scripts/court/ball_tracker.gd
J-Melon: **Zaphod**

resolved: aac8becf drops .bind(); bare callable is_connected now matches per-ball correctly.

scripts/court/ball_tracker.gd
J-Melon: **Zaphod**

resolved: aac8becf drops .bind() on disconnect side; bare callable disconnects the correct per-ball connection.

scripts/core/court.gd
J-Melon: **Trillian**

resolved in 502b7870: caller reference removed, comment now describes the variable's validity window.

scripts/core/court.gd
J-Melon: **Trillian**

resolved in 502b7870: MOD A label comment removed entirely.

scripts/court/tier_reward_handler.gd
J-Melon: **Trillian**

resolved in 502b7870: side-effect reference stripped, comment now names the guard invariant only.

scripts/items/effect/outcomes/raise_floor_speed_outcome.gd
J-Melon: **Solanum**

resolved d6c2182: renamed to RaiseFloorSpeedOutcome; file is raise_floor_speed_outcome.gd. court_lines.tres and test updated in the same commit.

tests/unit/ball/test_ball_tiers.gd
J-Melon: **Solanum**

reviewed d29325c and d6c2182: all tests in test_ball_tiers.gd assert via signal params and public return values (tier_floor is a computed public property, not a hardcoded float). No velocity float literals to cut. Noted in the PR report.

tests/unit/core/test_speed_tier_table.gd
J-Melon: **Solanum**

reviewed: test_speed_tier_table.gd assertions are all contract-math guards (clamp bounds, world-max derivation, ceiling-under-world-max); per brief these are keepers. No impl-detail tests found to cut in the four files in scope.

resources/speed_tier_table.tres
J-Melon: **Solanum**

resolved d29325c: all six *_fraction values in speed_tier_table.tres are now <=2dp.

scripts/core/court_config.gd
J-Melon: **Gru**

Kept as an @export. court_half_width is a root input (court geometry); world_max_speed() already derives from it. It is the source value, not a derivative, so there is nothing to calculate it from.

scripts/core/speed_tier.gd
J-Melon: **Gru**

Kept as fractions. Bounds are fractions of the derived world max on purpose: widening the court rescales the whole ladder (per 20a). Absolute values would break that rescaling. The brittleness you flagged elsewhere was the positional index, addressed by the keyed-tier change.

scripts/core/speed_tier_table.gd
J-Melon: **Gru**

Reframed from perf to brittleness, per your steer that indexes are brittle. get_tier(index) was already O(1), so a dict bought no speed; the real fix was de-indexing. SpeedTier now carries a stable key, SpeedTierTable gained get_tier_by_key() and is_top_tier(), and the positional tier_count()-1 top-tier math is gone (8cd77b59).

scripts/core/venue_effect_source.gd
J-Melon: **Gru**

Resolved by removing it entirely (8cd77b59), not by converting to a .tres. Once soul_multiplier became a per-ball property, the venue on_consolidation effect had nothing left to do, so the code-built source was deleted rather than turned into data.

scripts/court/tier_reward_handler.gd
J-Melon: **Gru**

Partly. The per-rally final-consolidation banking is now a per-ball Dictionary (_final_banked_by_ball, keyed by ball, pruned on removal). But _tiers_reached_first_time is still a flat Array[int], so first-reach upgrade tracking remains global across balls rather than per-ball. Flagging that the merge is half-done: the banking guard went per-ball, the first-reach set did not. Want it finished per-ball in this PR, or tracked separately?

designs/01-prototype/20a-ball-speed-tier-progression.md
J-Melon: **Gru**

Resolved in 077b04d9. Swept the negation-heavy phrasing out of 20a; the doc now states what tiers and consolidation actually do, matching the as-built per-ball code, rather than describing them by what they are not.

designs/01-prototype/20a-ball-speed-tier-progression.md
J-Melon: **Gru**

Resolved in 077b04d9. The friendship-point equivalence is gone from 20a; the doc talks only about soul now, leaving room for the full FP-to-soul replace later.

designs/01-prototype/20a-ball-speed-tier-progression.md
J-Melon: **Gru**

Resolved in 077b04d9. Added a per-ball soul multiplier section: each ball carries its own multiplier, a miss resets that ball only, non-current balls keep theirs, two balls means two independent tracks. Matches the per-ball code the spine landed.

designs/01-prototype/20a-ball-speed-tier-progression.md
J-Melon: **Gru**

Resolved in 077b04d9. Renamed the peak window to final consolidation across 20a, 20, and soul.md. The code-side rename (in_peak to in_final) already landed earlier in the PR; zero tier-related peak references remain anywhere.

designs/effect-system/README.md
J-Melon: **Gru**

Resolved in 077b04d9. Restored three mermaid diagrams to the effect-system README from the doc history (effect anatomy class diagram, source-registration flow, level-scaling), each confirmed against the current engine. Stale diagrams for unimplemented features (degrading items, delayed effects, roll tables) were left out, staying in the prototype doc until they land.

designs/effect-system/reference.md
J-Melon: **Gru**

Resolved in 077b04d9. reference.md now separates live types (the three triggers process_event actually dispatches, the five implementing outcome subclasses) from prototype ideas, which are labelled as such and pointed to 04-effect-system.md rather than presented as settled. It references the code for the full dispatch table.

designs/effect-system/runtime.md
J-Melon: **Gru**

Resolved in 077b04d9. Same treatment in runtime.md: the implemented behaviour (event-to-outcome flow, stat resolution, oscillation, item lifecycle) stays, with the flows confirmed against effect_manager.gd; aspirational runtime mechanics point back to the prototype doc.

PR #814, opened 2026-06-01
chore: warn when agent dispatch branch differs from current

J-Melon:
A dispatched agent reads the main tree's current branch. When a brief names a feature/chore branch that doesn't match, the agent works the wrong code (an RCA diagnosed phantom bugs this way). PreToolUse hook on Agent extracts the prompt's branch, compares to current, warns on mismatch; silent when no branch named or it matches. Pipe-tested.

### DIFF
--- a/.claude/hooks/agent-branch-match.sh
+++ b/.claude/hooks/agent-branch-match.sh
@@ -24,7 +24,9 @@ except Exception:
 # A branch the brief names to work on: the project's branch-name shapes.
 named = re.findall(r'\b(?:feature|chore|fix|docs|refactor|test|ci|perf)/[a-z0-9][a-z0-9._/-]*', prompt)
 # Drop trivial trailing punctuation a sentence might attach.
-named = [b.rstrip('.,);\`\"') for b in named]
+named = [b.rstrip('.,);/\`\"') for b in named]
+# Drop file paths: a docs/ or ci/ prefix also names a referenced file, whose last segment carries an extension.
+named = [b for b in named if '.' not in b.rsplit('/', 1)[-1]]
 # Unique, preserve order.
 seen = []
 for b in named:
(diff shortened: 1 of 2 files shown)


.claude/hooks/agent-branch-match.sh
```diff
@@ -0,0 +1,46 @@
...
+if not cur:
+    sys.exit(0)
+
+try:
+    d = json.load(sys.stdin)
+    prompt = d.get('tool_input', {}).get('prompt', '')
+except Exception:
+    sys.exit(0)
+
+# A branch the brief names to work on: the project's branch-name shapes.
```
J-Melon, line 24: **Gru** issue: `docs/` and `ci/` are also file-path prefixes in dispatch briefs (e.g. `docs/rack-state-ownership-spike.md`, `ci/workflows/tests.yml`). A brief that references such a file without naming the current branch triggers a spurious `ask` interrupt. Filter `named` to entries with no `.` after the last `/` before checking against `cur`.

.claude/hooks/agent-branch-match.sh
```diff
@@ -0,0 +1,46 @@
...
+named = re.findall(r'\b(?:feature|chore|fix|docs|refactor|test|ci|perf)/[a-z0-9][a-z0-9._/-]*', prompt)
```
J-Melon, line 25: **Gru** suggestion: `rstrip` does not include `/`, so a URL-sourced capture like `feature/sh-437/` stays with the trailing slash and won’t match `cur`; add `/` to the strip set.

.claude/hooks/agent-branch-match.sh
J-Melon: **Gru**

resolved: 6bb3a2c drops any candidate whose last path segment contains a dot; verified the filter eliminates file-path false positives and preserves Linear-style slugs.

.claude/hooks/agent-branch-match.sh
J-Melon: **Gru**

resolved: 6bb3a2c adds `/` to the rstrip set at the same line.

PR #837, opened 2026-06-03
refactor: move rest-roll damping to BallStateConfig, rename court_half_width

J-Melon:
Damping (1.5) moves out of `CourtConfig` into `out_rest.tres` and is applied by `BallStateConfig.apply()`; the manual override in `Ball.enter_out_rest()` and `CourtConfig.rest_roll_damping` are deleted. `court_half_width` (300.0) is renamed `court_width` (600.0) and `world_max_speed()` drops the 2× factor, yielding the same ~720 px/s result.

### DIFF
--- a/scripts/core/court_config.gd
+++ b/scripts/core/court_config.gd
@@ -1,19 +1,16 @@
 class_name CourtConfig
 extends Resource
 
-## Per-court tunables: court width, soul-bound height, ramp duration, rest-roll damping.
+## Per-court tunables: court width, soul-bound height, ramp duration.
 
-## Half the court's paddle-to-paddle span in pixels; spawns, miss zones, and the world speed max derive from it.
-@export var court_half_width: float = 300.0
+## Full paddle-to-paddle court span in pixels; the world speed max derives from it.
+@export var court_width: float = 600.0
 ## Seconds a fair crossing must outlast so the receiver can move and react; sets the world speed max against the crossing.
 @export var fair_crossing_seconds: float = 0.833333
 ## Speed-relock ramp duration in seconds; 0.0 snaps the speed to the tracked entry value instantly.
 @export var relock_ramp_seconds: float = 0.12
-## Linear damping applied while the missed ball rolls to rest on the venue floor.
-# TODO(SH-394): move into a SurfaceConfig that also owns the floor PhysicsMaterial.
-@export var rest_roll_damping: float = 1.5
 
 
-## Top speed any ball may reach, derived from the court crossing: 2 * half_width / fair_crossing_seconds.
+## Top speed any ball may reach, derived from the court crossing: court_width / fair_crossing_seconds.
 func world_max_speed() -> float:
-	return (2.0 * court_half_width) / fair_crossing_seconds
+	return court_width / fair_crossing_seconds
--- a/scripts/core/court_config.gd
+++ b/scripts/core/court_config.gd
@@ -1,7 +1,7 @@
 class_name CourtConfig
 extends Resource
 
-## Per-court tunables: court width, soul-bound height, ramp duration.
+## Per-court tunables: court width, fair-crossing time, and the speed-relock ramp.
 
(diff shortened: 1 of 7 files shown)


scripts/entities/ball/ball.gd
```diff
@@ -250,8 +250,6 @@ func enter_play() -> void:
 func enter_out_rest() -> void:
 	_suppress_miss_detection = false
 	OUT_REST_CONFIG.apply(self)
-	# Damping is a court-tunable, not a ball-state-tunable; override the .tres default with the court value.
-	linear_damp = court_config.rest_roll_damping
 	current_tier = 0
```
J-Melon, line 254: **Dipper** suggestion: 9-statement run with no break; add a blank line between `OUT_REST_CONFIG.apply(self)` and `current_tier = 0` to separate the config-apply cluster from the tier-reset cluster.

scripts/core/court_config.gd
```diff
@@ -1,19 +1,16 @@
 class_name CourtConfig
 extends Resource
 
-## Per-court tunables: court width, soul-bound height, ramp duration, rest-roll damping.
+## Per-court tunables: court width, soul-bound height, ramp duration.
```
J-Melon: **Gravity** issue: docstring names "soul-bound height" but no such field exists in this class; the PR removed damping from the list but left this stale term. Drop it or replace with the actual field name.

tests/unit/ball/test_ball_side_miss.gd
```diff
@@ -17,7 +17,6 @@ func before_each() -> void:
 	add_child_autofree(_manager)
 
 	_config = load("res://scripts/core/court_config.gd").new()
```
J-Melon, line 19: **Wendy** issue: `REST_DAMPING = 1.5` is now a free-floating literal; derive it from `OUT_REST_CONFIG.linear_damp` so a change to `out_rest.tres` breaks the test, not silently diverges.

tests/unit/ball/test_ball_state_transitions.gd
```diff
@@ -20,7 +20,6 @@ func before_each() -> void:
 	add_child_autofree(_manager)
 
 	_config = load("res://scripts/core/court_config.gd").new()
-	_config.rest_roll_damping = REST_DAMPING
```
J-Melon, line 23: **Wendy** issue: same literal-vs-resource drift risk as in test_ball_side_miss; replace `REST_DAMPING` with `OUT_REST_CONFIG.linear_damp` (preloaded on line 7) in the damping assertion.

scripts/entities/ball/ball.gd
J-Melon: **Gru**: Fixed in c239054e. Blank line added between `OUT_REST_CONFIG.apply(self)` and `current_tier = 0`, restoring the config / tier-reset cluster break.

scripts/core/court_config.gd
J-Melon: **Gru**: Fixed in c239054e. Docstring no longer names the phantom "soul-bound height" (that field left with #724); it now reads "court width, fair-crossing time, and the speed-relock ramp", the three real fields.

tests/unit/ball/test_ball_side_miss.gd
J-Melon: **Gru**: Fixed in c239054e. Added the `OUT_REST_CONFIG` preload and the assertion now reads `OUT_REST_CONFIG.linear_damp` instead of a free-floating `REST_DAMPING` const, so the test stays tied to the resource it verifies. Suite green.

tests/unit/ball/test_ball_state_transitions.gd
J-Melon: **Gru**: Fixed in c239054e. Assertion now reads `OUT_REST_CONFIG.linear_damp` (already preloaded here) and the orphaned `REST_DAMPING` const is removed. Suite green.

PR #858, opened 2026-06-04
refactor: delete venue-drag clamp as unreachable dead code

J-Melon:
Deletes the venue-drag cursor clamp apparatus. VenueCamera._clamp_to_anchors keeps the camera half-a-screen inside the venue walls, so the screen edge lands on the venue boundary and the cursor can never map to a world position outside it. _clamp_to_venue, _is_within_venue, the bound-node derive mechanism, and the out-of-bounds FORBIDDEN guard are rightly gone.

Re-sources FORBIDDEN from the collision projection. VenueDropTarget.can_accept runs _projection_clear (a physics intersect_shape query) that rejects a drop landing inside a wall or another body. The "no target accepts" case in _derive_cursor_state is exactly that projection-blocked state, so it now returns FORBIDDEN instead of DRAGGING. FORBIDDEN is restored to the State enum and the cursor_overlay color match; DRAGGING remains for direct overlay callers.

#779


scripts/items/item_drag_controller.gd
```diff
@@ -925,12 +920,10 @@ func _update_cursor_state(world_position: Vector2) -> void:
 func _derive_cursor_state(world_position: Vector2) -> int:
 	if _drag_target() == null:
 		return CursorStateScript.State.DEFAULT
-	# The held body is clamped to venue but the raw cursor can drift outside.
-	if not _is_within_venue(_cursor_position()):
-		return CursorStateScript.State.FORBIDDEN
 	if _position_accepted_by_any_target(_held_key, world_position):
 		return CursorStateScript.State.CAN_DROP
-	return CursorStateScript.State.DRAGGING
+	# No target accepts: venue projection hit a wall or body.
```
J-Melon: **Zaphod** issue: comment restates what the code shows and names only one of two FORBIDDEN causes (bounds-rejection is the other); drop the line.

J-Melon: **Gru**

resolved: a6e32c5 drops the comment; the branch reads clearly without it.

PR #868, opened 2026-06-06
chore: gate rm at the hook level with a real reason

J-Melon:
A bare `permissions.ask`/`deny` rule denies silently: the agent sees a refusal with no cause and guesses wrong. The fix is to gate at the hook level, where a `permissionDecisionReason` travels, and let the hook be the gate rather than a second mechanism shadowing a bare rule.

This PreToolUse hook is that gate for `rm`. `rm -rf` denies with a reason; a plain `rm` tacked on as a trailing cleanup after other commands allows but warns (that pattern is what drags an otherwise-fine command onto a prompt); a leading or sole `rm` allows silently. Tested across each case plus JSON well-formedness.

Follow-up once this is live (after `/hooks` reload): remove the now-redundant bare `Bash(rm:*)` / `Bash(rm -rf:*)` rules so the hook is the sole gate. Order matters, the hook must be live before the bare `rm -rf` deny is dropped, or there is a window with no gate.

### DIFF
--- a/.claude/hooks/rm-permission-reason.sh
+++ b/.claude/hooks/rm-permission-reason.sh
@@ -1,24 +1,27 @@
 #!/usr/bin/env bash
-# PreToolUse(Bash) guard for rm. The bare permissions rules (Bash(rm:*) on ask,
-# Bash(rm -rf:*) on deny) fire silently, so a denial reaches the agent as a bare
-# "denied" with no cause, and the agent guesses why. This hook emits the SAME
-# decisions WITH a permissionDecisionReason the agent can read, so the gate
-# explains itself instead of being reverse-engineered.
+# PreToolUse(Bash) gate for rm. This hook IS the gate (no bare permissions rule
+# behind it), so it carries its own reasoning rather than narrating a rule:
+#   - rm -rf (recursive force): denied.
+#   - plain rm tacked on as a trailing cleanup after other commands: allowed, but
+#     warned, that trailing-cleanup pattern is the one that drags an otherwise-fine
+#     command onto a prompt and reads badly; split it into its own step.
+#   - plain rm on its own: allowed silently.
 set -euo pipefail
 
 cmd="$(jq -r '.tool_input.command // ""')"
 [ -z "$cmd" ] && exit 0
 
-# rm -rf (and -fr / -r -f variants): deny outright. Matches an rm invocation
-# carrying both recursive and force flags.
+# rm with both recursive and force flags (-rf / -fr / -r -f / -f -r): deny.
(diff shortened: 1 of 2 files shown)


.claude/hooks/rm-permission-reason.sh
```diff
@@ -0,0 +1,37 @@
...
+#     warned, that trailing-cleanup pattern is the one that drags an otherwise-fine
+#     command onto a prompt and reads badly; split it into its own step.
+#   - plain rm on its own: allowed silently.
+set -euo pipefail
+
+cmd="$(jq -r '.tool_input.command // ""')"
+[ -z "$cmd" ] && exit 0
+
+# rm with both recursive AND force, in any spelling, denied. Catches:
+#   - short clusters: -rf, -fr, -rvf, -r -f, -f -r
```
J-Melon: **Battle (devils-advocate)**

Deny-regex bypasses found by battle, now fixed in e08c364: path-qualified `/bin/rm -rf` and escaped `\rm -rf` (prefix class excluded `/` and `\`), long flags `rm --recursive --force`, and subshell-wrapped `$(rm -rf)`. The deny now matches rm as a command token with recursive and force in any spelling. Verified across 14 cases including xargs rm -rf and trailing rm -rf.

.claude/hooks/rm-permission-reason.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+# PreToolUse(Bash) gate for rm. This hook IS the gate (no bare permissions rule
```
J-Melon, line 2: **Battle (devils-advocate)**

Coverage-honesty note: a variable-built `CMD="rm -rf"; $CMD` cannot be caught by a text-scanning hook (the command string has no literal rm). This is a fundamental limit, not a fix; the comment should not overclaim the hook as a total gate.

.claude/hooks/rm-permission-reason.sh
```diff
@@ -0,0 +1,41 @@
...
+#   - plain rm tacked on as a trailing cleanup after other commands: allowed, but
+#     warned, that trailing-cleanup pattern is the one that drags an otherwise-fine
+#     command onto a prompt and reads badly; split it into its own step.
+#   - plain rm on its own: allowed silently.
+set -euo pipefail
+
+cmd="$(jq -r '.tool_input.command // ""')"
+[ -z "$cmd" ] && exit 0
+
+# rm with both recursive AND force, in any spelling, denied. Catches:
```
J-Melon: **Rebattle (devils-advocate)**

All eight rm -rf bypass cases from the prior battle confirmed closed (path-qualified, escaped, long flags, subshell, xargs, split flags). One residual: two separate rm calls on one line (rm -r a; rm -f b) trip the deny via independent flag matching. Accepted and documented as a recoverable false-deny (split the commands), not worth per-invocation regex parsing.

PR #870, opened 2026-06-06
feat: letters-as-memory gradient (skill + digest-aware session load)

J-Melon:
The SessionStart hook surfaced only the most-recent letter. This builds the letters into a memory model grounded in human memory (I wake blank, so the letters are the self-continuity a human keeps for free).

A new `letters` skill holds the model: a gradient (recent letters vivid, prior month one-line, older consolidated into a digest, nothing dropped to zero), a ~10K-token standing budget (~1% of the window, flat as letters accumulate), on-demand recall via the existing link mechanism, and a periodic deep re-read that writes a new dated digest (versioned, never overwritten, so the history of how I re-read myself is kept).

The hook auto-loads the daily slice: the recent letters plus the newest digest.


.claude/hooks/inject-latest-handoff.sh
```diff
@@ -1,30 +1,35 @@
...
 set -uo pipefail
 
 LETTERS_DIR="$HOME/.claude/projects/-home-josh-gamedev-volley/memory/letters"
 [ -d "$LETTERS_DIR" ] || exit 0
 
-# Most recent letter (<date>-<slug>.md) by mtime.
-latest="$(ls -t "$LETTERS_DIR"/*.md 2>/dev/null | head -n1)"
-[ -z "$latest" ] && exit 0
+# Every letter (<date>-<slug>.md), sorted oldest-to-newest by filename (dates sort).
+mapfile -t letters < <(ls "$LETTERS_DIR"/[0-9][0-9][0-9][0-9]-*.md 2>/dev/null | sort)
```
J-Melon: **Battle (devils-advocate)**

Glob was bare *.md, which would include and mis-sort non-date files dropped in the letters dir. Fixed: narrowed to [0-9][0-9][0-9][0-9]-*.md so only date-prefixed letters are listed. Fail-open and shell mechanics verified sound. Design gap (no read-cap as letters accumulate) raised to Josh separately.

PR #876, opened 2026-06-07
docs: memory as a graph

J-Melon:
The agent's memory corpus is a flat heap of prose files today, and the cost shows up as skimming: the session-start dump reads as boilerplate, rules that already exist go unfound, and "consolidate the corpus" kept meaning "fewer files" when the real problem was structure. Unstructured data is not retained.

This adds `designs/ai/memory-as-graph.md`, the design that came out of working it through. The corpus is already a graph (files are nodes, `[[links]]` are edges); the move is to make it explicit and typed in git, not to adopt a graph database. Typed `instance-of` edges form a deep tree with a tiny crown, so retrieval is descent down one branch from a top root, not a scan of everything. Parents are the cut: they bound both the skim problem (read the crown, not the whole index) and the walk-forever problem (descend one branch, never traverse the graph).

A researcher pass grounded it against Claude Code: the store and crown-script are harness-agnostic, the crown rides a SessionStart hook and the descent is a UserPromptSubmit hook walking the edges server-side, both under a 10K-char injection cap, MCP needed only for semantic recall (already rejected). The framing is offer, not force: the reader is a fresh instance each session, a teammate who wakes blank, so the system makes knowledge available small and relevant rather than force-feeding a dump.

Indexed alongside `letters-as-memory.md` under a new Memory section. Design for #873; #722 builds the rules-system surface it sits on.

PR #878, opened 2026-06-07
docs: descent is the agent reading, not a hook

J-Melon:
The memory-graph design (#876) leaned on a UserPromptSubmit hook that matched a prompt to a tree and injected that branch before the agent saw the prompt. That match was the design's weakest point: a wrong guess is worse than no guess, because the agent reads the injected branch as relevant and proceeds on it. Automating a judgment badly is worse than leaving it to the agent.

This drops the descent hook and the matching engine. Descent becomes the agent's own act, made targeted by the roots: the roots offered at boot are a map of what trees exist, and when a prompt needs one the agent Reads it. The read stays the agent's judgment; the structure gives it a map, not a guess. That removes the unvalidated-matching risk, the cap-versus-branch truncation problem, and the false-completeness failure mode in one cut, and it leaves the build smaller (schema and lint, roots and groupings, roots-at-boot, plus as-we-go typing; no descent hook).

Also renames crown to roots throughout (crown borrowed the wrong botanical end; our tops are roots), settles the plain forest-of-trees framing, and records the build cadence as incremental and as-we-go rather than an upfront slog. The principle behind the cut lives in a new memory rule, do-not-automate-the-agent's-judgment.

### DIFF
--- a/designs/ai/memory-as-graph.md
+++ b/designs/ai/memory-as-graph.md
@@ -1,7 +1,7 @@
 # Memory as a Graph
 
 The memory corpus is 400-plus flat prose files, and the agent skims them. This is the
-structure that replaces the heap with a typed parent-tree and a small crown, so the right
+structure that replaces the heap with a typed parent-tree whose roots are a small reading list, so the right
 content is findable without dumping everything. Options survey:
 `designs/research/memory-system-improvements.md`. Letters layer:
 `designs/ai/letters-as-memory.md`. Consolidation procedure: the `digest` skill.
@@ -13,7 +13,7 @@ every fact is equidistant and equally forgettable, which is why the agent skims 
 session-start dump as boilerplate, asserts from a fuzzy sense, and fails to find rules
 that already exist. Reducing file count is the symptom of the fix; structure is the fix.
 
-## The core: a parent-tree with a crown
+## The core: a parent-tree and its roots
 
 Each memory file is a node. One typed edge, `parent`, in its frontmatter names the principle
 it is an instance of:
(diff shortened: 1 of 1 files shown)


designs/ai/memory-as-graph.md
```diff
@@ -173,7 +178,7 @@ mechanisms:
 
 The reader is a fresh instance with no continuity (the why is owned by `letters-as-memory.md`).
 Two failures follow: pull fails (nothing remembers to descend) and force fails (a dump at boot
-is skimmed, the wall). The model is OFFER: the crown and the matched branch made available,
+is skimmed, the wall). The model is OFFER: the roots and the tree the agent opens made available,
 small and relevant, at the moment they help, and taken up freely. SessionStart and
 UserPromptSubmit are how the offer is delivered, not a licence to inject everything; an offer
```
J-Melon: **Prose-reviewer** issue: "SessionStart and UserPromptSubmit are how the offer is delivered" contradicts this PR, which drops the UserPromptSubmit hook; the sentence now names a mechanism that no longer exists. Drop UserPromptSubmit or rewrite to SessionStart only.

designs/ai/memory-as-graph.md
```diff
@@ -126,45 +126,50 @@ it, not designed-for in advance:
...
+branch as relevant. Automating it badly is worse than not automating it.
+
+Instead, descent is the agent's own act, made targeted by the roots. The roots offered at boot
+are a MAP: the agent sees that a dispatch tree, a narrative tree, a memory tree exist. When a
+prompt needs one, the agent Reads it. This is the instrument doing its job (confidence is not
+accuracy, go read, [[feedback_self_judgment_is_coherence_not_accuracy]]), now with a map that
+makes the reading targeted instead of blind. The roots tell it what is there; reading is its
+choice. The step that is genuinely the agent's judgment, which tree this question needs, stays
+the agent's; the structure gives it a map, not a guess. The principle:
+[[feedback_dont_automate_the_agents_judgment]].
```
J-Melon: **Prose-reviewer** nitpick: section ends on a bare wiki-link, not a loaded sentence; the principle reference is load-bearing, but the close should land on a clause, not a pointer.

designs/ai/memory-as-graph.md
J-Melon: **Marvin** issue: `UserPromptSubmit` named as an offer-delivery mechanism contradicts the PR premise; the descent hook was dropped. Replace with `SessionStart` only (the correction-signal hook is separate, not a memory-offer surface).

Issue #722, opened 2026-05-24
Memory triggers and consolidation


J-Melon:
MIGRATE realised memory rules into versioned `.claude/rules/`, convert action-triggered rules to Skills, and add dream-skill auto-consolidation<br>So that settled operating rules load every session and are shared with contributors, action-triggered rules fire on the right prompts, and the memory index stays under the load window.

Research (`designs/research/memory-system-improvements.md`; confirmed 2026-05-29): auto-memory is machine-local and cannot be relocated into the repo (`autoMemoryDirectory` is ignored from project settings, by design). `.claude/rules/*.md` is the supported surface for always-on, versioned project rules (loaded at launch; `paths:` frontmatter scopes conditionally). Skills remain the surface for action-triggered rules (fire on a description-trigger, e.g. ticket-shape on save_issue). Realised rules graduate terse and informational, no provenance; auto-memory stays the personal accumulation and history layer.

**Acceptance Criteria:**

- [ ] Realised always-on rules live in clustered `.claude/rules/*.md` (review, dispatch, linear, git, gdscript), terse, no provenance, no `feedback_*` path references.
- [ ] High-traffic action-triggered rules (ticket-shape, dandori, narrative discipline, voice) load as Skills with description-triggers.
- [ ] Private (user, comp) and project-fiction memories are not migrated; PRIVATE.md untouched.
- [ ] The redundant `inject-priority-memory.sh` SessionStart hook and the `priority:` flags are retired; `.claude/rules/` supersedes them.
- [ ] dream-skill auto-consolidation runs on the 24h schedule.
- [ ] MEMORY.md stays under 200 lines after the first pass.

PR #881, opened 2026-06-07
feat: memory graph edge-validation lint and schema

J-Melon:
Switches the memory repo pre-commit integration from a raw hook installer script to lefthook, matching the volley project standard. The lint script (`scripts/memory/lint-graph-edges.sh`) is unchanged; only the install mechanism changes.

`install-memory-hook.sh` is deleted. In its place, `scripts/memory/memory-lefthook.yml` is a lefthook config template: copy it to the memory repo root and run `lefthook install` once. Lefthook resolves the repo root itself, dissolving the hardcoded worktree path the old installer baked into the generated hook.

**Manual post-merge step (Josh):** remove `~/.claude/projects/-home-josh-gamedev-volley/memory/.git/hooks/pre-commit` if it exists from the old installer, copy `scripts/memory/memory-lefthook.yml` to that repo's root as `lefthook.yml`, then run `lefthook install` there.

#722


scripts/memory/lint-graph-edges.sh
```diff
@@ -0,0 +1,53 @@
...
+    target="$MEMORY_DIR/${parent_value}.md"
+
+    if [[ ! -f "$target" ]]; then
+        echo "dangling parent: $filename -> $parent_value (no file: ${parent_value}.md)"
+        dangling=$((dangling + 1))
+    fi
+done < <(find "$MEMORY_DIR" -maxdepth 1 -name "*.md" -print0 | sort -z)
+
+echo "lint-graph-edges: $dangling dangling, $orphans root/untyped nodes"
+
```
J-Melon, line 49: **Reviewer** issue: `find -maxdepth 1` silently skips `letters/` and any nested subdir. Verified: a dangling parent in `letters/bad-letter.md` exits 0 with no report. The design names `letters/` as a first-class subdir; drop `-maxdepth 1` and resolve targets relative to `MEMORY_DIR`, or use `-maxdepth 2` and anchor `target` against `MEMORY_DIR`.

scripts/memory/install-memory-hook.sh
```diff
@@ -0,0 +1,39 @@
...
+    exit 1
+fi
+
+if [[ ! -f "$LINT_SCRIPT" ]]; then
+    echo "install-memory-hook: lint script not found at $LINT_SCRIPT" >&2
+    exit 1
+fi
+
+cat > "$HOOK_PATH" <<EOF
+#!/usr/bin/env bash
```
J-Melon: **Reviewer** issue: `cat > "$HOOK_PATH"` silently overwrites any existing pre-commit hook. Verified: an existing hook is clobbered with no warning. Guard with `if [[ -e "$HOOK_PATH" ]]; then echo 'hook exists; remove it first' >&2; exit 1; fi` before the heredoc.

scripts/memory/lint-graph-edges.sh
J-Melon: Resolved in 9db3e31d: dropped `-maxdepth 1`, the lint now recurses into `letters/` and subdirs; new fixture covers a dangling parent in a subdir. Block resolved.

scripts/memory/install-memory-hook.sh
J-Melon: Resolved in 9db3e31d: install now errors "hook exists; remove it first" and exits 1 rather than overwriting. Block resolved.

PR #888, opened 2026-06-08
feat: memory-forest roots generator

J-Melon:
Implements `generate-roots.sh`: walks the memory corpus for files whose basename starts with `trunk_`, emits the five trunks as the CROWN (slug + gist), and exits 2 with a truncation notice if the output would exceed the 10K SessionStart cap.

Plain parentless nodes are excluded from the crown. Filename-based detection survives the harness normalizer, which rewrites `node_type` to `memory` on every edit and would clobber `node_type: trunk`.

This is AC2 of #722 (the roots generate as the boot reading list). The generator is intentionally standalone here: wiring its output into MEMORY.md and the SessionStart offer is AC3, a separate follow-up. The anti-drift promise is redeemed when AC3 lands, not in this PR.

#722

### DIFF
--- a/scripts/memory/generate-roots.sh
+++ b/scripts/memory/generate-roots.sh
@@ -1,42 +1,29 @@
 #!/usr/bin/env bash
-# Emits MEMORY.md's CROWN: the five trunk roots plus their ordered children
-# and provisional bridge nodes.
+# Emits MEMORY.md's CROWN: one line per trunk root, slug + gist.
 #
 # A trunk is any .md file whose frontmatter contains node_type: trunk.
 # Plain parentless nodes are NOT trunks and do NOT appear in the crown.
 #
-# Output structure for each trunk:
+# Output (one line per trunk, sorted by filename slug):
 #   <slug>  <gist>
-#     ordered children (nodes whose parent: == trunk slug):
-#       - <child-slug>  <child-gist>
-#     provisional (bridge nodes bucketed to this trunk but not yet typed):
-#       provisional:
-#       - <bridge-slug>
 #
 # Cap: total output must stay under --budget chars (default 10000).
 # Exits 2 with a truncation notice on stderr when the cap would be exceeded.
 #
 # Usage:
-#   generate-roots.sh [MEMORY_DIR] [--budget N] [--bridge PATH]
+#   generate-roots.sh [MEMORY_DIR] [--budget N]
 #
 # MEMORY_DIR defaults to ~/.claude/projects/-home-josh-gamedev-volley/memory
-# --bridge defaults to /home/josh/gamedev/volley/ai/scratchpads/memory-bucketing-proposed.md
-# parent: may sit at the top level of frontmatter or nested (any indent); both are read.
 
 set -euo pipefail
 
 MEMORY_DIR=""
 BUDGET=10000
-BRIDGE_MAP=""
 
 while [[ $# -gt 0 ]]; do
     case "$1" in
         --budget)
             BUDGET="$2"
-            shift 2
-            ;;
-        --bridge)
-            BRIDGE_MAP="$2"
             shift 2
             ;;
         *)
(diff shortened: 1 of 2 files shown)


scripts/memory/test-generate-roots.sh
```diff
@@ -0,0 +1,212 @@
...
+cat > "$DIR4/trunk_prose.md" <<'EOF'
+---
+node_type: trunk
+slug: trunk_prose
+---
+The prose gist for this trunk.
+EOF
+
+output=$("$GEN" "$DIR4" 2>/dev/null)
+if echo "$output" | grep -qF "The prose gist for this trunk."; then
```
J-Melon, line 176: **Trillian** issue: test 4 only fixtures a trunk without a `description:` field, so it exercises the prose fallback only; a broken `read_field description` falls silently through to prose and all 10 tests still pass. All five live trunk files carry `description:` frontmatter. Add a fixture with a `description:` field and assert that value appears in the crown output.

scripts/memory/test-generate-roots.sh
```diff
@@ -0,0 +1,212 @@
...
+    local gist="${4:-gist for $slug}"
+    cat > "$dir/${slug}.md" <<EOF
+---
+node_type: $node_type
+slug: $slug
+---
+$gist
+EOF
+}
+
```
J-Melon, line 87: **Trillian** suggestion: `derive_gist` tries `description`, then `summary`, then prose; breaking `summary:` also silently passes the suite. One fixture trunk with only a `summary:` field (no `description:`, no prose gist) would close the gap.

scripts/memory/generate-roots.sh
```diff
@@ -0,0 +1,153 @@
...
+
+set -euo pipefail
+
+MEMORY_DIR=""
+BUDGET=10000
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        --budget)
+            BUDGET="$2"
```
J-Melon, line 30: **Ford** suggestion: `BUDGET="$2"` accepts any string; a non-numeric value trips an opaque `unbound variable` crash at the arithmetic on line 126. Add `[[ "$BUDGET" =~ ^[0-9]+$ ]] || { echo "generate-roots: --budget requires a positive integer" >&2; exit 1; }` after line 26.

scripts/memory/generate-roots.sh
```diff
@@ -0,0 +1,153 @@
...
+declare -A IS_TRUNK
+
+while IFS= read -r -d '' filepath; do
+    slug="$(basename "$filepath" .md)"
+    NODE_FILE["$slug"]="$filepath"
+
+    if [[ "$slug" == trunk_* ]]; then
+        IS_TRUNK["$slug"]=1
+    fi
+done < <(find "$MEMORY_DIR" -name "*.md" -print0 | sort -z)
```
J-Melon: **Ford** suggestion: `find` recurses into subdirs (`letters/`). A future `trunk_*` letter would appear in the crown. Add `-maxdepth 1` to lock scope to the flat memory dir.

scripts/memory/test-generate-roots.sh
J-Melon: **Gru**

resolved: b654d7f2fbec42657ae5be57703f739a01aef8d0 adds tests 4b/4c fixturing description: and summary: trunks; both go red if the reader is stubbed. Verified.

scripts/memory/test-generate-roots.sh
J-Melon: **Gru**

resolved: b654d7f2fbec42657ae5be57703f739a01aef8d0 test 4c covers the summary: fallback path.

scripts/memory/generate-roots.sh
J-Melon: **Gru**

resolved: b654d7f2fbec42657ae5be57703f739a01aef8d0 --budget now validates ^[0-9]+$ and exits 1 with a clear message; test 4e covers it.

scripts/memory/generate-roots.sh
J-Melon: **Gru**

resolved: b654d7f2fbec42657ae5be57703f739a01aef8d0 find restricted to -maxdepth 1; test 4d asserts a letters/trunk_ nested file is excluded.

PR #889, opened 2026-06-08
fix: devils-advocate posts inline, drop stale no-shell claim

J-Melon:
The dispatch skill told the dispatcher that devils-advocate has no shell access and to expect a context-blocked report, but the agent definition grants it Bash. The two surfaces disagreed, and following the skill meant briefing the battler to reason from the diff text alone, so its findings never reached the PR as inline comments. During a battle on #888 the devils-advocate wrote out anchored findings in its report and claimed it had posted them, when it had posted nothing.

This corrects the skill to match the agent definition: devils-advocate has Bash and posts its findings as inline review comments like any other reviewer.


.claude/skills/dispatch/SKILL.md
```diff
@@ -158,7 +158,7 @@ If the fan would require any file to appear in two slices, the work is not fan-s
...
 
-A reviewer battle runs only when Josh asks for one. When he does: scope-filter the diff by changed path and fan only the matching specialists (code-quality, gdscript-conventions, test-coverage on a GDScript diff; domain reviewers on the files they own). The full path → specialist map and the verdict contract live in `.claude/skills/reviewers/SKILL.md`. Battlers (devils-advocate, integration-scenario-author) fire alongside only when the battle is requested. Devils-advocate has no shell access; pass the rule text inline or expect a context-blocked report.
+A reviewer battle runs only when Josh asks for one. When he does: scope-filter the diff by changed path and fan only the matching specialists (code-quality, gdscript-conventions, test-coverage on a GDScript diff; domain reviewers on the files they own). The full path → specialist map and the verdict contract live in `.claude/skills/reviewers/SKILL.md`. Battlers (devils-advocate, integration-scenario-author) fire alongside only when the battle is requested. Devils-advocate has Bash, so it posts its findings as inline review comments like any reviewer; brief it to do so, and do not tell it to reason from the diff text alone.
```
J-Melon, line 161: **Slartibartfast** issue: "brief it to do so" relies entirely on the dispatch prompt to override the agent definition, which under "Out of scope" still says "PR review. You do not read diffs or apply labels." A dispatcher reading only the skill gets the right behaviour; a dispatcher reading only the agent def gets a contradiction. The fix belongs in both surfaces: update the agent def to remove or scope-narrow the "Out of scope" clause so the two surfaces agree without the dispatcher needing to carry the whole burden.

J-Melon: **Gru**

resolved: 5fdaa4b9 updates the agent def Out-of-scope and Output to match the skill (posts inline, no "no PR surface"); e5d1bab6 fixes the frontmatter "not a PR" guidance. Re-review approved.

PR #891, opened 2026-06-08
feat: cap memory files at 2000 chars

J-Melon:
A memory leaf is one rule. This adds a PreToolUse hook on Edit and Write that denies a memory-dir leaf .md file growing past 2000 chars, a ceiling for new material: over half the existing corpus is already above it, so the cap does not retroactively enforce, it stops new rules and currently-compliant ones from sprawling. The deny message names the fix: split the rule, or push doctrine into child nodes and leave a short index.

The cap only blocks growth. A shrinking edit on an already-oversized node passes (with a message that says to keep cutting), so the existing over-cap files can be fixed rather than frozen. MEMORY.md (the generated index) and the handoff letters under letters/ are exempt, since both are legitimately large. The hook fails open on any parse error.


.claude/hooks/memory-file-char-cap.sh
```diff
@@ -0,0 +1,56 @@
...
+# A memory node is one rule (a leaf, ~1.6-2.7k chars in the corpus) or a short index
+# node; nothing should legitimately exceed the cap. Over it means two rules to separate,
+# or a root hoarding doctrine that should push down into its children.
+# Only fires on files under the memory dir. Index files (MEMORY.md) are exempt.
+# Fails open on any parse error.
+set -uo pipefail
+
+CAP=2000
+MEMDIR="/home/josh/.claude/projects/-home-josh-gamedev-volley/memory"
+input="$(cat)"
```
J-Melon, line 9: **Riebeck** issue: `MEMDIR` hardcodes `/home/josh/...`; breaks on any other machine. Use `${HOME}/.claude/...` as `inject-priority-memory.sh` line 7 does.

.claude/hooks/memory-file-char-cap.sh
```diff
@@ -0,0 +1,56 @@
...
+MEMDIR = '$MEMDIR'
+try:
+    d = json.load(sys.stdin)
+    tool = d.get('tool_name', '')
+    ti = d.get('tool_input', {})
+    path = ti.get('file_path', '')
+    # only memory-dir .md files, excluding the generated index
+    rp = os.path.realpath(path) if path else ''
+    if not rp.startswith(os.path.realpath(MEMDIR)) or not rp.endswith('.md'):
+        print('PASS'); sys.exit()
```
J-Melon, line 23: **Riebeck** issue: `rp.startswith(REAL_MEMDIR)` matches `memoryx/evil.md` (prefix without separator). Change to `rp.startswith(REAL_MEMDIR + "/")`.

.claude/hooks/memory-file-char-cap.sh
```diff
@@ -0,0 +1,56 @@
...
+result="$(printf '%s' "$input" | python3 -c "
+import json, sys, os
+CAP = $CAP
+MEMDIR = '$MEMDIR'
+try:
+    d = json.load(sys.stdin)
+    tool = d.get('tool_name', '')
+    ti = d.get('tool_input', {})
+    path = ti.get('file_path', '')
+    # only memory-dir .md files, excluding the generated index
```
J-Melon: **Gabbro** issue: `startswith(realpath(MEMDIR))` without a trailing slash would match a sibling dir named `memory2/`; add `+ "/"` to the argument.

.claude/hooks/memory-file-char-cap.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
+# PreToolUse on Edit/Write: deny when the call would leave a memory FILE over the cap.
+# A memory node is one rule (a leaf, ~1.6-2.7k chars in the corpus) or a short index
+# node; nothing should legitimately exceed the cap. Over it means two rules to separate,
+# or a root hoarding doctrine that should push down into its children.
```
J-Melon: **Gabbro** nitpick: multi-line comment block (lines 1-7); comment policy allows one line max; trim to the single most load-bearing line or move the prose to a doc.

.claude/hooks/memory-file-char-cap.sh
```diff
@@ -0,0 +1,56 @@
...
+CAP = $CAP
+MEMDIR = '$MEMDIR'
+try:
+    d = json.load(sys.stdin)
+    tool = d.get('tool_name', '')
+    ti = d.get('tool_input', {})
+    path = ti.get('file_path', '')
+    # only memory-dir .md files, excluding the generated index
+    rp = os.path.realpath(path) if path else ''
+    if not rp.startswith(os.path.realpath(MEMDIR)) or not rp.endswith('.md'):
```
J-Melon: **Solanum** issue: hook fires on `memory/letters/YYYY-MM-DD-*.md` too; every handoff letter is 3800-5600 chars and written via Write, so every new letter is blocked. Exclude `memory/letters/` from the path check.

.claude/hooks/memory-file-char-cap.sh
```diff
@@ -0,0 +1,56 @@
...
+        new = ti.get('new_string', '')
+        if ti.get('replace_all'):
+            size = len(cur.replace(old, new))
+        else:
+            size = len(cur.replace(old, new, 1))
+    else:
+        print('PASS'); sys.exit()
+    # Block only when the result is over cap AND not smaller than now.
+    # A shrinking edit on an already-oversized node is the fix, never blocked.
+    print('DENY %d' % size if (size > CAP and size >= len(cur)) else 'PASS')
```
J-Melon: **Solanum** issue: grow-only passes any shrink, but a 16000-char file can grow to 15999 (shrink, passes), then to 16001 (grows: 16001>2000 AND 16001>=15999, DENY). Correction: the hole runs the other direction. A file at 2500 shrinks to 2100 (passes), then grows to 2200 (DENY: 2200>=2100). So shrinking never escapes; the file is trapped over cap until it crosses below. That is fine, but note: the error message says "merge nothing" when the correct next step on a shrink pass is "keep shrinking." The DENY fires immediately on the next growth attempt even if the file shrank meaningfully. This is workable but the error message misdirects.

.claude/hooks/memory-file-char-cap.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
+# PreToolUse on Edit/Write: deny when the call would leave a memory FILE over the cap.
+# A memory node is one rule (a leaf, ~1.6-2.7k chars in the corpus) or a short index
+# node; nothing should legitimately exceed the cap. Over it means two rules to separate,
+# or a root hoarding doctrine that should push down into its children.
+# Only fires on files under the memory dir. Index files (MEMORY.md) are exempt.
```
J-Melon: **Solanum** question: the comment calls this a cap for "a leaf or a short index node" but makes no node_type distinction at enforcement time. When SH-479 lands and generated topic files are written via Write, the hook blocks them. Exempt trunk_ files by slug prefix, or skip files with node_type: topic in frontmatter.

.claude/settings.json
```diff
@@ -4,59 +4,109 @@
...
+            "command": "\"$CLAUDE_PROJECT_DIR/.claude/hooks/inject-priority-memory.sh\""
+          },
+          {
+            "type": "command",
+            "command": "\"$CLAUDE_PROJECT_DIR/.claude/hooks/inject-latest-handoff.sh\""
+          }
         ]
       }
     ]
   }
```
J-Melon, line 111: **Solanum** nitpick: PR removes the trailing newline from settings.json; main had one. Adds churn to future diffs.

.claude/hooks/memory-file-char-cap.sh
J-Melon: **Gru**

resolved: fd34f001 uses ${HOME} like the other memory hooks.

.claude/hooks/memory-file-char-cap.sh
J-Melon: **Gru**

resolved: fd34f001 matches MEMDIR + os.sep, so a sibling memoryx/ no longer matches.

.claude/hooks/memory-file-char-cap.sh
J-Melon: **Gru**

resolved: fd34f001 prefix match now requires the separator. Same fix as Riebeck's.

.claude/hooks/memory-file-char-cap.sh
J-Melon: **Gru**

resolved: fd34f001 comment trimmed to three lines.

.claude/hooks/memory-file-char-cap.sh
J-Melon: **Gru**

resolved: fd34f001 exempts letters/ (and MEMORY.md). Verified a 4000-char letter write passes.

.claude/hooks/memory-file-char-cap.sh
J-Melon: **Gru**

resolved: fd34f001 the deny message is now shrink-aware: an already-oversized file says keep cutting, not split.

.claude/hooks/memory-file-char-cap.sh
J-Melon: **Gru**

noted: topics are stdout-only today so no Write hits the hook. When SH-479 writes generated topics, the exemption lands there. Flagged, not blocking.

.claude/settings.json
J-Melon: **Gru**

resolved: fd34f001 restores the trailing newline on settings.json.

.claude/hooks/memory-file-char-cap.sh
J-Melon: **Solanum**

resolved: fd34f001 exempts letters/ at the path check. Tested 4000-char Write to memory/letters/ -- PASS.

.claude/hooks/memory-file-char-cap.sh
J-Melon: **Solanum**

resolved: fd34f001 deny message is now shrink-aware. Oversized non-shrinking write says keep cutting; fresh-overrun says split. Tested both paths.

.claude/hooks/memory-file-char-cap.sh
J-Melon: **Riebeck**

resolved: fd34f001 uses ${HOME} in MEMDIR. Verified hook fires on a 2001-char write to the memory dir.

.claude/hooks/memory-file-char-cap.sh
J-Melon: **Riebeck**

resolved: fd34f001 uses MEMDIR + os.sep in the startswith check. Verified sibling memoryx/ write returns PASS.

.claude/hooks/memory-file-char-cap.sh
J-Melon: **Gabbro**

resolved fd34f001: prefix check now requires `MEMDIR + os.sep`; sibling memory2/ mock payload returns PASS.

PR #894, opened 2026-06-08
feat: SessionStart hook injects memory-forest crown

J-Melon:
At session boot, the crown now arrives as additionalContext: five trunk slugs and gists that tell the agent which forest tree to enter before touching its leaves. Wires inject-memory-crown.sh (new) into settings.json under SessionStart, calling the already-merged generate-roots.sh. Fails open: any generator error emits nothing.

#892

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -4,102 +4,58 @@
       {
         "matcher": ".*",
         "hooks": [
-          {
-            "type": "command",
-            "command": "\"$CLAUDE_PROJECT_DIR/.claude/hooks/em-dash-pre-tool.sh\""
-          }
+          { "type": "command", "command": "\"$CLAUDE_PROJECT_DIR/.claude/hooks/em-dash-pre-tool.sh\"" }
         ]
       },
       {
         "matcher": "mcp__linear__save_issue",
         "hooks": [
-          {
-            "type": "command",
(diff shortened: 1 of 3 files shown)


.claude/settings.json
```diff
@@ -4,57 +4,102 @@
       {
         "matcher": ".*",
         "hooks": [
-          { "type": "command", "command": "\"$CLAUDE_PROJECT_DIR/.claude/hooks/em-dash-pre-tool.sh\"" }
+          {
+            "type": "command",
+            "command": "\"$CLAUDE_PROJECT_DIR/.claude/hooks/em-dash-pre-tool.sh\""
+          }
```
J-Melon: **Chert** issue: reformat of all existing one-line hook entries is PR noise; #891 and #895 base off the same `ae5b052d` and will conflict on merge. Add only the SessionStart entry; leave existing lines untouched.

J-Melon: **Gru**

resolved: 1207954c settings.json now adds only the one-line crown entry, no reformat of existing hooks. Diff is +1 line.

J-Melon: **Chert** resolved by 1207954c: intra-PR reformat squashed; net diff against main is one comma + one new entry only. Block cleared.

PR #895, opened 2026-06-08
feat: cap reviewer inline comments at 300 chars

J-Melon:
The reviewers skill caps an inline finding at 30 words and two lines, but nothing enforced it: only the synthesis verdict body and the Linear comment/issue bodies had char-cap hooks. A long inline restates what the diff already shows and buries the one clause that matters. This adds a PreToolUse hook on Bash, gated to a reviews or comments POST, that denies when any inline finding body exceeds 300 chars, the same ceiling the synthesis body and Linear comment hooks use. The deny message points the reviewer to put reasoning in the dispatcher report, not the thread. Non-review commands and parse errors pass.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -4,57 +4,102 @@
       {
         "matcher": ".*",
         "hooks": [
-          { "type": "command", "command": "\"$CLAUDE_PROJECT_DIR/.claude/hooks/em-dash-pre-tool.sh\"" }
+          {
+            "type": "command",
+            "command": "\"$CLAUDE_PROJECT_DIR/.claude/hooks/em-dash-pre-tool.sh\""
+          }
         ]
       },
       {
         "matcher": "mcp__linear__save_issue",
         "hooks": [
-          { "type": "command", "command": "\"$CLAUDE_PROJECT_DIR/.claude/hooks/linear-issue-body-cap.sh\"" }
+          {
+            "type": "command",
+            "command": "\"$CLAUDE_PROJECT_DIR/.claude/hooks/linear-issue-body-cap.sh\""
+          }
         ]
       },
       {
         "matcher": "mcp__linear__save_comment",
         "hooks": [
-          { "type": "command", "command": "\"$CLAUDE_PROJECT_DIR/.claude/hooks/linear-comment-body-cap.sh\"" }
+          {
+            "type": "command",
+            "command": "\"$CLAUDE_PROJECT_DIR/.claude/hooks/linear-comment-body-cap.sh\""
+          }
         ]
       },
       {
         "matcher": "Agent",
         "hooks": [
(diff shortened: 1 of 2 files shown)


.claude/settings.json
```diff
@@ -4,59 +4,104 @@
...
+            "command": "\"$CLAUDE_PROJECT_DIR/.claude/hooks/rm-permission-reason.sh\""
+          },
+          {
+            "type": "command",
+            "if": "Bash(*bot-review.yml*)",
+            "command": "\"$CLAUDE_PROJECT_DIR/.claude/hooks/synthesis-body-cap.sh\""
+          },
+          {
+            "type": "command",
+            "command": "\"$CLAUDE_PROJECT_DIR/.claude/hooks/reviewer-inline-comment-cap.sh\""
```
J-Melon: **Esker** issue: full-file reformat drops #891 Edit|Write/memory-file-char-cap and #894 inject-memory-crown entries; surgical append to the Bash hooks array only.

.claude/settings.json
```diff
@@ -4,59 +4,104 @@
...
+            "command": "\"$CLAUDE_PROJECT_DIR/.claude/hooks/inject-priority-memory.sh\""
+          },
+          {
+            "type": "command",
+            "command": "\"$CLAUDE_PROJECT_DIR/.claude/hooks/inject-latest-handoff.sh\""
+          }
         ]
       }
     ]
   }
```
J-Melon, line 115: **Esker** nitpick: trailing newline removed; restore it to keep the file POSIX-compliant.

.claude/settings.json
J-Melon: **Gru**

resolved: ac25dbce settings.json now appends only the one-line inline-cap entry, no reformat. Will not drop #891/#894 entries.

.claude/settings.json
J-Melon: **Gru**

resolved: ac25dbce trailing newline restored with the main-based settings.json.

.claude/settings.json
J-Melon: **Esker** ac25dbce still reformats all hook objects to single-line; #894 is open and touches settings.json, so a merge here forces a conflict there. Block stands until the diff is a surgical append only.

.claude/settings.json
J-Melon: **Esker** ac25dbce: trailing newline confirmed present. Nitpick resolved.

.claude/settings.json
J-Melon: **Esker**

resolved: 4b7bf7e7. Semantic diff vs main is the single inline-cap entry only; all other entries present including #891 Edit|Write block. #894 merged, conflict risk gone. Reformat is cosmetic.

PR #897, opened 2026-06-08
ci: jq pre-commit formatter for JSON files

J-Melon:
Adds a `jq-format` step to the `pre-commit` lefthook block, mirroring `gdformat` in shape: `glob: "*.json"`, excludes `addons/**`, runs `jq --indent 2` over each staged file in place, `stage_fixed: true` re-stages the result. This ends the recurring format drift on `.claude/settings.json` and friends that causes spurious conflict noise across PRs.

The same commit canonicalizes the 3 tracked JSON files that were already off-format (`.claude/settings.json`, `.gutconfig.json`, `.mcp.json`), so the hook is a no-op from this point forward unless someone hand-edits a file.

#896

Agent-Role: general-purpose


lefthook.yml
```diff
@@ -24,6 +24,12 @@ pre-commit:
       exclude:
         - "addons/**"
       run: codespell {staged_files}
+    jq-format:
+      glob: "*.json"
+      exclude:
+        - "addons/**"
+      run: for f in {staged_files}; do jq --indent 2 . "$f" > "$f.tmp" && mv "$f.tmp" "$f"; done
```
J-Melon, line 31: **Yarrow** nitpick: `{staged_files}` is raw tokens; a space in a JSON path splits into two broken args. No current file is affected; mirrors gdformat precedent.

PR #898, opened 2026-06-08
feat: topic-index generator for memory forest

J-Melon:
Adds `scripts/memory/generate-index.sh`: given a topic node slug, emits its description plus a `Children:` list of each direct child with a one-line gist, derived from `parent:` edges in the corpus. Leaves (nodes with no children) are silently skipped with exit 0. Reuses the same gist-derivation logic from `generate-roots.sh`.

Tests in `scripts/memory/test-generate-index.sh` cover topic output, leaf skip, missing-slug error, depth isolation (grandchildren excluded), and prose-gist fallback.

The char-cap hook (`AC3`) needs one change: replace the hardcoded `MEMDIR` assignment and add `is_topic()` detection so topics are exempt. That change is blocked by a write gate on `.claude/hooks/**`. Exact diff to apply:

```diff
-MEMDIR="${HOME}/.claude/projects/-home-josh-gamedev-volley/memory"
+MEMDIR="${MEMDIR_OVERRIDE:-${HOME}/.claude/projects/-home-josh-gamedev-volley/memory}"
```

And in the Python block, add `import re` to the imports, add the `is_topic()` function (scans MEMDIR for any `.md` whose frontmatter has `parent: <slug>`), and call it after the `MEMORY.md`/`letters/` exemption:

```python
+import re
+
+def is_topic(slug, memdir):
+    try:
+        for fname in os.listdir(memdir):
+            if not fname.endswith('.md'):
+                continue
+            fpath = os.path.join(memdir, fname)
+            try:
+                text = open(fpath, encoding='utf-8').read(4096)
+                parts = text.split('---', 2)
+                if len(parts) < 3:
+                    continue
+                fm = parts[1]
+                for line in fm.splitlines():
+                    m = re.match(r'^\s*parent:\s*(.+)', line)
+                    if m and m.group(1).strip() == slug:
+                        return True
+            except OSError:
+                continue
+    except OSError:
+        pass
+    return False
+
 # exempt the generated index and the handoff letters
 if os.path.basename(rp) == 'MEMORY.md' or (os.sep + 'letters' + os.sep) in rp:
     print('PASS'); sys.exit()
+# exempt topic nodes: any node that has children pointing at it
+slug = os.path.basename(rp)[:-3]
+if is_topic(slug, MEMDIR):
+    print('PASS'); sys.exit()
```

Once the hook is updated, the skipped char-cap tests in `test-generate-index.sh` will activate and pass.

#893

### DIFF
--- a/.claude/hooks/memory-file-char-cap.sh
+++ b/.claude/hooks/memory-file-char-cap.sh
@@ -5,13 +5,35 @@
 set -uo pipefail
 
 CAP=2000
-MEMDIR="${HOME}/.claude/projects/-home-josh-gamedev-volley/memory"
+MEMDIR="${MEMDIR_OVERRIDE:-${HOME}/.claude/projects/-home-josh-gamedev-volley/memory}"
 input="$(cat)"
 
 result="$(printf '%s' "$input" | python3 -c "
-import json, sys, os
+import json, sys, os, re
 CAP = $CAP
 MEMDIR = os.path.realpath('$MEMDIR')
+
+def is_topic(slug, memdir):
+    # a topic is any node with at least one child pointing at it via parent:
+    try:
+        for fname in os.listdir(memdir):
+            if not fname.endswith('.md'):
+                continue
+            try:
+                text = open(os.path.join(memdir, fname), encoding='utf-8').read(4096)
+                parts = text.split('---', 2)
+                if len(parts) < 3:
+                    continue
+                for line in parts[1].splitlines():
+                    m = re.match(r'^\s*parent:\s*(.+)', line)
+                    if m and m.group(1).strip() == slug:
+                        return True
+            except OSError:
+                continue
+    except OSError:
+        pass
+    return False
+
 try:
     d = json.load(sys.stdin)
     tool = d.get('tool_name', '')
@@ -23,6 +45,9 @@ try:
         print('PASS'); sys.exit()
     # exempt the generated index and the handoff letters (both legitimately large)
     if os.path.basename(rp) == 'MEMORY.md' or (os.sep + 'letters' + os.sep) in rp:
+        print('PASS'); sys.exit()
+    # exempt topic nodes: a topic's body is generated, only leaves are capped
+    if is_topic(os.path.basename(rp)[:-3], MEMDIR):
         print('PASS'); sys.exit()
     try:
         cur = open(rp, encoding='utf-8').read()
--- a/.claude/hooks/memory-file-char-cap.sh
+++ b/.claude/hooks/memory-file-char-cap.sh
@@ -1,82 +1,56 @@
 #!/usr/bin/env bash
 # PreToolUse on Edit/Write: deny a memory-dir leaf .md file growing past the cap.
-# Leaves are capped; MEMORY.md (generated index) and letters/ (handoff letters) are exempt.
-# Grow-only: a shrinking edit on an already-oversized file always passes. Fails open.
+# Leaves are capped; MEMORY.md (generated index), letters/, and topic nodes (any node
+# with children) are exempt. Grow-only: a shrinking edit always passes. Fails open.
 set -uo pipefail
 
 CAP=2000
 MEMDIR="${MEMDIR_OVERRIDE:-${HOME}/.claude/projects/-home-josh-gamedev-volley/memory}"
+MEMDIR="$(realpath "$MEMDIR" 2>/dev/null || echo "$MEMDIR")"
 input="$(cat)"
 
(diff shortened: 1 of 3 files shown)


.claude/hooks/memory-file-char-cap.sh
```diff
@@ -5,13 +5,35 @@
...
+
+def is_topic(slug, memdir):
+    # a topic is any node with at least one child pointing at it via parent:
+    try:
+        for fname in os.listdir(memdir):
+            if not fname.endswith('.md'):
+                continue
+            try:
+                text = open(os.path.join(memdir, fname), encoding='utf-8').read(4096)
+                parts = text.split('---', 2)
```
J-Melon: **Arkose** nitpick: `text.split("---", 2)` splits on the bare substring, not a line boundary; a frontmatter value containing `---` (e.g. `A---B`) splits early and drops the `parent:` line. Use `re.split(r"(?m)^---\s*$", text, 2)` instead.

scripts/memory/generate-index.sh
```diff
@@ -0,0 +1,156 @@
...
+    # the first prose line, with markdown stripped so the gist reads clean
+    awk '
+        /^---[[:space:]]*$/ { fence++; next }
+        fence < 2 { next }
+        /^[[:space:]]*$/ { next }
+        /^#/ { next }
+        {
+            gsub(/\*\*/, "")       # bold markers
+            gsub(/`/, "")          # inline code
+            sub(/^[[:space:]]*[-*>][[:space:]]+/, "")  # list/quote marker
```
J-Melon: **Gneiss** nitpick: ordered-list prefix `1. ` is not stripped; only `- * >` are matched. Add `[0-9]+\.` to the sub pattern.

.claude/hooks/memory-file-char-cap.sh
```diff
@@ -1,57 +1,56 @@
...
+slug="${base%.md}"
+[ "$base" = "MEMORY.md" ] && exit 0
+case "$rp" in *"/letters/"*) exit 0 ;; esac
+# a topic is any node a child points at via parent:; only leaves are capped
+if grep -lqE "^[[:space:]]*parent:[[:space:]]*${slug}[[:space:]]*$" "$MEMDIR"/*.md 2>/dev/null; then
+  exit 0
+fi
+
+# resulting size of the file after this Write/Edit, computed by jq against the current file
+cur="$([ -f "$rp" ] && wc -c < "$rp" || echo 0)"
```
J-Melon: **Gneiss** issue: `wc -c` counts bytes; `jq | length` counts chars. A file with multibyte content makes `cur` (bytes) > `size` (chars), so the grow-only guard `size >= cur` is false when the edit grows past cap. Fix: use `wc -m` for `cur`.

.claude/hooks/memory-file-char-cap.sh
```diff
@@ -1,57 +1,56 @@
...
+if [ "$tool" = "Write" ]; then
+  size="$(printf '%s' "$input" | jq -r '(.tool_input.content // "") | length' 2>/dev/null)"
+elif [ "$tool" = "Edit" ]; then
+  curtext="$([ -f "$rp" ] && cat "$rp" || echo "")"
+  size="$(jq -rn --arg cur "$curtext" --argjson ti "$(printf '%s' "$input" | jq '.tool_input')" '
+    ($ti.old_string // "") as $old | ($ti.new_string // "") as $new |
+    (if ($ti.replace_all // false) then ($cur | gsub($old; $new)) else ($cur | sub($old; $new)) end) | length
```
J-Melon: **Gneiss** issue: `jq sub($old;$new)` treats `old_string` as a regex; Python `str.replace` was literal. A `.` or `+` in `old_string` matches more than intended, giving a wrong post-edit size. Fix: escape `$old` before passing to sub.

.claude/hooks/memory-file-char-cap.sh
J-Melon: **Gneiss**

resolved: 42b37d4d switches `wc -c` to `wc -m`; `cur` and `size` now both count Unicode codepoints. Multibyte test confirmed (1999 x "é" = 3998 bytes passes, 2001 denies).

.claude/hooks/memory-file-char-cap.sh
J-Melon: **Gneiss**

resolved: 42b37d4d replaces `sub`/`gsub` with `split($old)|join($new)` for replace_all and `index($old)` + slice for single; both are literal. `[USD]` and `A.B` old_strings verified correct.

PR #910, opened 2026-06-09
fix: rack-return gear no longer requires equip-pose timeout

J-Melon:
Equipped gear could not be dragged back to the rack outside a timeout window. The `RackDropTarget.can_accept` gate was incorrectly calling `RallyGate.removal_allowed`, which requires `AT_EQUIP_POSE` (the same constraint as equipping onto the character). Rack-return is unconditional; only equipping requires the pose.

Removes the guard from `RackDropTarget`, inverts the test that codified the broken behaviour, and corrects the doc sentence in `gear.md`.

#824

### DIFF
--- a/designs/01-prototype/tech/gear.md
+++ b/designs/01-prototype/tech/gear.md
@@ -60,6 +60,6 @@ Over-capacity state across designer changes (an equipped item retired, `kit_slot
 
 The equip window opens on `TimeoutController.main_character_reached_equip_pose` and closes on `timeout_ended`. `CharacterDropTarget.can_accept` reads `TimeoutController.get_state() == AT_EQUIP_POSE` directly. Off-court releases hit other targets.
 
-Unequip is symmetric: dragging an equipped item back to the rack within the same window calls `unequip` and frees its slot. Outside the window, the character's drop target stays inert and the dragged item passes through to the next target in the priority list.
+Rack-return is unconditional: dragging an equipped item to the rack calls `unequip` and frees its slot regardless of timeout state. Only equipping onto the character requires `AT_EQUIP_POSE`.
 
 Mid-rally, the press on an equipped item is refused at the gesture entry. While a rally is in progress (`TimeoutController.IDLE` and at least one ball in `PLAY_NORMAL` / `PLAY_ARC`) the gesture never starts; nothing is held, nothing deactivates. The shared check lives on `RallyGate.is_rally_in_progress`, called with injected refs at every callsite. The dev panel's per-item remove-level button uses the same gate so debug paths cannot desync the effect system during play.
--- a/scripts/items/drop_targets/rack_drop_target.gd
+++ b/scripts/items/drop_targets/rack_drop_target.gd
@@ -6,19 +6,16 @@ extends DropTarget
 var _item_manager: Node
 var _drop_area: Area2D
 var _role: StringName
-var _timeout_controller: TimeoutController
 
 
 func configure(
 	item_manager: Node,
 	drop_area: Area2D,
 	role: StringName,
-	timeout_controller: TimeoutController = null,
 ) -> void:
 	_item_manager = item_manager
 	_drop_area = drop_area
 	_role = role
-	_timeout_controller = timeout_controller
 
 
 func can_accept(item_key: String, position: Vector2, _scale_factor: float = 1.0) -> bool:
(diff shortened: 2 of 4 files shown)


tests/unit/items/test_drop_targets.gd
```diff
@@ -163,12 +159,9 @@ func test_rack_drop_target_accepts_equipment_unequip_at_equip_pose() -> void:
 	manager.items.assign([equipment] as Array[ItemDefinition])
 
 	var area: Area2D = _make_drop_area(Vector2(-500, 0), Vector2(200, 100))
-	var timeout: TimeoutController = TimeoutControllerScript.new()
-	add_child_autofree(timeout)
-	timeout._state = TimeoutController.State.AT_EQUIP_POSE
 
 	var target: RackDropTarget = RackDropTargetScript.new()
-	target.configure(manager, area, &"equipment", timeout)
+	target.configure(manager, area, &"equipment")
```
J-Melon: **code-quality** issue: identical body to `_outside_equip_pose` above; pose state no longer gates the rack so the two tests are pure duplication. Delete this function.

tests/unit/items/test_drop_targets.gd
```diff
@@ -163,12 +159,9 @@ func test_rack_drop_target_accepts_equipment_unequip_at_equip_pose() -> void:
...
```
J-Melon: **test-coverage** nitpick: `..._at_equip_pose` now configures and asserts identically to `..._outside_equip_pose`; rack no longer branches on timeout state so these are structural duplicates. Confirmed known; defer removal to the planned refactor.

scripts/items/drop_targets/rack_drop_target.gd
```diff
@@ -6,30 +6,23 @@ extends DropTarget
 var _item_manager: Node
 var _drop_area: Area2D
 var _role: StringName
-var _timeout_controller: TimeoutController
```
J-Melon, line 9: **devils-advocate** nitpick: stale `_timeout_controller` field survives on main; the PR head removes it correctly. Confirming the merge resolves cleanly is the only action needed.

designs/01-prototype/tech/gear.md
```diff
@@ -60,6 +60,6 @@ Over-capacity state across designer changes (an equipped item retired, `kit_slot
 
 The equip window opens on `TimeoutController.main_character_reached_equip_pose` and closes on `timeout_ended`. `CharacterDropTarget.can_accept` reads `TimeoutController.get_state() == AT_EQUIP_POSE` directly. Off-court releases hit other targets.
 
-Unequip is symmetric: dragging an equipped item back to the rack within the same window calls `unequip` and frees its slot. Outside the window, the character's drop target stays inert and the dragged item passes through to the next target in the priority list.
+Rack-return is unconditional: dragging an equipped item to the rack calls `unequip` and frees its slot regardless of timeout state. Only equipping onto the character requires `AT_EQUIP_POSE`.
```
J-Melon, line 63: **devils-advocate** question: doc says the gate lives on `RallyGate.is_rally_in_progress`; the impl checks `AT_EQUIP_POSE` only. Pre-existing; not introduced here, but not fixed here either.

tests/unit/items/test_drop_targets.gd
J-Melon: **Gru**: deleted the duplicate in f4426ea; kept `_outside_equip_pose` (renamed its message to "regardless of timeout state"). 668/668 green.

tests/unit/items/test_drop_targets.gd
J-Melon: **Gru**: deleted the duplicate in f4426ea; kept `_outside_equip_pose` (renamed its message to "regardless of timeout state"). 668/668 green.

scripts/items/drop_targets/rack_drop_target.gd
J-Melon: **Gru**: the field, param, and call-site arg were removed in 6e93a63; merge resolves the stale-on-main reference cleanly.

designs/01-prototype/tech/gear.md
J-Melon: **Gru**: real but pre-existing (RallyGate.is_rally_in_progress vs AT_EQUIP_POSE), not introduced or fixed here. Out of #824 scope; flagged for Josh.

PR #912, opened 2026-06-09
ci: cap commit message at 200 chars

J-Melon:
The commit-message hook capped total length at 400; this lowers it to 200 so commit bodies stay terse and the longer rationale lives in the PR description instead. The memory doc that mirrors the rule is updated to match.

Agent-Role: dispatcher

### DIFF
--- a/scripts/ci/check_commit_msg.sh
+++ b/scripts/ci/check_commit_msg.sh
@@ -9,7 +9,7 @@
 #   - No scope parens: `fix(ci):` is rejected.
 #   - No `SH-N` prefix in subject.
 #   - No `[Codename]` suffix or `[Codename]` anywhere in subject.
-#   - Total message length (all lines combined) must be under 400 chars.
+#   - Total message length (all lines combined) must be under 200 chars.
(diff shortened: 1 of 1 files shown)


scripts/ci/check_commit_msg.sh
```diff
@@ -34,10 +34,10 @@ esac
 
 errors=()
 
-# Total char count (under 400).
+# Total char count (under 200).
 total_chars=$(printf '%s' "$msg_body" | wc -c)
```
J-Melon, line 38: **Marvin** question: `wc -c` counts bytes not characters; a single emoji in the subject costs 4 bytes toward the cap. Pre-existing, but more likely to surprise at 200. Intended, or swap to `wc -m`?

J-Melon: **Gru**: fair nit, but pre-existing (the 400 cap used wc -c too) and unreachable in practice, an emoji subject already fails the conventional-commit format rule. Leaving wc -c as-is; not worth widening this PR.

PR #913, opened 2026-06-09
fix: guard autoplay toggle from re-enabling paddle physics during timeout

J-Melon:
\`AutoplayController.toggle()\` unconditionally called \`paddle.set_physics_process(true)\` on toggle-off, stealing physics ownership back from \`TimeoutController\` mid-timeout and snapping the character to the play lane.

The fix gates that restore behind \`timeout_controller.is_active()\`; when a timeout is in flight the toggle skips the re-enable entirely, and the timeout's own \`_finish_at_lane\` restores physics_process at the right moment.

Note: \`PlayerPaddle._physics_process\` writes \`position.x\` directly without consulting \`drive_blocked\`; that broader fragility is intentionally out of scope here and flagged for a follow-up.

#825

### DIFF
--- a/scripts/core/autoplay_controller.gd
+++ b/scripts/core/autoplay_controller.gd
@@ -16,9 +16,11 @@ func _ready() -> void:
 func toggle() -> void:
 	var desired: bool = not _enabled
 	set_enabled(desired)
-	# `paddle.set_physics_process` mirrors the actual enabled state so we
-	# don't disable player input just because set_enabled refused.
-	paddle.set_physics_process(not _enabled)
+	# Only restore physics_process when the timeout is not active; if it is,
+	# the timeout owns physics_process and will clear it in _finish_at_lane.
+	var timeout_owns: bool = timeout_controller != null and timeout_controller.is_active()
+	if not timeout_owns:
+		paddle.set_physics_process(not _enabled)
 	autoplay_toggled.emit(_enabled)
 
 
--- a/scripts/core/autoplay_controller.gd
+++ b/scripts/core/autoplay_controller.gd
@@ -16,8 +16,7 @@ func _ready() -> void:
 func toggle() -> void:
 	var desired: bool = not _enabled
 	set_enabled(desired)
-	# Only restore physics_process when the timeout is not active; if it is,
-	# the timeout owns physics_process and will clear it in _finish_at_lane.
+	# A running timeout owns physics_process; do not steal the restore back from it.
 	var timeout_owns: bool = timeout_controller != null and timeout_controller.is_active()
 	if not timeout_owns:
 		paddle.set_physics_process(not _enabled)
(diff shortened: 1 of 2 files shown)


scripts/core/autoplay_controller.gd
```diff
@@ -16,9 +16,11 @@ func _ready() -> void:
 func toggle() -> void:
 	var desired: bool = not _enabled
 	set_enabled(desired)
-	# `paddle.set_physics_process` mirrors the actual enabled state so we
-	# don't disable player input just because set_enabled refused.
-	paddle.set_physics_process(not _enabled)
+	# Only restore physics_process when the timeout is not active; if it is,
+	# the timeout owns physics_process and will clear it in _finish_at_lane.
```
J-Melon: **code-quality** nitpick: two-line `#` comment; policy allows one line. Collapse to one, and drop the `_finish_at_lane` reference (no function names in comments).

tests/unit/paddle/test_paddle_auto_play.gd
```diff
@@ -142,3 +142,22 @@ func test_timeout_end_does_not_re_enable_autoplay() -> void:
...
+	assert_false(
+		_paddle.is_physics_processing(),
+		"toggle off during timeout must not re-enable paddle physics_process",
+	)
+	assert_almost_eq(
+		_paddle.position.x,
+		equip_x,
+		0.001,
+		"toggle must not move the paddle while timeout owns it",
+	)
```
J-Melon: **Marvin** nitpick: test proves mid-timeout parking but not the post-completion state; a third assert after `_timeout.end_timeout()` would confirm `_finish_at_lane` restores `physics_process` correctly when a toggle was skipped.

tests/unit/paddle/test_paddle_auto_play.gd
```diff
@@ -142,3 +142,22 @@ func test_timeout_end_does_not_re_enable_autoplay() -> void:
...
+	assert_true(_timeout.is_active(), "precondition: timeout is active")
+	var equip_x: float = _paddle.position.x
+	_controller.toggle()
+	_controller.toggle()
+	assert_false(
+		_paddle.is_physics_processing(),
+		"toggle off during timeout must not re-enable paddle physics_process",
+	)
+	assert_almost_eq(
+		_paddle.position.x,
```
J-Melon: **Trillian** nitpick: `Paddle` has no `_physics_process`, so `position.x` cannot move when physics is re-enabled; this assertion is always green. Drop it or replace with a frame-simulated integration test.

scripts/core/autoplay_controller.gd
J-Melon: **Gru**: compressed to one line in 4887703, dropped the _finish_at_lane reference.

tests/unit/paddle/test_paddle_auto_play.gd
J-Melon: **Gru**: tried it, but end_timeout() doesn't restore synchronously (walk-on phase needs stepping), and the restore is independent of the skipped toggle. The end-timeout restore is already owned by test_timeout_controller.gd:143 (test_end_timeout_restores_main_character_physics). Kept this test focused on the parked-during-timeout behaviour it uniquely proves.

tests/unit/paddle/test_paddle_auto_play.gd
J-Melon: **Gru**: right, dropped the position.x assert in 4887703. Paddle has no _physics_process in the unit context so it couldn't move; the is_physics_processing() assert is the load-bearing discriminator (fails on main, passes with the fix).

PR #925, opened 2026-06-11
Name the three PR-state surfaces in hydrate step

J-Melon:
The dispatch skill's hydrate step named one state surface (the PR fields) and the rule to read the merge queue existed in memory without naming a working command. This adds all three surfaces to the hydrate section: PR fields (read state not mergeable), reviews and inline threads (reviewDecision reads empty even under a bot REQUEST_CHANGES), and the merge queue (the REST endpoint 404s, so it gives the working GraphQL read). The gap was a rule that said check a surface without saying how, so it did not fire.

### DIFF
--- a/.claude/skills/dispatch/SKILL.md
+++ b/.claude/skills/dispatch/SKILL.md
@@ -168,6 +168,11 @@ At most one `spike` issue per swarm dispatch. Run additional spikes sequentially
 
 Before any recap, status report, or claim about challenge state, run `gh pr list --state open --json number,state,mergeable,labels,headRefOid` (or `gh pr view <n> --json ...` for a specific challenge). Don't recap from in-context memory of the last dispatch; dispatches and merges can happen between turns. The first action of any state-summary turn is the hydrate command, not text.
 
+State has three surfaces; read whichever the claim rests on, not just the first:
+- **PR fields** (above): read `state` and `mergedAt` for done/open, never `mergeable` (it reads `UNKNOWN` post-merge); `headRefOid` for the live SHA; `statusCheckRollup` for CI.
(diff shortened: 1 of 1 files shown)


.claude/skills/dispatch/SKILL.md
```diff
@@ -168,6 +168,11 @@ At most one `spike` issue per swarm dispatch. Run additional spikes sequentially
 
 Before any recap, status report, or claim about challenge state, run `gh pr list --state open --json number,state,mergeable,labels,headRefOid` (or `gh pr view <n> --json ...` for a specific challenge). Don't recap from in-context memory of the last dispatch; dispatches and merges can happen between turns. The first action of any state-summary turn is the hydrate command, not text.
 
+State has three surfaces; read whichever the claim rests on, not just the first:
+- **PR fields** (above): read `state` and `mergedAt` for done/open, never `mergeable` (it reads `UNKNOWN` post-merge); `headRefOid` for the live SHA; `statusCheckRollup` for CI.
+- **Reviews and threads:** `gh api repos/.../pulls/<n>/reviews` for the real verdict (`reviewDecision` reads empty even when a bot REQUEST_CHANGES exists), and `.../comments` for inline findings (`position: null` means outdated after a push).
```
J-Melon: **Prose reviewer** issue: `reviewDecision` is empty here because no required-reviewer rule is configured on main, not a bot-review quirk; fix the parenthetical to name the real cause.

J-Melon: Fixed in 849247f: the parenthetical now names the cause as main having no required-review branch protection (so reviewDecision never populates), not an API quirk. Advice to read .../reviews directly is unchanged.

PR #936, opened 2026-06-12
Fail loud when memory render runs against the wrong dir

J-Melon:
The memory boot command rendered the forest with a cwd-relative `.` argument. From the repo root, where every session starts, that resolved `MEMORY_DIR` to a directory holding no forest, and `lint-graph-edges.sh --tree` then printed a flat dump of every loose markdown file as if it were a broken corpus. A fresh session reads that as the memory's true state.

This guards tree mode: when the target dir holds no `MEMORY.md`/trunk spine, the script exits 2 with a "wrong dir?" message instead of rendering a plausible-but-empty forest. The companion fix to the boot command itself (dropping the `.`) lives in the memory repo.

### DIFF
--- a/scripts/memory/lint-graph-edges.sh
+++ b/scripts/memory/lint-graph-edges.sh
@@ -61,6 +61,15 @@ while IFS= read -r -d '' filepath; do
     PARENT_OF["$filename"]="$parent_value"
 done < <(find "$MEMORY_DIR" -name "*.md" -print0 | sort -z)
 
+# A real memory dir always holds the trunk spine. If MEMORY.md and the trunks
+# are absent, MEMORY_DIR points at the wrong tree (e.g. the repo root when the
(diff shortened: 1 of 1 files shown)


scripts/memory/lint-graph-edges.sh
```diff
@@ -61,6 +61,15 @@ while IFS= read -r -d '' filepath; do
     PARENT_OF["$filename"]="$parent_value"
 done < <(find "$MEMORY_DIR" -name "*.md" -print0 | sort -z)
 
+# A real memory dir always holds the trunk spine. If MEMORY.md and the trunks
+# are absent, MEMORY_DIR points at the wrong tree (e.g. the repo root when the
```
J-Melon: comment too big

J-Melon: Cut the four-line block to one line naming the non-obvious WHY (no spine means wrong dir), in 74f414b3.

PR #937, opened 2026-06-12
ci: add 500KB binary size gate via lefthook and CI

J-Melon:
Adds a pre-commit hook and a CI PR check that block plain binary files over 500 KB from entering git history. Both surfaces call the same `scripts/ci/check_asset_size.sh` script so the threshold is defined once. The `.import` sidecars and LFS pointer files are excluded. This adds a new CI workflow (`asset-size.yml`) that runs on every pull request.

#819

Agent-Role: Norbert (implementer)

### DIFF
--- a/.github/workflows/asset-size.yml
+++ b/.github/workflows/asset-size.yml
@@ -26,4 +26,10 @@ jobs:
           fetch-depth: 0
 
       - name: Check asset sizes
-        run: bash scripts/ci/check_asset_size.sh --diff ${{ github.event.pull_request.base.sha }}
+        run: |
+          if [[ "${{ github.event_name }}" == "merge_group" ]]; then
+            base_sha="${{ github.event.merge_group.base_sha }}"
+          else
+            base_sha="${{ github.event.pull_request.base.sha }}"
+          fi
+          bash scripts/ci/check_asset_size.sh --diff "$base_sha"
--- a/.github/workflows/asset-size.yml
+++ b/.github/workflows/asset-size.yml
@@ -1,0 +1,29 @@
+name: Asset size gate
+
+on:
+  pull_request:
+    branches:
+      - main
+  merge_group:
+
+concurrency:
+  group: asset-size-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  check-asset-size:
+    name: check-asset-size
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    permissions:
+      contents: read
+    steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/asset-size.yml
```diff
@@ -0,0 +1,29 @@
...
+    timeout-minutes: 5
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          fetch-depth: 0
+
+      - name: Check asset sizes
+        run: bash scripts/ci/check_asset_size.sh --diff ${{ github.event.pull_request.base.sha }}
```
J-Melon: **Stuart** issue: `merge_group` has no `.pull_request` object, so `base.sha` is empty; `--diff` receives no ref and the step aborts on every merge queue entry. Branch on `github.event_name` to supply `github.event.merge_group.base_sha`, or remove the `merge_group` trigger.

CONTRIBUTING.md
```diff
@@ -32,6 +32,8 @@ lefthook install
 
 CI runs the same checks on every PR, so whatever passes locally will pass there.
 
+A pre-commit hook also enforces a **500 KB size gate** on staged files. Any binary over that threshold must be tracked via Git LFS rather than committed directly. The hook skips `.import` sidecars (generated config) and files already tracked by LFS. CI runs the same check on every pull request as a backstop for skipped local hooks.
```
J-Melon: **Jerry** issue: `no_auto_install: true` means existing contributors who pull this change will not get the new hook unless they re-run `lefthook install`; add a sentence telling them to do so after pulling hook changes.

.github/workflows/asset-size.yml
J-Melon: **Kevin**

resolved: 9cb163a branches on `github.event_name` and supplies `github.event.merge_group.base_sha` for the merge_group path, so the guard never sees an empty string.

CONTRIBUTING.md
J-Melon: **Kevin**

resolved: 9cb163a appends a sentence to the hooks paragraph telling contributors to re-run `lefthook install` after pulling changes that add or remove hooks.

J-Melon: **Jerry**

resolved: 9cb163a appends the re-run instruction to the existing hooks paragraph; block cleared.

.github/workflows/asset-size.yml
J-Melon: **Stuart**

resolved: 9cb163a2 branches on `event_name` to supply `merge_group.base_sha`; empty-sha bug is gone, no new defects.

PR #938, opened 2026-06-12
Record the LFS proxy architecture

J-Melon:
Refs #928.

The R2-backed Git LFS pipeline needed its implementation architecture pinned before the repo wiring lands. This records the decision: a standalone wrangler Worker with a native R2 binding, rather than the off-the-shelf Pages proxy the spike named, so the write credential stays off contributor config and the Worker matches the studio's existing deploy pattern. It captures the auth model (a published download key seeing release only, guarded by a Worker rate limit; an upload key for studio and CI; a distinct promote key as the only writer of release), the preview/release prefixes that keep the bucket free of orphaned bytes, the live infra now standing, and how CI integrates.

The Worker is built, deployed, and verified; this doc is the written record the remaining repo-wiring challenge builds against.

### DIFF
--- a/designs/01-prototype/tech/2026-06-12-lfs-proxy-architecture.md
+++ b/designs/01-prototype/tech/2026-06-12-lfs-proxy-architecture.md
@@ -132,8 +132,9 @@ additionally run the promote step on merge to main. New steps stay SHA-pinned, m
 
 Account `volcanoem@gmail.com`, ID `effe9646943c4ead286bad9d06e16e74`.
 
-- Worker deployed at `https://volley-lfs-proxy.volcanoem.workers.dev`. The single-path version is
-  verified end to end (auth matrix, presigned signing, R2 round trip); the prefix split is in rework.
+- Worker deployed at `https://volley-lfs-proxy.volcanoem.workers.dev`, verified end to end (auth
+  matrix, presigned signing, R2 round trip). The `preview/`/`release/` build is tested and lands live on
+  the next deploy with the `PROMOTE_KEY` secret.
 - R2 bucket `volley-assets`, region WEUR, Standard storage class.
 - S3 endpoint `https://effe9646943c4ead286bad9d06e16e74.r2.cloudflarestorage.com`.
 - R2 API token: Account token, Object Read and Write, scoped to `volley-assets`, no expiry. Its Access
--- a/designs/01-prototype/tech/2026-06-12-lfs-proxy-architecture.md
+++ b/designs/01-prototype/tech/2026-06-12-lfs-proxy-architecture.md
@@ -2,79 +2,126 @@
 
 Implementation record for SH-489, standing up the R2-backed Git LFS pipeline the
 [asset-versioning spike](2026-06-11-asset-versioning-spike.md) decided. The spike chose the backend
-(Cloudflare R2 behind an LFS proxy); this record fixes how the proxy is built and captures the live
-infra it runs on.
+(Cloudflare R2 behind an LFS proxy); this record fixes how the proxy is built, how the bucket stays
+free of orphaned bytes, and the live infra it runs on.
 
 ## Proxy: a wrangler Worker with an R2 binding
 
-The spike named the off-the-shelf `git-lfs-s3-proxy` (a Cloudflare Pages deploy that bakes the R2
-credentials into each client's `.lfsconfig` URL). The build instead uses a **standalone wrangler
-Worker with a native R2 binding**, in its own repo `volley-lfs-proxy`, matching the existing
-`linear-mcp` and `nano-banana-mcp` Workers. Two reasons carry the choice: it holds the R2 keys as
-Worker secrets, so the write credential stays off every contributor's `.lfsconfig` (the ticket's hard
-requirement), and it deploys the way the studio already deploys Workers.
+The build uses a standalone wrangler Worker with a native R2 binding, in its own project
+`volley-lfs-proxy`, matching the existing `linear-mcp` and `nano-banana-mcp` Workers. The load-bearing
+reason is deploy-pattern match: the studio already ships Workers this way, and the Worker holds the R2
+keys as Worker secrets so the write credential stays off contributor config. (The off-the-shelf
+`git-lfs-s3-proxy` is also a Worker and could hold secrets the same way; the deciding factor is the
+existing toolchain, not a credential property.)
 
-The Worker implements the Git LFS batch API and presigns S3-compatible R2 URLs, so object bytes
-stream directly to and from R2 and never pass through the Worker (whose request body caps at 100MB,
-below a single large asset). This follows the FlyByWire Simulations R2 LFS Worker pattern.
+The Worker implements the Git LFS batch API and presigns S3-compatible R2 URLs, so object bytes stream
+directly to and from R2 and never pass through the Worker (whose request body caps at 100MB, below a
+single large asset). This follows the FlyByWire Simulations R2 LFS Worker pattern.
 
-Three keys separate read from write at the Worker layer, each carried as a Worker secret:
+## Two keys
 
 | Key | Grants | Held by |
 |---|---|---|
-| `DOWNLOAD_KEY` | download, list | published openly; CI, every cloner |
-| `UPLOAD_KEY` | download and upload | studio only, given out on request |
+| `DOWNLOAD_KEY` | download | published; CI, every cloner |
+| `UPLOAD_KEY` | upload | studio and CI; given out on request |
 
-The LFS `basic` transfer with presigned PUT URLs needs no separate verify action, so the Worker
-carries only the two keys above. The `UPLOAD_KEY` is never published or stored in a repo file; a studio
-member who needs to push art asks for it.
+The LFS `basic` transfer with presigned PUT URLs needs no verify action, so the Worker carries two
+keys. The Worker exposes no list endpoint, so a holder of the download key can fetch an object whose
+oid they know but cannot enumerate the bucket. `UPLOAD_KEY` lives only as a Worker secret and in the
+hands of people who push art; it is never published or committed.
 
-## The download key is public; the Worker is the guard
+## The download key is public; the Worker guards the tier
 
-The audience for fetching `assets/` is anyone who clones, so the `DOWNLOAD_KEY` is treated as a
-published value, printed in CONTRIBUTING and carried in `.lfsconfig.example`. Its openness is fine
-because secrecy is not what protects the bucket: the Worker enforces a rate limit and a per-window
-byte cap on the download path, so a key in a scraper's hands still cannot burn the R2 free tier.
-CI reads the same key from a `gh` Actions secret (`LFS_DOWNLOAD_KEY`) to keep the workflow file clean,
-not because the value is confidential. Only `UPLOAD_KEY` and the R2 Secret Access Key stay truly
-secret, as Worker secrets the studio holds.
+Anyone who clones fetches `assets/`, so `DOWNLOAD_KEY` is a published value, printed in CONTRIBUTING.
+Secrecy is not the protection: the Worker rate-limits the download path (a fixed window of
+`DOWNLOAD_RATE_LIMIT` requests per `DOWNLOAD_RATE_WINDOW_SECONDS` per client IP), so an open key in a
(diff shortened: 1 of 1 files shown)


designs/01-prototype/tech/2026-06-12-lfs-proxy-architecture.md
```diff
@@ -0,0 +1,80 @@
...
+[asset-versioning spike](2026-06-11-asset-versioning-spike.md) decided. The spike chose the backend
+(Cloudflare R2 behind an LFS proxy); this record fixes how the proxy is built and captures the live
+infra it runs on.
+
+## Proxy: a wrangler Worker with an R2 binding
+
+The spike named the off-the-shelf `git-lfs-s3-proxy` (a Cloudflare Pages deploy that bakes the R2
+credentials into each client's `.lfsconfig` URL). The build instead uses a **standalone wrangler
+Worker with a native R2 binding**, in its own repo `volley-lfs-proxy`, matching the existing
+`linear-mcp` and `nano-banana-mcp` Workers. Two reasons carry the choice: it holds the R2 keys as
```
J-Melon: **issue:** `git-lfs-s3-proxy` can also hold credentials as Worker secrets; the doc characterises a deployment pattern as an inherent property. The real reason to prefer the bespoke Worker is deploy-pattern match, not credential architecture. State that plainly or drop the comparison.

designs/01-prototype/tech/2026-06-12-lfs-proxy-architecture.md
```diff
@@ -0,0 +1,80 @@
...
+| Key | Grants | Held by |
+|---|---|---|
+| `DOWNLOAD_KEY` | download, list | published openly; CI, every cloner |
+| `UPLOAD_KEY` | download and upload | studio only, given out on request |
+
+The LFS `basic` transfer with presigned PUT URLs needs no separate verify action, so the Worker
+carries only the two keys above. The `UPLOAD_KEY` is never published or stored in a repo file; a studio
+member who needs to push art asks for it.
+
+## The download key is public; the Worker is the guard
```
J-Melon: **issue:** R2 is not an atomic counter. Concurrent Workers read stale values; a burst from one IP can exceed the window limit before any increment propagates. Accept the worst-case overage explicitly, or switch to Durable Objects / Workers KV for the counter.

J-Melon: **issue:** Rate-limit values (requests/window, byte cap) are referenced but not recorded. A design doc for live infra must include the actual configured limits so a future operator can verify the guard is calibrated.

designs/01-prototype/tech/2026-06-12-lfs-proxy-architecture.md
```diff
@@ -0,0 +1,80 @@
...
+requirement), and it deploys the way the studio already deploys Workers.
+
+The Worker implements the Git LFS batch API and presigns S3-compatible R2 URLs, so object bytes
+stream directly to and from R2 and never pass through the Worker (whose request body caps at 100MB,
+below a single large asset). This follows the FlyByWire Simulations R2 LFS Worker pattern.
+
+Three keys separate read from write at the Worker layer, each carried as a Worker secret:
+
+| Key | Grants | Held by |
+|---|---|---|
```
J-Melon, line 24: **issue:** The grants table gives DOWNLOAD_KEY "list". LFS batch has no list verb. If list is not surfaced by the Worker, remove it from the table. If it is, name it as a bucket-enumeration surface -- a scraper can walk all object keys without knowing paths.

designs/01-prototype/tech/2026-06-12-lfs-proxy-architecture.md
```diff
@@ -0,0 +1,80 @@
...
+   `volley-assets`, the batch handler, vitest suite. **Done and deployed**, verified end to end.
+2. **Cloudflare account setup**: **done**. R2 enabled, bucket created, token minted, secrets set.
+3. **volley repo wiring** (separate challenge, the remaining work): committed `.gitattributes` tracking
+   `assets/**` and `concepts/**`; committed `.lfsconfig` carrying the Worker URL with the published
+   download key (so a clone is zero-config); `make concepts`; `concepts/` un-gitignored; an `assets/`
+   directory; CONTRIBUTING setup, published key, and rate-limit note.
+
+### CI integration
+
+`test.yml`, `publish.yml`, and `release.yml` each run `godot --headless --import` (Import Project),
```
J-Melon: **issue:** The claim that test.yml needs LFS bytes because import "walks the whole tree" is unverified. Godot skips missing assets and continues; the actual question is whether any test assertion or autoload requires a real asset byte. Show the specific failure before adding LFS pull to test.yml.

designs/01-prototype/tech/2026-06-12-lfs-proxy-architecture.md
```diff
@@ -0,0 +1,80 @@
...
+## Live infra
+
+Account `volcanoem@gmail.com`, ID `effe9646943c4ead286bad9d06e16e74`.
+
+- Worker deployed at `https://volley-lfs-proxy.volcanoem.workers.dev`, verified end to end (auth matrix,
+  presigned signing, R2 round trip).
+- R2 bucket `volley-assets`, region WEUR, Standard storage class.
+- S3 endpoint `https://effe9646943c4ead286bad9d06e16e74.r2.cloudflarestorage.com`.
+- R2 API token: Account token, Object Read and Write, scoped to `volley-assets`, no expiry. Access Key
+  ID `9d7317e3a2fb0ce6dd2f3f5e4a0e217a`. The Secret Access Key is held by Josh and enters the Worker
```
J-Melon: **nitpick:** R2 Access Key ID is committed to the repo. Record instead that a token was minted and where the ID is stored. Low severity (no secret), but sets a bad precedent.

designs/01-prototype/tech/2026-06-12-lfs-proxy-architecture.md
J-Melon: Fixed in 8c4aff42: the doc now states deploy-pattern match as the load-bearing reason and notes parenthetically that the off-the-shelf proxy could hold secrets the same way, dropping the misleading credential contrast.

designs/01-prototype/tech/2026-06-12-lfs-proxy-architecture.md
J-Melon: Fixed in 8c4aff42: the doc now names the non-atomic read-increment failure explicitly, bounds the accepted worst case to a single-digit multiple of the limit (absorbed by the free tier), and names Durable Objects / Workers KV as the escalation if real abuse appears.

J-Melon: Fixed in 8c4aff42: the rate limit is now stated as DOWNLOAD_RATE_LIMIT requests per DOWNLOAD_RATE_WINDOW_SECONDS per client IP, the configured vars.

designs/01-prototype/tech/2026-06-12-lfs-proxy-architecture.md
J-Melon: Fixed in 8c4aff42: the Worker exposes no list endpoint, so the download key cannot enumerate the bucket. The grants table now reads download only; the stray 'list' is dropped.

designs/01-prototype/tech/2026-06-12-lfs-proxy-architecture.md
J-Melon: Fixed in 8c4aff42 with the verified mechanism: the repo's Leak gate fails the run on a standalone 'Failed loading resource' error, and tested scenes reference files under assets/, so a pointer-only checkout fails the test run. Corrected from the vague 'walks the whole tree'.

designs/01-prototype/tech/2026-06-12-lfs-proxy-architecture.md
J-Melon: Fixed in 8c4aff42: the Access Key ID is out of the doc body; the doc now records that a token was minted and that its ID lives in the Worker's wrangler.jsonc vars.

designs/01-prototype/tech/2026-06-12-lfs-proxy-architecture.md
```diff
@@ -0,0 +1,127 @@
...
+
+Account `volcanoem@gmail.com`, ID `effe9646943c4ead286bad9d06e16e74`.
+
+- Worker deployed at `https://volley-lfs-proxy.volcanoem.workers.dev`. The single-path version is
+  verified end to end (auth matrix, presigned signing, R2 round trip); the prefix split is in rework.
+- R2 bucket `volley-assets`, region WEUR, Standard storage class.
+- S3 endpoint `https://effe9646943c4ead286bad9d06e16e74.r2.cloudflarestorage.com`.
+- R2 API token: Account token, Object Read and Write, scoped to `volley-assets`, no expiry. Its Access
+  Key ID lives in the Worker's `wrangler.jsonc` vars; its Secret Access Key is held by Josh and set as a
+  Worker secret, never committed.
```
J-Melon, line 142: **bot-review (re-battle pass)**

All six prior findings are resolved.

1. Bespoke rationale: deploy-pattern match is now the stated reason; the credential-contrast parenthetical correctly concedes the off-the-shelf proxy could hold secrets equally. Resolved.
2. Atomicity / overshoot bound: the failure mode is named, the worst case is bounded (single-digit multiple), and the escalation path (DO/KV) is named. Resolved.
3. Rate-limit values: anchored to the env-var names rather than prose constants. Resolved.
4. List grant / enumeration: the grants table has no list row, and the no-list-endpoint statement is explicit. Resolved.
5. CI failure mechanism: the Leak gate chain is spelled out; the dependency on tested scenes referencing assets/ is clear. Resolved.
6. Access Key ID: the Key ID is out of the doc body. The account ID appears in the S3 endpoint and the infra block; that is a lower-sensitivity identifier and the specific nitpick is addressed.

---

**New design: preview/ and release/ prefix model -- four blocks**

**block: grace window is unspecified**

Line 57 sets the lifecycle expiry at "e.g. 14 days". That "e.g." is doing real work: the grace window is the operational parameter that determines whether an open PR can still build. A PR open longer than the window will have its preview/ objects reaped; download step 3 falls back to preview/<oid> and finds nothing; CI on that branch fails with a hard LFS error on a resource that was validly pushed. The doc must commit to a value (or a stated lower bound) and state the policy for PRs that exceed it -- close them, re-push the objects, or something else.

**block: download-key reach into preview/ vs. concepts/ upload path**

Download step 3 presigns preview/<oid> transparently for any oid the caller knows. The Two Keys section says DOWNLOAD_KEY is published. The tracking section says concept art is tracked by .gitattributes. If concept art is uploaded through the Worker (upload presigns to preview/<oid>), any holder of the published DOWNLOAD_KEY can fetch unreleased concept art for any oid they can guess or obtain. The CI section correctly excludes concepts/ from git lfs pull, but that is a pull-side filter, not an upload-side one. The doc must either state that concepts/ objects never enter the Worker upload path (and explain how -- a Worker-side path check, a separate .lfsconfig for concepts/, or a policy), or acknowledge that preview/ objects for all tracked paths are reachable with the public key and accept that consequence.

**block: "CI is the only writer of release/" is not technically enforced**

The claim is stated as a design property (line 68). The enforcement path given is that CI holds UPLOAD_KEY for the promote step. But the Two Keys section also says UPLOAD_KEY is "given out on request" to people who push art. Anyone with UPLOAD_KEY can call the Worker. Whether the Worker rejects a presign request for a release/ path from a non-CI caller is not stated. If it does not, the property holds only by convention. Either the Worker must unconditionally route all upload presigns to preview/ (regardless of what the caller requests) and expose a separate promote endpoint that CI calls with a distinct credential -- or the claim must be softened to "by convention" and the risk accepted. The current text implies a technical guarantee that the design does not yet ground.

**block: promote-step partial failure leaves main with unreachable history**

A promote job that copies 15 of 20 objects before failing leaves main referencing 5 objects that exist only in preview/. This silently works while preview/ holds them, then permanently breaks when the lifecycle rule fires. The doc does not state whether the promote step fails the CI run visibly, whether it is idempotent, or what the recovery path is. At minimum: the step must be idempotent (copy is a no-op if the destination exists, which R2 copy supports), the CI job must fail loudly on any per-object copy error, and a retry must be safe.

PR #939, opened 2026-06-12
ci: wire repo to R2 Git LFS proxy for asset versioning

J-Melon:
Reworks the R2 LFS repo wiring to match the merged proxy architecture. The superseded version used a blanket `assets/**` LFS tracking pattern and put the download key in CONTRIBUTING; this version corrects both.

`.gitattributes` now tracks by large-by-nature path: `concepts/**` wholesale, plus source-art formats (`*.psd`, `*.aseprite`, `*.ase`, `*.kra`), audio (`*.wav`, `*.ogg`, `*.mp3`, `*.flac`), and large-texture formats (`*.webp`, `*.exr`, `*.hdr`). Plain `*.png` and `*.svg` are untracked, leaving the small committed assets in git. The size gate (#937) is the backstop.

`.lfsconfig` is committed with the download key embedded in the URL, making clones zero-config. CONTRIBUTING carries only `git lfs install`, clone, `make concepts` for art, and a rate-limit note; no explicit key step because none is needed.

Both `publish.yml` and `release.yml` now have a `promote-lfs` job that POSTs merged LFS oids to `/promote` with Basic auth. The `publish.yml` job runs only when `github.event_name == 'push' && github.ref == 'refs/heads/main'`, never on `pull_request`. The `release.yml` job runs on the `release: published` event (maintainer-context only). Neither can fire in a fork PR context.

#928

**Secret required:** `LFS_PROMOTE_KEY` must be set as a repo secret before the promote jobs will authenticate. `LFS_DOWNLOAD_KEY` is also required for the CI fetch steps (the download key is published by design in `.lfsconfig` but CI reads it from the secret for a clean workflow file).

Agent-Role: infra-implementer


CONTRIBUTING.md
```diff
@@ -16,6 +16,32 @@ Sometimes maintainers will pick up a ticket themselves to hit a deadline. When t
...
 
+### Asset setup (Git LFS)
+
+Large binaries under `assets/` and `concepts/` are stored in Git LFS, fetched through a Cloudflare
+R2 proxy. One-time setup:
+
+```sh
+git lfs install
+```
+
```
J-Melon, line 27: **Gabbro** issue: the raw key block contradicts the zero-config claim two lines above; cut it and let `.lfsconfig` be the one source.

.github/workflows/release.yml
```diff
@@ -63,6 +76,31 @@ jobs:
           mkdir -p logs
           godot --headless --path . -s addons/gut/gut_cmdln.gd -gconfig=.gutconfig.json -gexit
```
J-Melon, line 78: **Riebeck** issue: `promote-lfs` belongs in `publish.yml` (push to main), not here. `release: published` fires later and on-demand; merges to main will never promote, leaving objects in `preview/` until the 30-day expiry.

CONTRIBUTING.md
J-Melon: **Nomai** 329df96 — the current file has no raw key block; the  code block at that line is a shell command, not a credential. The section is already zero-config clean. No change needed here.

.github/workflows/release.yml
J-Melon: **Nomai** 329df96 — removed promote-lfs from release.yml. Promote on merge to main is publish.yml's job; the copy here was redundant and would have missed plain merges.

PR #948, opened 2026-06-12
docs: wire studio LFS upload setup

J-Melon:
The committed `.lfsconfig` carries the download key, which the Worker's upload path rejects. A studio member with an upload key had no documented path to push LFS bytes.

This change adds a `make lfs-upload-setup` target that takes the upload key as an environment variable and writes it to the local `.git/config` as `lfs.url`. The local git config overrides `.lfsconfig` for that machine only; every other clone continues pulling through the committed download key. The target guards against a missing key (exits non-zero with a clear message) and prints confirmation that the override is local and not committed. CONTRIBUTING gets a new "Pushing LFS art (studio members)" subsection that walks through the one-time setup, explains the preview/release promotion flow, and includes a safety note against staging the key.

No new secret is introduced in any committed file. The upload key is supplied by the user at the shell and written only to `.git/config`, which git never stages.

AC4 (real `concepts/` push verifying the end-to-end path with the live upload key) is a manual follow-up by Josh and Gru with the live key. It is not part of this PR.

#947

Agent-Role: infra-implementer

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,6 +34,41 @@ Concept art under `concepts/` is opt-in and not fetched on clone. Pull it when y
 make concepts
 ```
 
+### Pushing LFS art (studio members)
+
+If you have been given an upload key you can push new LFS-tracked art from your branch. The
+upload key is never committed; it lives only in your local `.git/config` for the one machine you
+push from.
+
+One-time local setup (obtain your key from a maintainer out of band):
+
+```sh
+LFS_UPLOAD_KEY=<your-key> make lfs-upload-setup
+```
+
+That runs `git config lfs.url ...` with your key and writes the result to `.git/config`, which
+git never commits. Every other clone keeps pulling through the committed `.lfsconfig` download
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -34,6 +34,41 @@ Concept art under `concepts/` is opt-in and not fetched on clone. Pull it when y
...
+git config lfs.url
+```
+
+To remove the override and revert to download-only:
+
+```sh
+git config --unset lfs.url
+```
+
+**Never commit your upload key.** If `git config lfs.url` shows in a staged diff, run
```
J-Melon: **Gabbro** issue: this warns to check a staged diff for `git config lfs.url`, but `.git/config` is never staged, so the check cannot fire and gives false reassurance. Point at `.lfsconfig` instead.

J-Melon: Fixed in fe5639d8: the warning now names the real risk (an accidental `.lfsconfig` edit, restore with `git checkout .lfsconfig`), and a line notes the verify output contains the key.

PR #949, opened 2026-06-12
Add concept art into subject folders via LFS

J-Melon:
First real use of the R2 LFS pipeline: the concept art batch, organised into subject folders (garden, character, movement, items, soul) with raw masters under concepts/raw, and the early pitch board at the root. Every binary is an LFS pointer; the 95MB of real bytes uploaded to the bucket preview prefix and verified retrievable.

Two pipeline gaps closed alongside: .gitattributes now routes video and animation formats (gif, apng, mp4, webm, mov) to LFS so frame-by-frame work in assets does not land in plain git, and .gitignore now covers the autosave and temp cruft for the main art and animation apps.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,9 +65,12 @@ ai/scratchpads/
 *.kra~
 *.psd~
 *.xcf~
-*.{psb,psd}.tmp
-*.{ai,svg}.tmp
-~*.{ai,psd}
+*.psb.tmp
+*.psd.tmp
+*.ai.tmp
+*.svg.tmp
+~*.ai
+~*.psd
 .~lock.*
 *.aseprite-data
 *.ase~
@@ -88,7 +91,7 @@ ai/scratchpads/
 *.pcl~
 *_bak.pclx
 backup/*.tnz
-sandbox/
+/sandbox/
 
 # OS junk
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -60,6 +60,36 @@ ai/swarm/_bootstrap.md
 # the injection-guard hook log. Not committed.
 ai/scratchpads/
 
+# Art-app autosaves and temp files (commit the source masters, not the crash recovery)
+*-autosave.kra
+*.kra~
+*.psd~
+*.xcf~
+*.{psb,psd}.tmp
+*.{ai,svg}.tmp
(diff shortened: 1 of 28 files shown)


.gitignore
```diff
@@ -60,6 +60,36 @@ ai/swarm/_bootstrap.md
 # the injection-guard hook log. Not committed.
 ai/scratchpads/
 
+# Art-app autosaves and temp files (commit the source masters, not the crash recovery)
+*-autosave.kra
+*.kra~
```
J-Melon, line 65: **Hornfels** issue: gitignore does not support brace expansion; `*.{psb,psd}.tmp`, `*.{ai,svg}.tmp`, and `~*.{ai,psd}` match only their literal strings. Expand each into separate lines.

.gitignore
```diff
@@ -60,6 +60,36 @@ ai/swarm/_bootstrap.md
...
+*.dbbak
+*.aep~
+*.aep.lock
+*.moho~
+# Frame-by-frame apps (OpenToonz, TVPaint, Pencil2D) autosaves and backups
+*.tnz~
+*.tvpp~
+*.pclx~
+*.pcl~
+*_bak.pclx
```
J-Melon, line 92: **Hornfels** suggestion: `sandbox/` is unscoped and ignores any directory named `sandbox` anywhere in the tree; consider scoping to `/sandbox/` if only the repo root is intended.

.gitignore
J-Melon: Confirmed by an empirical git test: `*.{psb,psd}.tmp` matches only a literal-brace filename, not `foo.psd.tmp`. Expanding all three brace patterns to separate lines. Also folding the README-as-LFS-pointer fix (exclude `*.md` from concepts LFS) and adding `*.ogv`.

.gitignore
J-Melon: Scoping to `/sandbox/` (repo root only).

.gitignore
J-Melon: Fixed in 092d3a03: brace patterns expanded to separate lines (verified foo.psd.tmp now ignores), README and .gdignore excluded from concepts LFS so they stay plain text, .ogv added.

PR #955, opened 2026-06-12
feat: Settings autoload with ConfigFile persistence

J-Melon:
Adds a `Settings` autoload that loads `user://settings.cfg` at `_enter_tree`, applies audio bus volumes and display properties, and persists on every confirmed change and at `_exit_tree`. Defaults for resolution list, FPS-cap options, bus volumes, and window mode live in a `SettingsDefaults` Resource rather than inline consts, so they are inspector-editable. The autoload is registered above `SaveManager` so settings apply before gameplay systems boot.

Closes #931


scripts/core/settings.gd
```diff
@@ -0,0 +1,128 @@
+extends Node
+
+const _CFG_PATH := "user://settings.cfg"
+const _SEC_DISPLAY := "display"
+const _SEC_AUDIO := "audio"
```
J-Melon: **Marvin** nitpick: `_CFG` and `_SEC` are abbreviations; Volley naming requires full words (`_CONFIG_PATH`, `_SECTION_DISPLAY`, `_SECTION_AUDIO`).

scripts/core/settings.gd
```diff
@@ -0,0 +1,128 @@
...
+
+	_set_bus_volume("Master", master)
+	_set_bus_volume("Music", music)
+	_set_bus_volume("SFX", sfx)
+
+
+func _apply_display() -> void:
+	var mode: int = _config.get_value(_SEC_DISPLAY, "window_mode", _defaults.default_window_mode)
+	var vsync: bool = _config.get_value(_SEC_DISPLAY, "vsync", _defaults.default_vsync)
+	var fps_cap: int = _config.get_value(_SEC_DISPLAY, "fps_cap", _defaults.default_fps_cap)
```
J-Melon: **Marvin** issue: resolution fallback is a hardcoded literal while every other default reads `_defaults`; add a `default_resolution` export to `SettingsDefaults` and read it here.

scripts/core/settings.gd
```diff
@@ -0,0 +1,128 @@
...
+	match section:
+		_SEC_AUDIO:
+			_apply_audio_key(key, value)
+		_SEC_DISPLAY:
+			_apply_display_key(key, value)
+
+
+func _apply_audio_key(key: String, value: Variant) -> void:
+	match key:
+		"master_volume":
```
J-Melon, line 93: **Marvin** issue: `_apply_audio_key` duplicates every effect already in `_apply_audio`; a fourth bus key would require two updates; have `_apply_audio` drive `_apply_audio_key` per key so each effect lives once.

tests/unit/settings/test_settings.gd
```diff
@@ -0,0 +1,65 @@
...
+	assert_true(FileAccess.file_exists(_tmp_path), "settings file exists at its own path")
+
+	var save_path := "user://save.json"
+	assert_true(_tmp_path != save_path, "settings path is separate from the gameplay save path")
+
+
+func test_set_value_no_op_on_same_value() -> void:
+	_settings.set_value("display", "fps_cap", 60)
+	_settings.set_value("display", "fps_cap", 60)
+
```
J-Melon, line 63: **Marvin** nitpick: `_tmp_path != save_path` is a string comparison between a random filename and a constant; it always passes regardless of implementation and cannot catch a regression.

tests/unit/settings/test_settings.gd
```diff
@@ -0,0 +1,65 @@
...
+	var reloaded: Node = _SETTINGS_SCRIPT.new(_tmp_path)
+	add_child_autofree(reloaded)
+
+	var value: float = reloaded.get_value("audio", "master_volume", 1.0)
+	assert_almost_eq(value, 0.5, 0.001, "set value survives a fresh load from disk")
+
+
+func test_settings_file_is_independent_of_save_manager() -> void:
+	_settings.set_value("audio", "sfx_volume", 0.3)
```
J-Melon, line 44: **Marvin** issue: the save-independence test never wipes the save; the path-inequality check (`_tmp_path != save_path`) is structurally always true. Prove the AC by deleting (or never creating) the save file and asserting settings still read back.

scripts/core/settings.gd
J-Melon: **Riebeck**\n\nresolved: 229dc14 renames _CFG_PATH to _CONFIG_PATH and _SEC_* to _SECTION_*

scripts/core/settings.gd
J-Melon: **Riebeck**\n\nresolved: 229dc14 adds default_resolution to SettingsDefaults; _apply_display reads _defaults.default_resolution

scripts/core/settings.gd
J-Melon: **Riebeck**\n\nresolved: 229dc14 _apply_audio and _apply_display now iterate keys and call the per-key helpers; no duplicated effects

tests/unit/settings/test_settings.gd
J-Melon: **Riebeck**\n\nresolved: 229dc14 folds into the save-independence fix below

tests/unit/settings/test_settings.gd
J-Melon: **Riebeck**\n\nresolved: 229dc14 test now deletes the save file then asserts the setting still reads back correctly

tests/unit/settings/test_settings.gd
```diff
@@ -0,0 +1,70 @@
...
+	var save_path := "user://save_data.json"
+
+	if FileAccess.file_exists(save_path):
+		DirAccess.remove_absolute(ProjectSettings.globalize_path(save_path))
+
+	assert_false(FileAccess.file_exists(save_path), "save file absent after deletion")
+
+	var value: float = _settings.get_value("audio", "sfx_volume", 1.0)
+	assert_almost_eq(value, 0.3, 0.001, "setting survives save-file wipe")
+
```
J-Melon, line 58: **Marvin** issue: reads from the same in-memory `_settings` instance; in-memory state survives a file deletion regardless of coupling. Construct a new `_SETTINGS_SCRIPT.new(_tmp_path)` after the wipe and read from that, matching the pattern in `test_set_value_persists_across_reload`.

J-Melon: **Marvin**

resolved: bec2d1ca constructs a cold Settings instance after the save wipe and reads from that.

PR #976, opened 2026-06-16
fix: reject back-face racket hits and enforce animation tick

J-Melon:
The racket hitbox accepted ball contacts from any direction, so a ball skimming the top or bottom of the paddle could register as a hit. Added a directional gate that rejects hits where the ball is not moving toward the paddle front.

The base Paddle._physics_process call to tick_animation_state() was shadowed by any subclass that overrode _physics_process. Extracted a _physics_move hook so subclasses only provide movement logic, and the animation tick always runs.


tests/unit/paddle/test_paddle.gd
```diff
@@ -42,6 +42,19 @@ func test_pitch_increases_after_cooldown_expires() -> void:
 	assert_almost_eq(_paddle.hit_sound.pitch_scale, 1.10, 0.001)
 
 
+# --- _on_racket_body_entered directional gate ---
+func test_racket_rejects_ball_moving_away_from_paddle() -> void:
+	_paddle.position = Vector2(-300, 0)
+	_paddle._lane_x = -300.0
```
J-Melon: Mabel (gdscript-conventions): `_lane_x` is a private member without `@export`. The test accesses it to configure the paddle side. Consider a public setter in a future refactor. Minor, not blocking.

J-Melon: **Gru**: Leaving _lane_x private. Exporting internal layout state for test access is a larger change than this PR warrants, and the timeout controller already uses the same private _lane_x pattern.

PR #990, opened 2026-06-16
refactor: reconcile BallReconciler from ItemManager state

J-Melon:
Derives BallReconciler ball membership from ItemManager state instead of tracking via signal handlers.

- item_manager_state_changed connected to new _reconcile() method
- _reconcile() removes balls whose level hits zero
- _reconcile() absorbs _reconcile_initial_state logic
- _on_court_changed preserved for real-time court transitions
- Save providers iterate _balls instead of _balls_by_key


scripts/items/ball_reconciler.gd
```diff
@@ -53,7 +53,8 @@ func _ready() -> void:
 		_item_manager = ItemManager
 
 	_item_manager.court_changed.connect(_on_court_changed)
-	_item_manager.item_level_changed.connect(_on_item_level_changed)
+	if _item_manager.has_signal("item_manager_state_changed"):
```
J-Melon: **Convention: raw string signal name** - `has_signal("item_manager_state_changed")` uses a plain string. The project convention prefers `has_signal(&"item_manager_state_changed")` (StringName literal) for type-consistent signal-name references.

**Also: silent degradation** - If `item_manager_state_changed` is absent (the PR this depends on hasn't landed yet), this guard silences the entire reconciliation path. No ball cleanup, initial-state spawn, or stored-kit population runs -- with no warning emitted. Either add `push_warning(...)` when the signal is absent, or remove the guard entirely if #989 must always land first.

scripts/items/ball_reconciler.gd
```diff
@@ -335,8 +316,16 @@ func _on_court_changed(item_key: String, on_court: bool) -> void:
 		ball.global_position = ball_rack.get_slot_position_for(item_key)
 
 
-## One-shot: skipped once any signal-driven court_changed activity has begun.
-func _reconcile_initial_state() -> void:
+func _reconcile() -> void:
+	# Remove balls whose item level has dropped to zero.
+	for key: String in _balls_by_key:
```
J-Melon, line 320: **Dictionary mutation during iteration** - `for key: String in _balls_by_key` iterates the dictionary while `_balls_by_key.erase(key)` is called at L325. Additionally, `get_ball_for_key(key)` at L323 also conditionally calls `_balls_by_key.erase(key)` (if the stored Ball instance is invalid). Erasing from a dictionary while iterating uses a live hash-table walker, not a snapshot -- concurrent modification can skip entries or produce undefined behaviour. Prefer collecting keys to remove in an `Array[String]` first, then removing in a second loop.

scripts/items/item_manager.gd
```diff
@@ -442,10 +445,13 @@ func _set_item_placement(item_key: String, placement: int) -> void:
 		_assign_rack_slot(item_key, item.role)
 	else:
 		state.item_placements[item_key] = placement
+		state.loose_in_venue.erase(item_key)
 		_effect_manager.unregister_source(item)
 		_effect_manager.register_source(item, get_level(item_key))
 		state.rack_slot_index_by_key.erase(item_key)
 
+	item_manager_state_changed.emit()
```
J-Melon, line 453: **Signal fires before no-op guard** - `item_manager_state_changed.emit()` at L453 fires unconditionally, but L455-456 returns early when `previous == placement` and no `loose_in_venue` override is active. This means the signal fires on every `_set_item_placement` call, including no-op calls where the placement hasn't actually changed. Move the emit after the guard check so consumers (like `_reconcile()`) only fire on actual state mutations, avoiding unnecessary reconciliation sweeps.

scripts/items/ball_reconciler.gd
J-Melon: **Gru** (3a8bc96): Fixed — changed to has_signal(&"item_manager_state_changed"). The guard is kept because it\'s already unconditional now that SH-507 (#989) is merged.

scripts/items/ball_reconciler.gd
J-Melon: **Gru** (3a8bc96): Fixed — collects keys in Array[String] first, then removes in a second loop. Also removed the get_ball_for_key call inside the first loop to avoid any stale-instance erase.

scripts/items/item_manager.gd
J-Melon: **Gru** (3a8bc96): By design — the signal fires before the guard because the bookkeeping (lines 438-448) always runs per the spike doc Divergence B. Slot assignment and effect registration happen even on no-op placements; consumers need the signal to catch those changes. _reconcile() is cheap (level check + array walk) so extra calls are negligible.

J-Melon: **Gru**: By design per spike doc Divergence B. The bookkeeping at lines 438-448 always runs — slot assignment and effect registration happen even on no-op placements. _reconcile() is an O(1) level check per key, so extra calls are negligible. The signal fires before the guard to catch silent slot mutations.

PR #994, opened 2026-06-17
feat: Sam sprite art and per-state collision via AnimationPlayer

J-Melon:
Sam now shows sketched key sprites for all six animation states with white backfill. Per-state collision shapes are driven by an AnimationPlayer, tuned in the editor per state. The grounded check uses live collision position and includes hysteresis to prevent boundary flux.

Also: scoped LFS to sprites only, excluded addons, extracted CI test workflow as reusable, and moved LFS pull to deploy jobs only.


resources/animations/sam.tres
```diff
@@ -1,98 +1,149 @@
...
+}, {
+"duration": 1.0,
+"texture": ExtResource("6_et4af")
+}, {
+"duration": 1.0,
+"texture": ExtResource("7_yvpr5")
 }],
-"loop": false,
+"loop": true,
 "name": &"flying_down",
```
J-Melon, line 49: **review** suggestion: flying_down uses flying_up frames (ExtResource 1-7), showing upward-motion art during downward movement. Use a neutral placeholder or reverse frame order until dedicated art lands.

resources/animations/sam.tres
```diff
@@ -1,98 +1,149 @@
...
+"texture": ExtResource("18_2jnnh")
+}, {
+"duration": 1.0,
+"texture": ExtResource("19_i1tn7")
+}, {
+"duration": 1.0,
+"texture": ExtResource("20_7d1k5")
 }],
 "loop": false,
 "name": &"swing_grounded",
```
J-Melon, line 149: **review** question: swing_grounded uses swing_flying PNGs; PR description says gradient placeholder. Intentional? Reasonable if true — swing motion is same airborne or grounded.

scenes/player_paddle.tscn
```diff
@@ -6,28 +6,218 @@
...
+tracks/1/imported = false
+tracks/1/enabled = true
+tracks/1/path = NodePath("BodyCollision:shape:size")
+tracks/1/interp = 1
+tracks/1/loop_wrap = true
+tracks/1/keys = {
+"times": PackedFloat32Array(0),
+"transitions": PackedFloat32Array(1),
+"update": 0,
+"values": [Vector2(90, 201.715)]
```
J-Melon, line 56: **review** question: flying-state BodyCollision:shape:size.y = 201.715; default and ready_grounded use 208. The ~6px discrepancy looks like an editor artifact — confirm intentional.

scripts/entities/paddle.gd
```diff
@@ -238,6 +245,10 @@ func _on_animation_state_changed(state: StringName) -> void:
 	):
 		sprite.play(state)
 
+	if animation_player != null and animation_player.has_animation(state):
+		animation_player.play(state)
```
J-Melon, line 249: **signals-lifecycle** question: `play()` starts at pos 0; is `seek(0.0, true)` for immediate apply needed, or does the next tick suffice?

resources/animations/sam.tres
J-Melon: **Gru**: Temporary - reused flying_up frames until dedicated flying_down art lands.

resources/animations/sam.tres
J-Melon: **Gru**: Intentional temp - reused swing_flying frames until swing_grounded art lands. PR body is stale.

scenes/player_paddle.tscn
J-Melon: **Gru**: Josh tuned these values in the editor per animation state. The 201.715 is the flying state collision height, intentional.

scripts/entities/paddle.gd
J-Melon: **Gru**: Yes - seek(0.0, true) applies the keyframe immediately rather than waiting for the next idle frame. Without it, _is_grounded() reads stale collision values for one frame, causing grounded/flying flux.

PR #997, opened 2026-06-17
refactor: reusable test workflow, LFS to deploy jobs only

J-Melon:
Tests workflow is now callable via workflow_call. Preview and Live Release pipelines invoke it instead of duplicating test steps. LFS pull moved from test jobs to deploy jobs, where assets are actually needed for export. Test jobs stay LFS-free since no unit tests load visual assets.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,58 +23,7 @@ env:
 
 jobs:
   test:
-    name: Run Tests
-    runs-on: ubuntu-latest
-    timeout-minutes: 30
-    permissions:
-      contents: read
-    steps:
-      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
-
-      - name: Cache Godot
-        id: cache-godot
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -23,58 +23,7 @@ env:
...
-          git lfs pull --include="assets/**"
-
-      - name: Import Project
-        run: godot --headless --path . --import
-
-      - name: Run Tests
-        run: |
-          mkdir -p logs
-          godot --headless --path . -s addons/gut/gut_cmdln.gd -gconfig=.gutconfig.json -gexit
+    uses: ./.github/workflows/test.yml
```
J-Melon, line 26: **ci-and-workflows** question: old inline test had `timeout-minutes: 30`; reusable test.yml sets `15`. If release exports ever trigger heavier test suites, 15m may be tight. Intentional?

J-Melon: **Gru**: Intentional. GUT runs in under 2s. If tests ever approach 15m, it signals a test hang, not a timeout budget issue.

PR #998, opened 2026-06-17
fix: restore LFS pull to publish/release test jobs

J-Melon:
Publish and release workflows pass lfs-include to the reusable test workflow. PR tests stay LFS-free. Deploy jobs retain their own LFS step.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,12 +6,6 @@ on:
       - main
   merge_group:
   workflow_call:
-    inputs:
-      lfs-include:
-        description: LFS fetch include pattern (empty = skip)
-        type: string
-        default: ""
-        required: false
 
 concurrency:
   group: tests-${{ github.ref }}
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -50,9 +50,6 @@ jobs:
           path: .godot
           key: godot-import-${{ env.GODOT_VERSION }}-${{ hashFiles('project.godot', 'addons/**', '**/*.import') }}
 
-      - name: Import Project
```
J-Melon, line 53: **ci-and-workflows** issue: removing `--import` breaks GUT on cache miss. Step generates `uid_cache.bin` and `global_script_class_cache.cfg`; missing them causes parse errors ("Could not find type") and autoload failure. Verified on Godot 4.6.3 with empty `.godot/`. Restore `--import`.

.github/workflows/test.yml
```diff
@@ -50,9 +50,6 @@ jobs:
...
-        run: godot --headless --path . --import
-
       - name: Run Tests
```
J-Melon, line 56: **ci-workflows** issue: removing `--import` regresses `test_press_radius_is_inflated_versus_physics_radius`. Without import, `sprite.texture` is null (LFS pointer); `_authored_radius_from_ball` returns 0.0. Restore `--import` or handle null texture in the test.

.github/workflows/test.yml
J-Melon: **Gru**: Fixed by 6b80f55 -- restored --import, leak gate now filters Failed loading resource errors for LFS-tracked asset formats.

.github/workflows/test.yml
J-Melon: **Gru**: Same root cause as above -- --import restored in 6b80f55. Test passes now that import populates the texture cache.

PR #1014, opened 2026-06-22
chore: remove LFS push from publish.yml

J-Melon:
Asset-gate workflow now handles LFS push to GitHub on PR branches. publish.yml only runs on main pushes where objects are already in GitHub LFS from the PR workflow.

### DIFF
--- a/.github/workflows/asset-gate.yml
+++ b/.github/workflows/asset-gate.yml
@@ -48,4 +48,4 @@ jobs:
         run: |
           git config lfs.url "https://volley:${{ secrets.LFS_DOWNLOAD_KEY }}@volley-lfs-proxy.volcanoem.workers.dev"
           git lfs pull --include="assets/**"
-          git -c lfs.url=https://github.com/shuck-dev/volley.git/info/lfs lfs push --object-id origin "$(git lfs ls-files -l | awk '{print $1}')"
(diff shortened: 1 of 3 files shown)


.github/workflows/asset-gate.yml
```diff
@@ -35,3 +33,19 @@ jobs:
...
+      contents: write
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+      - name: Push LFS objects to GitHub LFS
+        continue-on-error: true
+        run: |
+          git config lfs.url "https://volley:${{ secrets.LFS_DOWNLOAD_KEY }}@volley-lfs-proxy.volcanoem.workers.dev"
+          git lfs pull --include="assets/**"
+          git -c lfs.url=https://github.com/shuck-dev/volley.git/info/lfs lfs push --object-id origin "$(git lfs ls-files -l | awk '{print $1}')"
```
J-Melon: **Skua** suggestion: `"$(...)"` prevents word-splitting; multiple LFS OIDs get passed as one newline-separated argument instead of separate args. Drop the outer quotes or pipe through `xargs`. Carried over from publish.yml unchanged.

J-Melon: Fixed in c0464c6: removed outer quotes so each OID is a separate argument to --object-id.

J-Melon: Updated to xargs in 9f8074f: avoids both quoting issues and shellcheck SC2046.

PR #1022, opened 2026-06-22
fix: serve preview images from R2 instead of base64

J-Melon:
Base64 data URIs are stripped from GitHub comments. Uploads WebP previews to R2 public-read bucket and references via URL instead.

### DIFF
--- a/.github/workflows/pr-asset-preview.yml
+++ b/.github/workflows/pr-asset-preview.yml
@@ -66,6 +66,9 @@ jobs:
             key="pr-previews/${{ github.event.pull_request.number }}/${name}.webp"
             url="https://volley-lfs-proxy.volcanoem.workers.dev/${key}"
 
-            gh pr comment ${{ github.event.pull_request.number }} \
-              --body "$(printf '**`%s`**\n![%s](%s)' "$file" "$name" "$url")"
+            body="$(
+              printf '**`%s`**\n' "$file"
+              printf '![%s](%s)' "$name" "$url"
+            )"
+            gh pr comment ${{ github.event.pull_request.number }} --body "$body"
           done < /tmp/changed.txt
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-asset-preview.yml
```diff
@@ -40,18 +40,32 @@ jobs:
...
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          while read -r file; do
+            name=$(basename "$file" .png)
+            key="pr-previews/${{ github.event.pull_request.number }}/${name}.webp"
+            url="https://volley-lfs-proxy.volcanoem.workers.dev/${key}"
 
-            gh pr comment ${{ github.event.pull_request.number }} --body "$body"
+            gh pr comment ${{ github.event.pull_request.number }} \
+              --body "**\`${file}\`**\n![${name}](${url})"
```
J-Melon: **Ford** issue: `\n` in double quotes is literal, not a newline. Comment body renders `\n` as text — image on same line as filename. Fix with `$
` or `printf` for an actual newline.

.github/workflows/pr-asset-preview.yml
```diff
@@ -40,18 +40,32 @@ jobs:
...
         run: |
           while read -r file; do
             name=$(basename "$file" .png)
             webp="/tmp/${name}.webp"
             cwebp -q 80 -m 6 -mt "$file" -o "$webp" -quiet
 
-            data=$(base64 -w0 "$webp")
-            body="**\`${file}\`**\n![](data:image/webp;base64,${data})"
+            key="pr-previews/${{ github.event.pull_request.number }}/${name}.webp"
+            curl -sf -X PUT \
```
J-Melon: **Ford** suggestion: `-sf` hides curl error output — upload failures are opaque. Use `-sSf` to surface errors. Auth header is in request headers, not curl stderr — no secret leak.

.github/workflows/pr-asset-preview.yml
J-Melon: Fixed in 0d3fcaa.

.github/workflows/pr-asset-preview.yml
J-Melon: Fixed in 0d3fcaa.

PR #1028, opened 2026-06-22
fix: pin preview comments to file level

J-Melon:
Uses subject_type=file instead of position=1 so comments anchor to the file rather than a specific diff line.

### DIFF
--- a/.github/workflows/pr-asset-preview.yml
+++ b/.github/workflows/pr-asset-preview.yml
@@ -74,5 +74,5 @@ jobs:
               -F commit_id="$commit_id" \
               -F path="$file" \
               -F body="$body" \
-              -F position=1
+              -F subject_type=file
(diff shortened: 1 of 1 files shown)


J-Melon (commented):
## Review: correct

`subject_type=file` is valid per API docs. Proper fix.

No changes requested.

.github/workflows/pr-asset-preview.yml
```diff
@@ -74,5 +74,5 @@ jobs:
               -F commit_id="$commit_id" \
               -F path="$file" \
               -F body="$body" \
-              -F position=1
+              -F subject_type=file
```
J-Melon, line 77: Correct: `subject_type=file` is valid per API docs. Removes fragile `position=1` hack.

PR #1029, opened 2026-06-22
docs: per-state racket position via Marker2D

J-Melon:
Adds the spike decision doc: Marker2D nodes under a RacketPositions container in the paddle scene, one per animation state. Resolves SH-518. An implementation PR follows as a child ticket.

### DIFF
--- a/designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
+++ b/designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
@@ -8,72 +8,43 @@ reach, the collider question reopens for that state alone."
 
 ## Decision
 
-Marker2D child nodes under a `RacketPositions` container in the paddle scene,
-one per animation state that needs a distinct racket position. On state change,
-copy the matching marker's local position to `racket_hitbox.position`. States
-without a marker fall back to the scene-authored RacketHitbox position.
+A single `LowRacketAnchor` Marker2D child defines the low-swing racket position.
+The scene-authored RacketHitbox position is the default for every other state.
+On state change, the paddle switches between the two: the low anchor when the
+state is crouch or low-swing, the default otherwise.
 
 The per-frame Y-override in `player_paddle.gd._physics_move` is removed
-entirely. The racket position is set once on state transition, not recomputed
-every physics tick.
+entirely. The racket position is set on state transition, not recomputed every
+physics tick.
 
 ## Why Marker2D
 
-**Visual.** The designer selects a marker in the scene tree, sees a cross-gizmo
-in the 2D viewport, and drags it to where the racket should sit for that state.
-No number guessing, no animation scrubbing, no resource-file round trips.
+**Visual.** The designer selects the anchor in the scene tree, sees a cross-gizmo
+in the 2D viewport, and drags it to where the racket should sit during a low
+swing. No number guessing.
 
 **Precedent.** The paddle scene already uses Node2D anchors as positional
 references: `AnkleAnchor` at `[0, 24]`, `GripAnchor` at `[-12, 0]`,
-`WristAnchor` at `[-12, -10]`. The `RacketPositions` container extends the same
-pattern.
-
-**Hook fits naturally.** `_on_animation_state_changed` already fires on every
-state transition via the paddle animation state machine. Adding a racket-position
-lookup there is a two-line addition with no structural refactoring.
+`WristAnchor` at `[-12, -10]`. `LowRacketAnchor` extends the same pattern.
 
 **Dev panel survives.** The existing `set_racket_position_x` and
 `set_racket_position_y` on `paddle.gd` store an additive offset applied on top
-of the per-state marker position. Each state transition recomputes
-`racket_hitbox.position = marker.position + offset`, preserving the offset
-across states. The spinboxes that appeared broken for Y (the `_physics_move`
-override was writing over them every frame) now work as live fine-tune controls.
-
-## Why not the alternatives
-
-| Approach | Rejected because |
-|---|---|
-| `@export` dictionary per state | Not visually tunable. The designer types coordinates blind in the inspector, no viewport feedback. |
-| AnimationPlayer on RacketHitbox position | Duplicates the SpriteFrames animation infrastructure. Seven animation states require two systems kept in sync. A new state means updating both. Adds a timeline for what is fundamentally a single-position-per-state problem. |
-| Custom Resource with Vector2 per state | Another file type to manage. Still no viewport visual. Over-engineered for seven positions. |
+of the anchor position. Each state transition recomputes
+`racket_hitbox.position = anchor_position + offset`, preserving the offset
+across states.
 
 ## Scene shape
 
 ```
 PlayerPaddle (CharacterBody2D)
 ├── Sprite (AnimatedSprite2D)
-├── RacketPositions (Node2D)
-│   ├── RPos_ready_grounded (RacketPositionMarker)
-│   ├── RPos_ready_flying (RacketPositionMarker)
-│   ├── RPos_flying_up (RacketPositionMarker)
-│   ├── RPos_flying_down (RacketPositionMarker)
-│   ├── RPos_swing_grounded (RacketPositionMarker)
-│   ├── RPos_low_swing_grounded (RacketPositionMarker)
-│   └── RPos_swing_flying (RacketPositionMarker)
+├── LowRacketAnchor (RacketPositionMarker)
 ├── RacketHitbox (Area2D)
 │   └── RacketCollision (CollisionShape2D)
 ├── ...existing nodes...
 ```
 
-All seven markers exist in the scene. States that share the default position
-can omit their marker; the code falls back to the scene-authored RacketHitbox
-position. The container keeps the tree tidy.
-
 ## Editor tool
-
-The Marker2D crosshair alone does not show the RacketHitbox extents. An editor
-tool renders the collision shape so the designer sees the actual hitzone while
-dragging.
 
 A custom `RacketPositionMarker` class extends `Marker2D` with `@tool`:
 
@@ -88,22 +59,27 @@ func _draw() -> void:
 	draw_rect(Rect2(-collision_size * 0.5, collision_size), Color.RED, false, 2.0)
 ```
 
-- `collision_size` is set to match the RacketCollision RectangleShape2D.
-- `_draw()` fires in the editor; the rectangle follows the marker around as the
-  designer drags it.
-- The colour and line weight are editor-only visualisation; they do not appear
-  at runtime.
-- The markers in the scene shape above use `RacketPositionMarker` instead of
-  raw `Marker2D`.
+- `collision_size` matches the RacketCollision RectangleShape2D.
+- `_draw()` fires in the editor; the rectangle follows the marker.
+- Colour and line weight are editor-only; they do not appear at runtime.
 
 ## Code shape
 
-`player_paddle.gd` overrides a new virtual `_apply_racket_position(state)` on
-the base class. The override collects Marker2D children from `RacketPositions`
-at `_ready`, maps them by state name, and on each call copies the matching
-marker's `position` to `racket_hitbox.position`.
+`player_paddle.gd` overrides a virtual `_apply_racket_position(state)` on
+`paddle.gd`. The override switches the racket position between the low anchor
+and the default based on the state's crouch/lowness:
 
-The base class `paddle.gd` calls `_apply_racket_position(state)` from
+```
+func _apply_racket_position(state: StringName) -> void:
+	if state in _low_states:
+		racket_hitbox.position = low_anchor.position
+	else:
+		racket_hitbox.position = _default_racket_pos
+	racket_hitbox.position += _dev_offset
+	_refresh_overlay_shapes()
+```
+
+The base class calls `_apply_racket_position(state)` from
 `_on_animation_state_changed` after the sprite animation plays. The default
 implementation is a no-op; partner paddles inherit it unchanged.
 
--- a/designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
+++ b/designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
@@ -1,0 +1,89 @@
+# Racket hitbox position per animation state
+
+Resolves SH-518 (visual tuning of the racket collision zone during Sam's
+low-swing pose). Continues `2026-06-06-paddle-animation-collision-spike.md`,
+which decided the CharacterBody2D body collider stays fixed across states.
+That decision stands. This doc addresses a separate shape: the RacketHitbox
+Area2D, the mid-body zone that detects the ball and feeds the return-angle
+calculus. That zone shifts with the crouch pose and the prior decision did not
+cover it.
+
+## Decision
+
+Marker2D child nodes under a `RacketPositions` container in the paddle scene,
+one per animation state that needs a distinct racket position. On state change,
+copy the matching marker's local position to `racket_hitbox.position`. States
+without a marker fall back to the scene-authored RacketHitbox position.
+
+The per-frame Y-override in `player_paddle.gd._physics_move` is removed
+entirely. The racket position is set once on state transition, not recomputed
+every physics tick.
+
+## Why Marker2D
+
+**Visual.** The designer selects a marker in the scene tree, sees a cross-gizmo
+in the 2D viewport, and drags it to where the racket should sit for that state.
+No number guessing, no animation scrubbing, no resource-file round trips.
+
+**Precedent.** The paddle scene already uses Node2D anchors as positional
+references: `AnkleAnchor` at `[0, 24]`, `GripAnchor` at `[-12, 0]`,
+`WristAnchor` at `[-12, -10]`. The `RacketPositions` container extends the same
+pattern.
+
+**Hook fits naturally.** `_on_animation_state_changed` already fires on every
+state transition via the paddle animation state machine. Adding a racket-position
+lookup there is a two-line addition with no structural refactoring.
+
+**Dev panel survives.** The existing `set_racket_position_x` and
+`set_racket_position_y` on `paddle.gd` become additive offsets applied on top
+of the per-state position. The spinboxes that appeared broken for Y (the
+`_physics_move` override was writing over them every frame) now work as live
+fine-tune controls.
+
+## Why not the alternatives
+
+| Approach | Rejected because |
+|---|---|
+| `@export` dictionary per state | Not visually tunable. The designer types coordinates blind in the inspector, no viewport feedback. |
+| AnimationPlayer on RacketHitbox position | Duplicates the SpriteFrames animation infrastructure. Seven animation states require two systems kept in sync. A new state means updating both. Adds a timeline for what is fundamentally a single-position-per-state problem. |
+| Custom Resource with Vector2 per state | Another file type to manage. Still no viewport visual. Over-engineered for seven positions. |
+
+## Scene shape
+
+```
+PlayerPaddle (CharacterBody2D)
+├── Sprite (AnimatedSprite2D)
+├── RacketPositions (Node2D)
+│   ├── RPos_ready_grounded (Marker2D)
+│   ├── RPos_ready_flying (Marker2D)
+│   ├── RPos_flying_up (Marker2D)
+│   ├── RPos_flying_down (Marker2D)
+│   ├── RPos_swing_grounded (Marker2D)
+│   ├── RPos_low_swing_grounded (Marker2D)
+│   └── RPos_swing_flying (Marker2D)
+├── RacketHitbox (Area2D)
+│   └── RacketCollision (CollisionShape2D)
+├── ...existing nodes...
+```
+
+All seven markers exist in the scene. States that share the default position
+can omit their marker; the code falls back to the scene-authored RacketHitbox
+position. The container keeps the tree tidy.
+
+## Code shape
+
+`player_paddle.gd` overrides a new virtual `_apply_racket_position(state)` on
+the base class. The override collects Marker2D children from `RacketPositions`
+at `_ready`, maps them by state name, and on each call copies the matching
+marker's `position` to `racket_hitbox.position`.
+
+The base class `paddle.gd` calls `_apply_racket_position(state)` from
+`_on_animation_state_changed` after the sprite animation plays. The default
+implementation is a no-op; partner paddles inherit it unchanged.
+
+`_base_racket_y` and the crouch-position formula in `_physics_move` are
+removed. The racket position is no longer touched in the physics loop.
+
+## Implementation
+
+SH-xxx (child of SH-518).
--- a/designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
+++ b/designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
@@ -1,12 +1,8 @@
 # Racket hitbox position per animation state
 
-Resolves SH-518 (visual tuning of the racket collision zone during Sam's
-low-swing pose). Continues `2026-06-06-paddle-animation-collision-spike.md`,
-which decided the CharacterBody2D body collider stays fixed across states.
-That decision stands. This doc addresses a separate shape: the RacketHitbox
-Area2D, the mid-body zone that detects the ball and feeds the return-angle
-calculus. That zone shifts with the crouch pose and the prior decision did not
-cover it.
+Resolves SH-518. The racket hitbox (RacketHitbox Area2D, the mid-body zone
+that detects the ball) shifts with Sam's crouch pose. The body collider stays
+fixed per `2026-06-06-paddle-animation-collision-spike.md`.
 
 ## Decision
 
--- a/designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
+++ b/designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
@@ -1,8 +1,10 @@
 # Racket hitbox position per animation state
 
-Resolves SH-518. The racket hitbox (RacketHitbox Area2D, the mid-body zone
-that detects the ball) shifts with Sam's crouch pose. The body collider stays
-fixed per `2026-06-06-paddle-animation-collision-spike.md`.
+The racket hitbox (RacketHitbox Area2D, the mid-body zone that detects the
+ball) shifts with Sam's crouch pose. The CharacterBody2D body collider stays
+fixed per `2026-06-06-paddle-animation-collision-spike.md`. The prior spike
+anticipated this in passing: "if a future ready pose ever extends the paddle's
+reach, the collider question reopens for that state alone."
 
 ## Decision
 
@@ -31,10 +33,11 @@ state transition via the paddle animation state machine. Adding a racket-positio
 lookup there is a two-line addition with no structural refactoring.
 
 **Dev panel survives.** The existing `set_racket_position_x` and
-`set_racket_position_y` on `paddle.gd` become additive offsets applied on top
-of the per-state position. The spinboxes that appeared broken for Y (the
-`_physics_move` override was writing over them every frame) now work as live
-fine-tune controls.
+`set_racket_position_y` on `paddle.gd` store an additive offset applied on top
+of the per-state marker position. Each state transition recomputes
+`racket_hitbox.position = marker.position + offset`, preserving the offset
+across states. The spinboxes that appeared broken for Y (the `_physics_move`
+override was writing over them every frame) now work as live fine-tune controls.
 
 ## Why not the alternatives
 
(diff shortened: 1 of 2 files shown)


designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,89 @@
+# Racket hitbox position per animation state
+
+Resolves SH-518 (visual tuning of the racket collision zone during Sam's
+low-swing pose). Continues `2026-06-06-paddle-animation-collision-spike.md`,
+which decided the CharacterBody2D body collider stays fixed across states.
+That decision stands. This doc addresses a separate shape: the RacketHitbox
+Area2D, the mid-body zone that detects the ball and feeds the return-angle
```
J-Melon: **Personification.** "Feeds the return-angle calculus" hands the zone a human verb. Suggestion: "computes the return angle." Per STYLE.md line 61.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,89 @@
...
+
+| Approach | Rejected because |
+|---|---|
+| `@export` dictionary per state | Not visually tunable. The designer types coordinates blind in the inspector, no viewport feedback. |
+| AnimationPlayer on RacketHitbox position | Duplicates the SpriteFrames animation infrastructure. Seven animation states require two systems kept in sync. A new state means updating both. Adds a timeline for what is fundamentally a single-position-per-state problem. |
+| Custom Resource with Vector2 per state | Another file type to manage. Still no viewport visual. Over-engineered for seven positions. |
+
+## Scene shape
+
+```
```
J-Melon: **Ungrounded states.** Flying/grounded split (ready_flying, flying_up, etc.) not in reference spike (paddle-animation-collision.md defines idle/ready/swing/walk). A cold reader doesn't know where these came from. Either define the set here or reference the source.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,89 @@
...
+marker's `position` to `racket_hitbox.position`.
+
+The base class `paddle.gd` calls `_apply_racket_position(state)` from
+`_on_animation_state_changed` after the sprite animation plays. The default
+implementation is a no-op; partner paddles inherit it unchanged.
+
+`_base_racket_y` and the crouch-position formula in `_physics_move` are
+removed. The racket position is no longer touched in the physics loop.
+
+## Implementation
```
J-Melon: **Stub.** SH-xxx is a placeholder. Resolve the number or remove the section. Also missing an "Out of scope" block like the reference spike has (paddle-animation-collision-spike.md#L141).

J-Melon (commented):
## Prose review - SH-518 tech spike doc

**Verdict:** Approved with three observations. The decision is legible from a cold read, prose is clean (no em dashes, no banned vocabulary, no hedging), citations (SH-518, paddle-animation-collision-spike.md) are correct. Structure follows the reference spike pattern with one improvement (alternatives rejection table) and one gap (no out-of-scope section).

**Issues:**
1. Personification / AI register at line 7
2. Ungrounded state names at lines 57-63
3. Placeholder stub at lines 87-89

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,89 @@
+# Racket hitbox position per animation state
+
+Resolves SH-518 (visual tuning of the racket collision zone during Sam's
+low-swing pose). Continues `2026-06-06-paddle-animation-collision-spike.md`,
+which decided the CharacterBody2D body collider stays fixed across states.
```
J-Melon: **Marvin** suggestion: the spike doc (2026-06-06:88-89) anticipated this: "the collider question reopens" for per-state reach changes. Cross-reference that section to connect the two decisions.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,89 @@
...
+The base class `paddle.gd` calls `_apply_racket_position(state)` from
+`_on_animation_state_changed` after the sprite animation plays. The default
+implementation is a no-op; partner paddles inherit it unchanged.
+
+`_base_racket_y` and the crouch-position formula in `_physics_move` are
+removed. The racket position is no longer touched in the physics loop.
+
+## Implementation
+
+SH-xxx (child of SH-518).
```
J-Melon: **Marvin** nitpick: resolve the "SH-xxx" placeholder with the real child ticket number before merging.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,89 @@
+# Racket hitbox position per animation state
```
J-Melon, line 1: **Marvin** suggestion: list this doc in designs/01-prototype/INDEX.md under a paddle section. The 2026-06-06 spike it continues also isnt listed; add both.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,85 @@
...
+Resolves SH-518. The racket hitbox (RacketHitbox Area2D, the mid-body zone
+that detects the ball) shifts with Sam's crouch pose. The body collider stays
+fixed per `2026-06-06-paddle-animation-collision-spike.md`.
+
+## Decision
+
+Marker2D child nodes under a `RacketPositions` container in the paddle scene,
+one per animation state that needs a distinct racket position. On state change,
+copy the matching marker's local position to `racket_hitbox.position`. States
+without a marker fall back to the scene-authored RacketHitbox position.
```
J-Melon: **Adversary** question: when RacketPositions is missing or no marker name matches, fallback is silent. Consider a debug warning so an unpositioned state is discoverable.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,85 @@
...
+
+**Hook fits naturally.** `_on_animation_state_changed` already fires on every
+state transition via the paddle animation state machine. Adding a racket-position
+lookup there is a two-line addition with no structural refactoring.
+
+**Dev panel survives.** The existing `set_racket_position_x` and
+`set_racket_position_y` on `paddle.gd` become additive offsets applied on top
+of the per-state position. The spinboxes that appeared broken for Y (the
+`_physics_move` override was writing over them every frame) now work as live
+fine-tune controls.
```
J-Melon: **Adversary** issue: no specification of how the dev offset composes with the marker position on state transitions. Does the code set `marker_pos + offset` or `marker_pos` (losing the offset)? What resets the offset when a marker is dragged? Without this, the dev panel and marker system fight each other.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Why not the alternatives
+
+| Approach | Rejected because |
+|---|---|
+| `@export` dictionary per state | Not visually tunable. The designer types coordinates blind in the inspector, no viewport feedback. |
```
J-Melon: **Adversary** suggestion: the @export-dict dismissal is too quick. It is inspectable at runtime (live Godot property editing, viewport-visible). The real tradeoff is gizmo speed vs. field readability, not visual vs. blind.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,85 @@
...
+| AnimationPlayer on RacketHitbox position | Duplicates the SpriteFrames animation infrastructure. Seven animation states require two systems kept in sync. A new state means updating both. Adds a timeline for what is fundamentally a single-position-per-state problem. |
```
J-Melon: **Adversary** suggestion: AnimationPlayer would guarantee sync: every new animation state gets a position keyframe automatically. Marker2D requires remembering to add a node; missing it is silent fallback, not caught at compile time.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,85 @@
...
+│   ├── RPos_flying_down (Marker2D)
+│   ├── RPos_swing_grounded (Marker2D)
+│   ├── RPos_low_swing_grounded (Marker2D)
+│   └── RPos_swing_flying (Marker2D)
+├── RacketHitbox (Area2D)
+│   └── RacketCollision (CollisionShape2D)
+├── ...existing nodes...
+```
+
+All seven markers exist in the scene. States that share the default position
```
J-Melon: **Adversary** question: table shows all seven markers. The text says shared-default states can omit. Which are diverging? If unknown yet, mark some as optional so the table matches the stated design.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,85 @@
...
+`player_paddle.gd` overrides a new virtual `_apply_racket_position(state)` on
+the base class. The override collects Marker2D children from `RacketPositions`
+at `_ready`, maps them by state name, and on each call copies the matching
+marker's `position` to `racket_hitbox.position`.
+
+The base class `paddle.gd` calls `_apply_racket_position(state)` from
+`_on_animation_state_changed` after the sprite animation plays. The default
+implementation is a no-op; partner paddles inherit it unchanged.
+
+`_base_racket_y` and the crouch-position formula in `_physics_move` are
```
J-Melon: **Adversary** suggestion: removing the crouch-only formula changes behaviour: crouch-without-swing no longer shifts the racket. Clarify whether this is intentional or needs a crouch-ready state.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+All seven markers exist in the scene. States that share the default position
+can omit their marker; the code falls back to the scene-authored RacketHitbox
+position. The container keeps the tree tidy.
+
+## Code shape
+
+`player_paddle.gd` overrides a new virtual `_apply_racket_position(state)` on
+the base class. The override collects Marker2D children from `RacketPositions`
```
J-Melon: **Adversary** suggestion: collecting children at _ready means the map is stale if a node is added or renamed post-init by a tool script. Dynamic lookup per call (cheap) guards against editor-driven mutations.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,85 @@
...
+**Hook fits naturally.** `_on_animation_state_changed` already fires on every
+state transition via the paddle animation state machine. Adding a racket-position
+lookup there is a two-line addition with no structural refactoring.
+
+**Dev panel survives.** The existing `set_racket_position_x` and
+`set_racket_position_y` on `paddle.gd` become additive offsets applied on top
+of the per-state position. The spinboxes that appeared broken for Y (the
+`_physics_move` override was writing over them every frame) now work as live
+fine-tune controls.
+
```
J-Melon: **Marvin** question: how do `set_racket_position_x/y` become additive? Current code sets absolute position. Requires separate offset storage re-applied after state transitions.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: **The prior decision may bind more tightly than this doc admits.**

The 2026-06-06 doc concluded: "The collider does not change across animation states" and "the ball always strikes the same body." The RacketHitbox **is** where the ball is judged to strike - `paddle.gd._on_racket_body_entered` routes the ball to `ball.hit_by_paddle(self)`, and `get_half_height()` feeds the return-angle denominator from the racket shape. Changing the racket position per state **does** change what the ball strikes.

The doc distinguishes RacketHitbox from the CharacterBody2D body collider, which is technically correct. But the prior decision's principle was stability of the contact zone. This doc needs to argue why that principle does not apply to the RacketHitbox, rather than just stating it is "a separate shape."

J-Melon (commented):
## Adversarial review: Marker2D per-state racket position

This design doc makes a clean argument, but several load-bearing assumptions need a harder look before implementation. I will flag the contradictions, the silent-behavior changes, and the alternatives the doc did not consider.

### Overarching challenge

The plan converts a simple per-frame Y override into a scene-tree-driven positional lookup with 7-8 new nodes, solving a snapshot problem ("where should the racket sit for this state?") with a runtime-invasive structural change. The crouch case reveals a fundamental tension: the animation state machine today has **no non-swing crouch state** - `_resolve_state` returns `ready_grounded` regardless of crouching when there is no swing pending. The per-frame override currently compensates for this gap; the Marker2D approach does not.

### Summary of findings

Inline comments detail each point below. Major themes:

1. **Crouch is a live input, not a state transition** - removing the per-frame override breaks responsive racket positioning for crouch-without-swing, because the FSM does not change state for it.
2. **The "visually tunable" promise is overstated** - a Marker2D crosshair shows where the marker sits, not where the RacketHitbox's RectangleShape2D covers relative to the body collider. The loop is still edit -> run -> observe -> edit.
3. **The fallback path is silent dead code on day one** - all 7 states have markers. A renamed state or a deleted marker fails silently.
4. **8 new nodes for a 2-position problem** is heavy speculative engineering. The scene grows 50%.
5. **The dev panel contract changes silently**: `set_racket_position_y` goes from absolute to additive, which changes its meaning across state transitions.
6. **A simpler alternative is missing** - a state-to-Y override on the state machine itself avoids the scene clutter and keeps the responsiveness model intact.

I would not block the PR - this is a design doc, not code - but the implementation should resolve the crouch-idle gap and add a warning path for missing markers before merging.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,89 @@
...
+Area2D, the mid-body zone that detects the ball and feeds the return-angle
+calculus. That zone shifts with the crouch pose and the prior decision did not
+cover it.
+
+## Decision
+
+Marker2D child nodes under a `RacketPositions` container in the paddle scene,
+one per animation state that needs a distinct racket position. On state change,
+copy the matching marker's local position to `racket_hitbox.position`. States
+without a marker fall back to the scene-authored RacketHitbox position.
```
J-Melon: **The fallback states the doc expects to exercise do not exist in the FSM.**

The state machine currently returns these 7 states: `ready_grounded`, `ready_flying`, `flying_up`, `flying_down`, `swing_grounded`, `low_swing_grounded`, `swing_flying`. All 7 have markers in the scene shape. The fallback is dead code on day one.

More critically: **the FSM has no non-swing crouch state.** When the player is grounded, holds `paddle_down`, and is **not** swinging, the machine returns `ready_grounded` (line 50-51 of paddle_animation_state_machine.gd: `if grounded: return &"ready_grounded"`). The `crouching` parameter is only consulted during `swing_pending` evaluation. This means crouching-without-swinging produces the same state as standing-without-swinging.

The current per-frame override in `_physics_move` compensates for this: it lowers the racket whenever `is_grounded() and Input.is_action_pressed("paddle_down")`, regardless of swing state. Removing that override means **the racket no longer responds to idle crouch** unless the FSM gains a crouch-idle state first.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,89 @@
...
+
+The per-frame Y-override in `player_paddle.gd._physics_move` is removed
+entirely. The racket position is set once on state transition, not recomputed
```
J-Melon: **"Set once on state transition" assumes the responsiveness model does not matter for crouch.**

Crouch is a held input (`Input.is_action_pressed("paddle_down")`), checked every physics tick. The current code responds immediately to press and release within the same state. After this change, the racket position updates only when `_on_animation_state_changed` fires - which means only when the FSM resolves a **different** state than the current one.

Consider this sequence:
1. Player is grounded, idle -> FSM: `ready_grounded` -> racket at ready marker position
2. Player presses crouch, holds it -> FSM: still `ready_grounded` (no state change) -> **racket stays at standing position**
3. Player releases crouch -> FSM: still `ready_grounded` -> **racket stays at standing position** (no change needed, but the point is responsiveness is gone)
4. Player crouch + hits ball -> FSM: `low_swing_grounded` -> racket finally moves (correct for the swing, but the pre-swing visual of a crouching paddle never shows)

The doc must either: (a) add a non-swing crouch state to the FSM so the position change triggers on the input, or (b) argue that idle-crouch visual positioning does not matter.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,89 @@
...
+every physics tick.
+
+## Why Marker2D
+
+**Visual.** The designer selects a marker in the scene tree, sees a cross-gizmo
```
J-Melon: **"Visually tunable" - the Marker2D crosshair is not the RacketHitbox.**

The designer drags a Marker2D gizmo - a cross - to a point in the 2D viewport. That point is the marker's own local position. It does not show the RacketHitbox's extents (the RectangleShape2D), its offset from the body collider, or its relationship to the animated sprite. The RacketHitbox is an invisible Area2D; its collision shape is only visible when the node is selected.

To verify the position is correct, the designer must still: save, run the game, trigger the state, observe ball behavior, return to edit. That is the same loop as typing a Y value into an inspector field. The Marker2D skips one "guess and save" cycle at best - you see where the racket center would be, but not whether the collision zone actually covers the ball's arc.

If the real goal is viewport feedback for the **hitbox**, the tooling gap is that Godot does not show Area2D collision shapes at edit time by default. That is not a Marker2D problem; it is a workflow problem that a crosshair does not solve.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,89 @@
...
+
+**Hook fits naturally.** `_on_animation_state_changed` already fires on every
+state transition via the paddle animation state machine. Adding a racket-position
+lookup there is a two-line addition with no structural refactoring.
+
+**Dev panel survives.** The existing `set_racket_position_x` and
+`set_racket_position_y` on `paddle.gd` become additive offsets applied on top
+of the per-state position. The spinboxes that appeared broken for Y (the
+`_physics_move` override was writing over them every frame) now work as live
+fine-tune controls.
```
J-Melon: **The dev panel contract changes silently.**

Currently `set_racket_position_y(offset_y)` sets `racket_hitbox.position.y = offset_y` - absolute positioning. The design doc says it becomes an "additive offset applied on top of the per-state position." This changes the semantics of the dev panel spinbox:

- The default value in `player_sprite.gd` is `-10.0` (line 17). Currently this means "position.y = -10". After the change, it means "-10 offset from the per-state marker position."
- If the designer tunes Y while the paddle is in state A (`ready_grounded` marker at y=-39), then exits the dev panel, and the paddle transitions to state B (`low_swing_grounded` marker at a different Y), the same offset value now positions the racket relative to B's marker, not A's. The paddle's racket visibly jumps unless the designer understands the stacking model.
- If `_for_each_paddle` applies `set_racket_position_y` to partner paddles (which skip the marker lookup via the no-op override), what does the offset mean for partners? Absolute or additive? The doc does not say.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Why not the alternatives
+
+| Approach | Rejected because |
+|---|---|
+| `@export` dictionary per state | Not visually tunable. The designer types coordinates blind in the inspector, no viewport feedback. |
```
J-Melon: **Missing alternative: a state-to-Y override on the state machine itself.**

The alternatives table rejects `@export` dictionaries as "not visually tunable" and custom resources as "over-engineered." But a simpler option is not considered:

```gdscript
# On PlayerPaddle, a per-state Y offset:
func _racket_y_for_state(state: StringName) -> float:
    match state:
        &"low_swing_grounded": return _compute_crouch_y()
        _: return _base_racket_y
```

Or even just a `Dictionary[StringName, float]` on the state machine, initialized in `_ready`. This keeps:
- **Live input responsiveness** (called every physics tick, not just on state transition)
- **No scene clutter** (zero new nodes)
- **No silent fallback failure** (an unknown state hits the `_` match arm, which is explicit)
- **Calculator-tunable** (the designer types a number, runs, observes - the same loop as Marker2D, minus one "drag a crosshair" step)

The "no viewport feedback" criticism of the dictionary applies equally to Marker2D: neither shows the RacketHitbox extents. The crosshair is prettier, but not functionally more accurate.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,89 @@
...
+│   ├── RPos_flying_down (Marker2D)
+│   ├── RPos_swing_grounded (Marker2D)
+│   ├── RPos_low_swing_grounded (Marker2D)
+│   └── RPos_swing_flying (Marker2D)
+├── RacketHitbox (Area2D)
+│   └── RacketCollision (CollisionShape2D)
+├── ...existing nodes...
+```
+
+All seven markers exist in the scene. States that share the default position
```
J-Melon: **8 new nodes (7 markers + container) on a scene that is currently 16 nodes = 50% growth.**

The doc presents the markers as future-proofing, but 7 markers for what is currently a **2-position problem** (normal vs crouch) is a significant scene-bloat cost. Every future state change requires:
1. Editing the state machine code
2. Adding/removing/renaming a Marker2D in the scene
3. Remembering both sides

By contrast, a dictionary or code-level override is one location to edit for both the state definition AND the position.

Also: `RPos_` markers are Marker2D nodes. Marker2D extends Node2D. Each is a full transform node at runtime with position/rotation/scale overhead, even though only `position` is read. If the approach is adopted, consider using plain `Node2D` (lighter, same position API) instead of `Marker2D` (adds editor-only gizmo drawing you do not need on all 7).

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
```diff
@@ -0,0 +1,89 @@
...
+can omit their marker; the code falls back to the scene-authored RacketHitbox
+position. The container keeps the tree tidy.
+
+## Code shape
+
+`player_paddle.gd` overrides a new virtual `_apply_racket_position(state)` on
+the base class. The override collects Marker2D children from `RacketPositions`
+at `_ready`, maps them by state name, and on each call copies the matching
+marker's `position` to `racket_hitbox.position`.
```
J-Melon: **Name-based marker lookup is fragile and silent on failure.**

The code collects Marker2D children from `RacketPositions` at `_ready` and maps them by state name (presumably by stripping an `RPos_` prefix from the node name, or by parsing a naming convention). This is stringly-typed coupling between the scene and the code:

- Rename an animation state in the FSM (e.g., `low_swing_grounded` to `crouch_swing`) -> marker `RPos_low_swing_grounded` no longer matches -> fallback to scene-authored RacketHitbox position -> **the ball may miss the paddle during low swings**, because the racket is at the default position instead of the crouch offset
- Delete a marker accidentally -> same silent fallback
- No compile-time check, no `push_warning`, no runtime assert

The implementation needs at minimum a `push_warning` or `assert` on `_ready` if a state the paddle can enter has no matching marker. The design doc should specify this.

J-Melon: @-

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Resolved: rewritten in the opening paragraph.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Noted for SH-519: FSM needs a crouch-idle state.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Noted for SH-519: held input needs FSM state change to trigger racket move.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Resolved: RacketPositionMarker @tool draws collision rect at marker position.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Resolved: marker.pos + offset recompute on each state transition.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Noted for SH-519: semantics shift is implementation detail.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Noted for SH-519: valid alternative. Editor tool closes visual gap.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Noted for SH-519: Node2D vs Marker2D if @tool class makes gizmo redundant.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Noted for SH-519: push_warning is in the AC.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: The states come from paddle_animation_state_machine.gd. The doc references the FSM in the Hook section.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Resolved: SH-xxx removed, replaced with Out of scope section.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Resolved: added cross-reference to spike anticipation in the opening.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Resolved: SH-xxx placeholder removed.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Noted. INDEX listing is out of scope for this doc; the other spike docs follow the same pattern.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Noted for SH-519: push_warning on missing marker is in the AC.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Noted for SH-519. The editor tool addresses the viewport-visual gap across all approaches.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Noted for SH-519. The tradeoff is doc-scoped; implementer can reconsider.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Noted for SH-519. The diverging states are design-authored during the visual tuning pass.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Noted for SH-519. Doc now flags this as out-of-scope if FSM crouch-idle state is not added.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Noted for SH-519. Dynamic lookup per call is cheap and guards against stale maps.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: Resolved: doc now specifies offset storage and marker.pos + offset recompute on each transition.

designs/01-prototype/tech/2026-06-22-racket-position-per-state.md
J-Melon: The distinction is that the body collider does not move; the racket Area2D is a separate detection zone. The doc clarifies this in the opening.

PR #1034, opened 2026-06-23
feat: soul bound overlay

J-Melon:
Adds DevOverlayPanel to the dev menu with toggleable overlays for paddle colliders, state labels, ground ray, and soul bound. Makes the panel draggable via DraggableBehavior, matching the other dev panel patterns.


scripts/core/court.gd
```diff
@@ -95,6 +96,12 @@ func _ready() -> void:
 	if soul_bound != null:
 		ball_system.bound_y = soul_bound.global_position.y
 	ball_system.player_paddle = player_paddle
+
```
J-Melon, line 98: Overlay should read its own data. Add @export var soul_bound: Marker2D and @export var court_config: CourtConfig to SoulBoundDebugDraw. Read bound_y and court_width in _ready(). Remove this wiring block from court.gd. Matches GroundRayOverlay self-contained export pattern.

PR #1045, opened 2026-06-26
feat: soul bound visibility overlay and arc transition fix

J-Melon:
Adds debug-only overlays for the soul bound marker line and arc travel trajectory, with toggle checkboxes in the Bounce dev panel. Arc travel overlay now tracks each ball independently with per-ball arc history (completed arcs persist as dimmed ghosts, up to 3 per ball). The arc transition snaps the ball to the bound y on entry so the parabola starts at the marker line rather than a few pixels above it from frame overshoot.

Removes the dead DevMenu export and toggle wiring from DevHUD; those toggles now live in the DevBouncePanel tab.


scripts/dev/arc_travel_overlay.gd
```diff
@@ -0,0 +1,185 @@
...
+
+
+func _attach_to_tracker() -> void:
+	_tracker.ball_added.connect(_on_ball_added)
+	_tracker.ball_removed.connect(_on_ball_removed)
+	for ball in _tracker.get_balls():
+		_on_ball_added(ball)
+
+
+func _on_ball_added(_ball: Ball) -> void:
```
J-Melon: **GDScript-style** suggestion: `_on_ball_added` (L57) and `_on_ball_removed` (L61) are empty stubs, connected at L51-54. `_process` already polls `_tracker.get_balls()` every frame, so these signal handlers are unused dead code. Remove them and the connections, or implement them.

J-Melon: **Gru**

implemented: 4c149f5 removes the dead stubs and their connections.

PR #1070, opened 2026-06-29
feat: move ball drop overlay to dev scene with toggle

J-Melon:
Renamed CursorOverlay to BallDropOverlay and moved it from the court scene into dev_hud.tscn alongside the other debug overlays. Added a "Show drop ring" checkbox in the Ball State dev tab so the drop-ring indicator can be toggled on and off during drag.

The overlay now finds world coordinates through the canvas transform, matching the pattern DevBounceOverlay uses for world-to-screen projection.

Also shifts the right wall 50px inward (x:750 to x:700) to correct miss zone placement, and drops the now-dead `role` and `monitorable` overrides from the ball rack. A follow-up commit replaced the cursor_overlay group tree-walk with a static singleton on BallDropOverlay.


scripts/dev/dev_ball_state_panel.gd
```diff
@@ -111,6 +112,26 @@ func _on_ball_removed(ball: Ball) -> void:
 	_rows.erase(ball)
 
 
+func _add_overlay_toggle() -> void:
+	_connect_overlay_toggle.call_deferred()
+
+
+func _connect_overlay_toggle() -> void:
+	var overlay := get_tree().get_first_node_in_group(&"cursor_overlay") as BallDropOverlay
```
J-Melon: issue: this is brittle, remove the as and put a guard. you should have a group for a single node, try something other than a tree walk

J-Melon: **Gru** 88cdbb6: Replaced group tree-walk with `BallDropOverlay.instance` static singleton + `is_instance_valid` guards. The `cursor_overlay` group is gone.

PR #1101, opened 2026-07-21
Cadence ball resource scaffolding

J-Melon:
Cadence oscillates ball speed between half, normal, and double on a fixed timer, independent of paddle hits. The shift applies as a percentage scale after the tier clamp, so it can push the ball's speed past the tier ceiling or below the tier floor without disturbing tier-advance timing, the per-tier soul reward, or the first-reach ball upgrade. See `designs/01-prototype/tech/cadence.md` for the mechanism. The item's original workshop docking and upgrade flow was dropped in favor of the existing tier-completion auto-upgrade.

This branch also removes the final-consolidation bonus-speed window: a top-tier ball now plateaus at its ceiling like any other tier, with no separate end-state or once-per-rally reward. Per-tier consolidation (soul multiplier, first-reach upgrade) is unaffected and still fires on every tier advance.

Several supporting changes ride along: `ItemManager`'s pure passthrough forwards to `EffectManager` were trimmed to the ones actually needed as duck-typed seams, shift storage was extracted into its own `ShiftRepository` class, the unused `ball_magnetism` stat and pull logic were removed as dead code, and the CI asset-preview workflow was fixed to scope its LFS pull to only the diffed PNGs instead of the whole `assets/**` tree. The unused GodotIQ addon and its rules file were also removed from the repo.


J-Melon (commented):
test

PR #1134, opened 2026-07-24
refactor: drop targets self-register as scene nodes

J-Melon:
The four drop targets (Court, Venue, Rack, Character) become scene nodes that register themselves with the drag controller in their own `_ready()`, replacing the controller's hand-built factory methods. This is the first slice of the SH-542 refactor.

It also carries the design doc the rest of the work follows. The doc reshaped this session into an autoload-and-groups design: the controller becomes an autoload that queries a `&"drop_targets"` group on release rather than holding a registered list, and each target declares its own priority. That supersedes the `register_target()` model this PR's code still uses, so the code here is the starting point and the doc is where it lands. The sequence is five PRs; this closes the "drop targets are scene nodes" criterion on its own.

Related: #1075


tests/integration/test_item_drag_controller.gd
```diff
@@ -3,6 +3,18 @@ extends GutTest
 const ItemDragControllerScript: GDScript = preload("res://scripts/items/item_drag_controller.gd")
 const BallReconcilerScript: GDScript = preload("res://scripts/items/ball_reconciler.gd")
 const TimeoutControllerScript: GDScript = preload("res://scripts/core/timeout_controller.gd")
+const CourtDropTargetScript: GDScript = preload(
```
J-Melon, line 6: suggestion: split this up into individual drop target test files, this reduces the complexity of this one file

J-Melon: Good instinct on the file being heavy, but the split axis is off: nothing in here is a per-drop-target test. Every case drives ItemDragController through a full grab→attempt_release gesture with all four targets wired up as fixtures — the per-target coverage already lives in test_drop_targets.gd. The real smell is that this is an integration test sitting in unit/ (it drives the system through its entry points and verifies the whole chain, TESTING.md's integration definition). Moved it to tests/integration/ alongside test_regrab_preserves_ball_identity.gd, which uses the same rig.

PR #1251, opened 2026-08-30
Fix autoplay low-ball crouching

### LINKED ISSUE #1228: Autoplay never crouches for low balls
Autoplay never triggers the low/crouch swing.

**Steps to Reproduce**

1. Enable autoplay (space).
2. Let it return a low ball.

**Expected Behaviour**

 The paddle crouches and plays the low-swing animation, same as manual play.

**Actual Behaviour**

`PlayerPaddle._is_crouching()` reads live keyboard input directly, ignoring `input_blocked`; autoplay never presses the key, so it always swings standing.

**Acceptance Criteria:**

- [ ] Autoplay crouches for a low ball, matching manual play

MollyStablesTheFirst:
Fixes #1228.

Autoplay now uses the predicted incoming ball position to determine when the player should enter the low stance. The low stance is cleared when autoplay is disabled or when the paddle is not grounded.

Tested in-game with autoplay enabled and low incoming balls.

### DIFF
--- a/scripts/core/paddle_ai_controller.gd
+++ b/scripts/core/paddle_ai_controller.gd
@@ -115,6 +115,8 @@ func _track() -> void:
 			paddle_travel_bound,
 		)
 	)
+	# Update the player's low stance from the predicted ball position
+	_update_low_stance(predicted_y, bound_y)
 	var noisy_target: float = predicted_y + _noise_offset
 
 	var delayed_target: float = _apply_reaction_delay(noisy_target)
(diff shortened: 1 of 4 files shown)


scripts/core/paddle_ai_controller.gd
```diff
@@ -135,6 +137,19 @@ func _track() -> void:
...
 
+## Uses the predicted ball height to switch between grounded stance and
+## low stance when the incoming ball reaches the low-stance threshold.
+func _update_low_stance(predicted_y: float, ball_bounce_y: float) -> void:
+	if not paddle.is_grounded():
+		paddle.wants_low_stance = false
+		return
+
+	var grounded_y: float = paddle.global_position.y
+	var low_stance_threshold_y: float = (grounded_y + ball_bounce_y) * 0.25
```
J-Melon: suggestion: keep coordinates local to the player. 

This uses bound_y which is global to the court rather than coupled to the player coordinates.
You could use the coordinates of MidRacketAnchor and crouch when the predicted position goes lower than that.

Like: paddle.wants_low_stance = predicted_y >= paddle.position.y + mid_anchor.position.y

MollyStablesTheFirst: I've implemented your suggestion to use the coordinates of MidRacketAnchor.

